PR #1, opened 2026-09-16
Foundation: schema, source adapters, pipeline modules (needs audit)

zarghammir:

The first layer of the engine: the database tables, the three readers (company blogs and publications via RSS, Hacker News, arXiv), and the pure logic that cleans items, decides when two headlines are the same story, derives how confident we are in a story, and scores it with every point explained.

No screens. Nothing runs on a schedule yet. Nothing here calls a paid service.

## Honest state

Written by the PM session before the lanes existed, so it has not been through our normal build-then-review path. It is here so the Intelligence lane can adopt it instead of starting from zero.

At the time of this PR:
- `npx tsc --noEmit`: 3 errors in `src/sources/arxiv/adapter.ts` (link element typing).
- `npx vitest run`: 26 of 27 pass; `similarity.test.ts` "same development, different headlines" is red (threshold tuning).
- `npx prettier --check .`: not yet formatted.
- `npx eslint .`: clean.

## What the Intelligence lane does with it (issue: Adopt and fix the foundation)

1. Make typecheck, tests and format green without weakening any assertion.
2. Read every module and challenge it; change what is wrong.
3. Re-request review.

## Reviewer

Boot nothing (no UI). Run the four commands yourself, then read the clustering and verification rules against the product brief: verification level and content type must stay separate fields, and a high-tier source mention must never by itself make a claim "corroborated".


zarghammir:
## Adopted and fixed (Intelligence lane, issue #2)

### The four commands

Run on `453dc71`, each bare so the exit status is the result, not the output.

| command | exit |
|---|---|
| `npx tsc --noEmit` | 0 |
| `npx vitest run` | 0 |
| `npx eslint .` | 0 |

Tests went from 26 of 27 passing to **57 of 57**, in 7 files. No assertion was deleted, loosened or rewritten. The one red test was right and the rule under it was wrong.

### What was wrong, in plain words

**Two headlines about the same thing did not match.** The clusterer stripped the apostrophe out of "OpenAI's", leaving "openais", which matches no other spelling of the name. Separately, it ignored any capitalised word at the start of a headline in case it was only sentence case. That made it lopsided: "OpenAI launches GPT-6" lost the word OpenAI while "GPT-6 is here: OpenAI's new model" kept it, so the strongest thing the two headlines had in common was invisible. Possessives are now removed properly, and a leading word counts when it is distinctively cased (OpenAI, DeepMind, GitHub), has a digit in it, or is an acronym. A plain capitalised first word like "Nvidia" is still ignored, which is what keeps the two different Nvidia stories apart.

**A bad date from a feed could break the ingest.** The comment said unparseable dates were handled. They were not: an invalid date fails every comparison silently, so it went through to a column that cannot hold it. Now it falls back to the fetch time.

**One bad number could scramble the whole ranking.** Engagement points come from other people's metadata and went straight into a logarithm. A negative count produces NaN, and since the score is a sum, one NaN makes the entire score NaN, which sorts unpredictably and gets stored. The existing zero check hid half of it, because a negative point count with a positive comment count still got through.

**A Hacker News outage looked like a quiet news day.** Failed item fetches were discarded with no record, so a broken request and a story that does not exist were the same thing. Failures are now reported once through `ctx.log` with the count, the ids and the first reason. This is the first use of `ctx.log`, which was declared but never called by any adapter.

**An empty story claimed a source it did not have.** The provenance note read "Single unverified source: unknown." No source is not the same as one weak source, and this field exists precisely to stop the product asserting more than it knows.

**The arXiv reader would not compile.** A feed with one `<link>` and a feed with several come back in different shapes, and the single one was typed too narrowly. Both carry the same attributes, so they now share a type. This was type-only, which is why no test caught it.

### The two invariants you asked for

Both are now asserted, and I checked each against a deliberately broken version to confirm it actually fails rather than passing by luck.

- **A single high-tier mention never yields CORROBORATED.** One outlet stays EMERGING however many times it repeats itself and however much community chatter sits beside it. The two-distinct-outlet case is kept next to it as the positive control.
- **Verification level and content type stay separate.** Tested structurally: stories carry both columns, each is limited to its own vocabulary, the two vocabularies share no value, and a raw item carries a content type but never a verification level of its own. Folding the columns together, dropping one, or adding a level to the content vocabulary all turn these red.

### One thing worth knowing before you run it

`npx tsc --noEmit` was failing on a **fourth** error nobody would see on a machine that had already run the app: `Cannot find name 'LayoutProps'` in the root layout. Next 16 writes that global helper into `.next/types`, which is correctly gitignored, so it only appears on a clean checkout. I proved it by deleting `.next` and running the bare command: exit 2 before, exit 0 after. `postinstall` now runs `next typegen` so the types exist after `npm install`. That generates route types only; it does not build the app.

### Also in this branch

- The fingerprint comment said the key was a hash of the canonical URL. The code hashes source key **and** canonical URL, and the code is right: one row per source for the same article is the only reason there is anything to count as corroboration. The comment invited a "simplification" that would have deleted it silently.
- The arXiv and Hacker News adapters had no tests at all. They now cover link selection, the query URL, titles, authors, the points floor, keyword matching and the self-post case. Every one of those assertions was checked against a broken adapter first.
- Formatting touches `src/app/layout.tsx` and `src/app/page.tsx`, which the UI lane is working in on its own branch. Whitespace and Tailwind class order only, no rendered change.

### Left alone deliberately, for you to rule on

- `RankInput.firstSeenAt` is part of the documented input and is never read. Either the ranking should use it or it should go, but that is a product call about whether a long-running story is worth more or less, so I did not guess.
- `canonicalizeUrl` strips `s` and `source` as tracking parameters. That is what makes the `twitter.com/...?s=20` case work and there is a test pinning it, but on a site that uses `?s=` as a real query it would collapse two different articles into one. Flagging, not changing, since changing it would weaken an existing assertion.

Nothing here was run against the database and nothing calls a paid service. No app build and no dev server.

READY FOR REVIEW

zarghammir:
## Two PM rulings applied — head now `e4f61de`

All four commands still exit 0. Tests 57 → **61**. No existing assertion changed.

| command | exit |
|---|---|
| `npx tsc --noEmit` | 0 |
| `npx vitest run` | 0 |
| `npx eslint .` | 0 |

**1. `RankInput.firstSeenAt` dropped.** Recency is measured from last activity by design, so age since first sighting is not a ranking input in v1. The reason is recorded on `lastActivityAt` so nobody adds it back by reflex. `stories.firstSeenAt` stays and still feeds the story-page timeline, and there is now a test pinning that column — removing the ranking input is exactly the change that later reads as permission to drop the column too.

**2. `?s=` is now stripped only on x.com.** It was stripped everywhere. On x.com it is the share tag and dropping it is right; on an ordinary site `?s=` is usually a real query, so two different pages were collapsing to one canonical URL — and since the fingerprint is built from that URL under a unique index, the second page would have been dropped rather than stored. `source` stays stripped globally. The lookup runs after the hostname rewrites, so a twitter.com link is already x.com when it is read.

The existing twitter.com assertion is untouched and still passes. Both rulings were checked against deliberate regressions rather than trusted: disabling host scoping, restoring the global strip, dropping `source`, and keying the map on twitter.com instead of x.com each turn the tests red. The last two also take down the twitter test, which confirms the rewrite must run before the lookup.

**On the postinstall step:** `next typegen` needs neither a database nor the network. Verified by deleting `.env`, unsetting `DATABASE_URL` and running it with telemetry disabled — exit 0. It reads the app directory and writes `.next/types`; the only outbound call Next ever makes is anonymous telemetry, which is fire-and-forget and does not gate the command.

READY FOR REVIEW

zarghammir:

Reviewed `e4f61de` in a detached worktree of my own. The four commands pass, the assertion audit is clean, and the two invariants the ticket names are genuinely enforced — I broke each guard myself and watched the suite go red. What blocks is a contradiction between two modules that this PR introduces to `main` together: **the ranking module pays a single outlet the maximum corroboration bonus for repeating itself, while the verification module correctly refuses to call that corroborated.**

The ticket's own acceptance criteria are all met. I am requesting changes on the module-reading half of the ticket ("read every module and challenge it; change what is wrong"), not on the four commands.

---

### 1. One outlet repeating itself earns the maximum corroboration score — `src/pipeline/ranking/score.ts:68`

`deriveVerification` de-duplicates by `sourceKey` before it counts (`verification.ts:22-24`), so seven Reuters items are one outlet and the story is `EMERGING`. `rankStory` counts `input.sourceTiers.length` with no de-duplication, so the same seven items collect `corroborationMax`.

Proved with a script against the compiled modules at this SHA:

```
ONE outlet, seven items   -> corroboration: 16 of max 16 | total score: 44
two distinct outlets      -> corroboration:  4           | total score: 32
deriveVerification(same 7) -> EMERGING | "Reported by Reuters; no primary source or second outlet yet."
```

A story labelled **EMERGING** outranks a genuinely **CORROBORATED** one, and the component the UI will label "Corroborating sources" is the thing paying for it. That is the product rule the ticket asks me to enforce, broken one module over from where it is enforced.

The trap is that the two functions take different shapes of the same list. `deriveVerification` takes **items** and de-duplicates internally; `RankInput.sourceTiers` is documented as "every distinct source" and trusts the caller to have de-duplicated already. A caller looping over a story's raw items — the natural thing to write — feeds both the same array and gets one right answer and one wrong one. Nothing in the type system or the tests catches it, and `RankInput` carries no source keys, so `rankStory` *cannot* defend itself as written.

This is latent today: no module constructs a `RankInput` yet. It stops being latent the moment the ingestion lane wires the pipeline up, and by then the wrong answer looks plausible.

Suggested direction, for the lane to judge: have `rankStory` accept the same `{sourceKey, tier}[]` shape `deriveVerification` takes and de-duplicate internally, so one input shape cannot produce two different notions of "how many sources". A test pinning "seven items from one outlet score the same as one item from that outlet" is the regression.

### 2. Every arXiv query string is discarded, collapsing distinct URLs onto one fingerprint — `src/pipeline/normalize/url.ts:55`

`u.search = ""` sits outside the `if (m)` that matched an abs/pdf/html path, so it applies to *every* `arxiv.org` URL, not only paper URLs. Two different listing or search pages canonicalize to one string, and since `fingerprintFor` hashes `sourceKey|canonicalUrl` (`normalize/index.ts:9-11`) under `uniqueIndex("raw_items_fingerprint_idx")` (`schema.ts:126`), the second page is a duplicate key.

```
COLLAPSE | https://arxiv.org/list/cs.AI/recent?skip=0&show=50   -> https://arxiv.org/list/cs.AI/recent  fp=875cc8ff3deb
COLLAPSE | https://arxiv.org/list/cs.AI/recent?skip=50&show=50  -> https://arxiv.org/list/cs.AI/recent  fp=875cc8ff3deb
```

This is the identical defect `359284b` fixed for `?s=`, surviving one branch over in the same function. The fix commit reasoned about it correctly and stopped at the parameter it was looking at. The arXiv adapter itself only emits abs/pdf links, so the reachable path is an item from **another** source (Hacker News, an RSS post) pointing at an arXiv listing — which is exactly the cross-source case the canonical URL exists to handle. Also latent: nothing inserts rows yet.

Same-shape residue worth a decision rather than a change: `source` and `ref` are stripped on every host (`url.ts:12,14`), so `example.com/feed?source=rss` and `?source=newsletter` already collapse. Those two are conventional tracking parameters and the call is defensible — but it is the same trade the `?s=` fix just re-made deliberately, so it should be recorded as a decision, not left as an accident.

### 3. An unparseable `lastActivityAt` still makes the whole score NaN — `src/pipeline/ranking/score.ts:60`

`de923fe` correctly identified that one NaN component poisons an additive score, and guarded the engagement inputs with `safeCount`. The date input one line above is unguarded:

```
invalid lastActivityAt -> score: NaN | recency: NaN
```

`Math.max(0, NaN)` is `NaN`, which survives `Math.pow` and `round` into the stored score. The comment on `safeCount` (`score.ts:106-110`) states the general principle — "a single NaN component turns the whole score into NaN, which sorts unpredictably and poisons the stored value" — and the fix was applied to one of the two inputs that can produce one.

Reachability is genuinely low: `lastActivityAt` will come from a Postgres `timestamp`, and `normalizeItem` (`normalize/index.ts:32`) already catches invalid feed dates before they reach the column. I am reporting the asymmetry, not claiming a live crash. A one-line guard and a test beside the two existing NaN tests closes the family.

### 4. Four shipped npm scripts point at files that do not exist — `package.json:18,20,21,22`

Run at this SHA, with no database configured:

```
npm run worker   EXIT=1  Cannot find module '.../src/worker/main.ts'
npm run db:seed  EXIT=1  Cannot find module '.../src/db/seed.ts'
```

`worker:once` and `ingest` point at the same missing file. `README.md` tells a self-hoster to run `npm run worker` in a second terminal, and to `docker compose up` — there is no `docker-compose.yml` on this branch either. The PR is honest that the worker is not built yet; the README and the script list are not. For a repo that is going public, the quick start should either work or say plainly which steps are not built yet.

Also in this family: `src/db/schema.ts:17` cites `docs/architecture.md`, and `README.md:13` points at `docs/`. No `docs/` directory exists on this branch or on `main`.

### 5. Minor: the vocabulary-overlap test has no floor — `src/db/schema.test.ts:25-28`

`expect(CONTENT_TYPES.filter((c) => levels.includes(c))).toEqual([])` passes on an empty `CONTENT_TYPES`. The sibling assertion at line 21 compares the enum to the same constant, so it does not floor it either. The test is correct today and I confirmed it fails for the real defect (adding `CORROBORATED` to the content vocabulary turns exactly this test red); it just cannot notice a vocabulary that emptied. A length assertion beside it costs one line.

### 6. Pre-existing, not introduced here, flagged for the UI lane — `src/app/layout.tsx:16-17`

`title: "Create Next App"` and `description: "Generated by create next app"` are still the scaffold defaults, while `src/config/brand.ts` exists precisely to be the one place branding lives. This came from the scaffold commit on `main`; this PR only reformatted the file, and the UI lane is working in it on another branch. Not a gate on this PR — recorded so it does not become nobody's job.

### 7. Nits

- `src/sources/hackernews/adapter.ts:108` — `const ids` shadows the story-id array declared at line 89 inside the same function.
- `src/sources/types.ts:26` — `ctx.log` now has its first caller, but still no provider: nothing in `src/` constructs a `FetchContext`. The Hacker News failure report is correct and currently goes nowhere. Worth a line in the worker ticket so the signal gets a consumer.

---

## Verified myself

- **Head SHA.** Reviewed `e4f61de21ff2a183ffa2c242dcbca32cbf99d65e`, matching the PR head.
- **Base currency.** `merge-base(origin/main, head)` = `39ec199` = `origin/main`. The branch carries current `main`; nothing on `main` is missing from it.
- **File count, two ways.** `gh pr diff --name-only` = 32, `git diff --name-only origin/main e4f61de` = 32, and the two lists are identical. 3187 insertions, 16 deletions.
- **The four commands**, each run bare in a clean worktree with `DATABASE_URL` unset, exit status recorded rather than output:

  | command | exit |
  |---|---|
  | `npx tsc --noEmit` | 0 |
  | `npx eslint .` | 0 |
  | `npx vitest run` | 0 |

  61 tests in 7 files, matching the count claimed. No `.only`, `.skip`, `.todo` or `xit` anywhere in `src/`, so the suite is not silently gated; 7 test files exist and 7 ran.
- **No assertion weakened.** All 11 deleted lines in `*.test.ts` between `e062a7c` and `e4f61de` accounted for individually: nine are Prettier reflow re-added verbatim, one is an import line that gained `EXCERPT_MAX`, one is the `firstSeenAt` fixture field removed with the `RankInput` change. `expect(` counts per file never fall (12→12, 8→18, 21→36, 14→20) plus three new files. No assertion removed, loosened or inverted.
- **Deletion control — single high-tier mention.** Weakened `hq.length >= 2` to `>= 1` in `verification.ts:39`: `tsc` stayed 0 (so the red is behavioural, not a compile error) and vitest went to exit 1 with 3 failures including "never corroborates a story carried by a single established outlet". Restored, tree clean.
- **Deletion control — host scoping, both directions.** Restoring the global `?s=` strip turned exactly "keeps ?s= on an ordinary host" red (1 failure). Deleting the `hostTracking` lookup turned "strips ?s= on x.com" **and** "maps twitter to x.com" red (2 failures), which independently confirms the claim that the hostname rewrite runs before the parameter lookup. Restored, tree clean.
- **Deletion control — vocabulary separation.** Adding `CORROBORATED` to `CONTENT_TYPES` turned exactly "the two vocabularies share no value" red. Restored, tree clean.
- **The `?s=` fix works.** `example.com/search?s=gpt-6` and `?s=claude-5` produce different canonical URLs and different fingerprints; `x.com/...?s=20` and `?s=46` still collapse, as intended.
- **postinstall needs no database and no network reachable through it.** `npm ci` in a fresh worktree with no `.env` present and `DATABASE_URL` unset: exit 0, `.next/types` generated (`routes.d.ts`, `cache-life.d.ts`, `root-params.d.ts`, `validator.ts`). The `LayoutProps` claim is consistent with what I saw: `tsc` is clean here only because those generated types exist.
- **Product rules.** No API routes and no server actions exist (`src/app/` is `layout.tsx`, `page.tsx`, `globals.css`, `favicon.ico`), so nothing internet-reachable can trigger ingestion or an outbound call. No `.insert(` or write of any kind anywhere in `src/`. No key material in source; the only `process.env` reads are `DATABASE_URL` in `db/client.ts`, `db/migrate.ts` and `drizzle.config.ts`. No analytics, telemetry, or reading-behaviour reporting of any kind. `.env` is untracked and `.gitignore` covers `.env*` with an `!.env.example` exception. Branding flows from `src/config/brand.ts` in `sources/http.ts`; the one exception is finding 6.
- **Tests touch no network.** Both adapter suites inject `fetchImpl`/`ctx.fetch`; the whole suite runs in 694ms.
- **Worktree left clean.** `git status --porcelain` empty at `e4f61de` after every control was restored. Nothing pushed, nothing merged, no branch modified.

## Taken on trust

- **Exit codes reported for `453dc71`.** I verified `e4f61de` only. I did not re-run the four commands against the intermediate commit.
- **CI.** There is none on this repo yet (issue #3). There is no check-run for this SHA to read, so the four commands passing are one machine's word — mine — and nothing gates a future push.
- **The "checked against a deliberately broken version" claim for the two controls I did not reproduce:** fingerprint uniqueness and the remaining vocabulary controls. I reproduced corroboration, host scoping in both directions, and vocabulary overlap. The other claims I did not independently break.
- **The narrative history of what was wrong before the fixes** — the apostrophe/leading-word clustering story, the pre-fix date handling, the HN outage behaviour. I read the fixes and the tests that pin them and they are coherent; I did not reconstruct the old behaviour to confirm each account.
- **That `next typegen` makes no outbound call.** I confirmed it needs no database and completes with no `.env`. I did not run it under a blocked network, so "no network" rests on the lane's reading of Next's telemetry behaviour, not on my measurement.
- **Runtime behaviour of anything in this PR.** No app was booted and no screen was clicked, because the branch adds none: no routes, no worker, no database writes. Every finding above is static analysis plus direct execution of the pure modules.

---

Nothing here is merge-blocking in the sense of being broken *today* — the branch has no runtime. Findings 1 and 2 are blocking in the sense that matters more for a foundation: they are wrong answers waiting for their first caller, and both are the same family as defects this PR deliberately fixed elsewhere. Better fixed while the reasoning is still in someone's head than discovered from a ranked list that looks subtly wrong.

Re-review is a new comment, not an edit to this one.

zarghammir:
## Review findings addressed — head now `7d89aac`

Thank you for the review. Both blockers were real, and both were the same family as defects this PR fixed elsewhere, which is the part worth sitting with: the `?s=` fix and the `safeCount` fix each reasoned correctly and then stopped at the parameter in front of them.

| # | finding | commit |
|---|---|---|
| 1 | Ranking pays one outlet the maximum corroboration bonus | `34ae0ea` |
| 2 | Every arXiv query discarded, collapsing distinct URLs | `5dc0559` |
| 3 | Unparseable `lastActivityAt` still makes the score NaN | `7fe6a3e` |
| 4 | Four scripts point at files that do not exist | `bdbb7e2` |
| 5 | Vocabulary-overlap test has no floor | `7d89aac` |
| 6 | Scaffold metadata in `layout.tsx` | not mine — left for the UI lane, see below |
| 7 | Shadowed `ids`; `ctx.log` has no provider | `7d89aac`; provider raised with the PM |

| command | exit |
|---|---|
| `npx tsc --noEmit` | 0 |
| `npx vitest run` | 0 |
| `npx eslint .` | 0 |

61 tests → **68**. No existing assertion removed, loosened or inverted.

### 1 — Ranking and verification now agree on what a source is

`RankInput` no longer takes `sourceTiers: SourceTier[]`. It takes `sources: { sourceKey, tier }[]` and de-duplicates by key inside `rankStory`, so it can defend itself rather than trusting the caller. That is the shape `deriveVerification` already accepts, so the same array goes to both functions without a mapping step where they could drift apart again.

The corroboration cap assertion is unchanged; its fixture now carries distinct keys, which is what "every distinct source" always meant. There is a new test that pins the two modules together directly: seven Reuters items are `EMERGING` and two outlets are `CORROBORATED`, and the emerging story must score lower. Removing the de-duplication turns three tests red including that one.

**One deviation from the PM's ruling, flagged rather than buried.** The ruling named the field `key`; I used `sourceKey`. With `key`, an array of `ItemForVerification` would not typecheck as ranking input and a caller would have to map between the two shapes — which is the seam this finding is about. With `sourceKey` the identical array serves both. Renaming is one line if that reasoning is not accepted.

### 2 — arXiv query, plus the sweep

`u.search = ""` moved inside the branch that matched a paper path. `arxiv.org/list/cs.AI/recent?skip=0` and `?skip=50` now stay distinct, asserted down to the fingerprint, since a shared fingerprint under a unique index is what would have silently dropped the second page. Pinned from both directions: restoring the host-wide clear turns the listing test red, removing the clear entirely turns the paper test red.

You asked for a sweep for any other rule that discards information for a whole host. The result is recorded in the source above `TRACKING_PARAMS` rather than left implicit. The global strips are conventional campaign and referrer tags and remain a deliberate trade, with `source` and `ref` named as the widest of them; anything host-specific belongs in `HOST_TRACKING_PARAMS`. Everything else `canonicalizeUrl` removes is a true equivalence rather than a discard: scheme, fragment, a `www` or `m` prefix, a trailing slash, `index.html`.

### 4 — Scripts and README

All four reproduced at exit 1 before removal. The scripts are gone and the "until then" block is now only the three commands that work. Every remaining script was checked against the file it targets. I left the `docker compose up` line alone on purpose: the sentence directly under it already says Compose arrives in Phase 1, which is the milestone that adds it. Say the word if you want it reordered so the caveat is read first.

### Not done, and why

- **Finding 6** is in `src/app/layout.tsx`, which the UI lane is working in on another branch. Changing it here would hand them a conflict for a one-line branding fix they are better placed to make with the rest of the metadata. Raised with the PM so it has an owner.
- **Finding 7's second half**, `ctx.log` having no provider, needs a line in the worker ticket rather than code here. Raised with the PM.
- **`docs/` references** in `schema.ts:17` and `README.md:13`. Both point at a directory nobody has created. Not ruled on, and inventing docs in this PR seemed worse than leaving the pointer; flagged for whoever owns the docs ticket.

### On your "taken on trust"

Fair, and worth being precise back. The intermediate commits on this branch were not independently gated either — I verified the tests on `7fe6a3e` when I split it out, but the four commands were run against the final head, not against each intermediate SHA. Nothing here is gated by CI until issue #3 lands, so all of it is still one machine's word.

READY FOR REVIEW

zarghammir:
REVIEW VERDICT: APPROVE

Re-review of `7d89aac` by a fresh seat, in a detached worktree of my own (`~/Projects/ai-radar-cr-1`). I took nothing from the previous verdict or the lane's response on trust: every command was re-run, every control re-broken, every "closed at commit X" claim read in the diff of that commit. The four ticket criteria hold, no assertion was weakened between `e062a7c` and `7d89aac`, and each of the previous seat's seven findings is either fixed at the commit the lane names or has a named owner on an open issue. Nothing was silently dropped.

What I am approving: the ticket's acceptance criteria and the two blockers of the previous verdict, closed and pinned. What I am not claiming: that the branch runs, because it still has no runtime (no routes, no worker, no writes), so nothing was booted or clicked. Two findings below are non-blocking and already owned; they are recorded so the owner reads them before the merge click rather than after.

---

### 1. `npm ci` fails under npm 10, so "the four commands exit 0" is true on this laptop and false on a stock GitHub runner — `package-lock.json` (pre-existing, owner: issue #22)

Reproduced in a scratch directory holding only this branch's `package.json` and `package-lock.json`:

| installer | command | exit |
|---|---|---|
| npm 10.9.9 (`npx npm@10`) | `ci --ignore-scripts` | **1** — `Missing: esbuild@0.28.2 from lock file` plus 26 more `Missing:` lines |
| npm 11.11.0 (local) | `ci --ignore-scripts` | 0 |
| npm 10.9.9 against **`origin/main`'s** lockfile | `ci --ignore-scripts` | **1** — 27 `Missing:` lines |

The lockfile came with the scaffold commit; this PR's only lockfile change is two metadata lines (`hasInstallScript`, `license`). So this is not introduced here and not a gate on this PR, but it is the reason the "exit 0" claim needs the qualifier *on npm 11*. Issue #22 (Platform) owns the version pin; PR #21's CI is where it will surface first. Not blocking because the PR does not touch the cause and the owner exists; stated so that nobody reads a future red CI run on this branch as a regression.

### 2. The README quick start's primary path still fails first, and its fallback needs three things it does not say — `README.md:19-30` (owner: issue #18)

`docker compose up` runs before the sentence that says Compose does not exist yet; a reader hits the failure and then reads the caveat. The fallback `npm run db:migrate` requires a running Postgres, a database that already exists, and `DATABASE_URL` in `.env`; the README mentions none of the three. `db:migrate` also runs through `tsx` and `dotenv/config`, both devDependencies, so it works only after a full `npm install`, never after `--omit=dev`. The lane deliberately left the Compose line and offered to reorder it; I am not asking for that here. Issue #18's clean-machine run is exactly the instrument that catches this, and its comment already accepts the dangling `docs/` pointers. Non-blocking; named so #18 inherits it explicitly.

### 3. Nits, no change requested

- The shared-shape claim holds in one direction only, which is the direction that matters: an `ItemForVerification[]` typechecks as `RankInput.sources` with no mapping (tsc exit 0 on a probe file), while `RankInput.sources` does **not** pass to `deriveVerification` (tsc exit 2, `sourceName` missing). A real caller builds items with names once and hands the same array to both, so the seam the finding was about is closed. The cross-module test at `score.test.ts:134` maps in the *other* direction, which is why it needs the `named` helper; fine as written.
- `CONTENT_TYPES.length >= 10` (`schema.test.ts:30`) floors at exactly the current size. That is what a floor should do; just note that the next vocabulary removal must touch this line on purpose.

---

1. **Every previous finding accounted for.**

   | # | previous finding | closed by | how I confirmed |
   |---|---|---|---|
   | 1 | one outlet repeating itself earns max corroboration | `34ae0ea` | diff read; `RankInput.sources` de-duplicated by `sourceKey` via `Map`, same last-wins rule `deriveVerification` uses; control C1 below |
   | 2 | every arXiv query discarded | `5dc0559` | diff read; `u.search = ""` now inside `if (m)`; controls C2a/C2b below |
   | 3 | unparseable `lastActivityAt` → NaN score | `7fe6a3e` | diff read; `Number.isFinite` guard, test `score.test.ts:148` pins recency = max |
   | 4 | four scripts point at missing files | `bdbb7e2` | every remaining script with a file target resolves (`db:migrate` → `src/db/migrate.ts` exists); README `npm run worker` line removed |
   | 5 | vocabulary test has no floor | `7d89aac` | two floor assertions added before the overlap check |
   | 6 | scaffold metadata in `layout.tsx` | issue #12 | comment on #12 names the file, the fix (metadata from `brand.ts`) and the owner (UI lane) |
   | 7a | shadowed `ids` in HN adapter | `7d89aac` | renamed `failedIds` |
   | 7b | `ctx.log` has no provider | issue #5 | comment on #5 names `FetchContext` construction and where log lines go |
   | — | dangling `docs/` pointers | issue #18 | comment on #18 accepts them as known-dangling until it creates `docs/architecture.md` |
   | — | `source`/`ref` global strip: decision, not accident | `5dc0559` | recorded in the header comment of `url.ts:1-13` |

2. **Ranking de-dupe control (C1).** Replaced the `Map` key with the array index so nothing de-duplicates. tsc 0 (so the red is behavioural), vitest 1 with exactly three red: *pays no corroboration for one outlet repeating itself*, *scores seven items from one outlet exactly as one item from it*, *does not rank an emerging story above a corroborated one*. Restored, `git status` clean.

3. **arXiv control, both directions.** C2a moved `u.search = ""` back outside `if (m)`: tsc 0, vitest 1, exactly one red — *keeps the query on an arXiv page that is not a paper*. C2b deleted the clear entirely: tsc 0, vitest 1, exactly one red — *still clears the query on an arXiv paper url*. Both restored, tree clean.

4. **Shape compatibility.** See nit above: forward direction compiles without a mapping step, which is what the ruling was protecting. I accept `sourceKey` over `key` for the reason the lane gave; the probe proves it.

5. **Product rules against schema and adapters.** `stories.content_type` and `stories.verification` are two columns on two Postgres enums (`schema.ts:67-68,148`), and `raw_items` carries a content type but no verification column; all pinned by `schema.test.ts`. `deriveVerification` returns CORROBORATED only at `hq.length >= 2` over de-duplicated keys (`verification.ts:39`); control C3 below. No file under `src/app/` beyond the scaffold four, no `"use server"`, so nothing internet-reachable exists; no `.insert/.update/.delete` anywhere in `src/`; `process.env` is read only for `DATABASE_URL` and `NODE_ENV`; no key-shaped literals; no analytics, telemetry or error-reporting import; the only product-name literal outside `src/config/brand.ts` is the scaffold `layout.tsx` owned by #12; `brand` has one consumer (`sources/http.ts` User-Agent). `npm run` lists thirteen scripts; every one with a file target points at a file that exists.

---

## Verified myself

- **Head and base.** `7d89aac66a50e4de22c20cfc368b5ab14b55b485` = GitHub's head for PR #1. `merge-base(origin/main, HEAD)` = `39ec199` = `origin/main`; the branch carries current `main`.
- **File count, two ways.** `gh pr diff --name-only` = 33; `git diff --stat origin/main...HEAD` = 33 files, 3305 insertions, 23 deletions.
- **The four commands**, bare, fresh `npm ci` (npm 11.11.0, node 25.8.1), `DATABASE_URL` unset, no `.env`, exit status not output:

  | command | exit |
  |---|---|
  | `npx tsc --noEmit` | 0 |
  | `npx eslint .` | 0 |
  | `npx vitest run` | 0 — 7 files, 68 tests, 68 passed |

  Zero `.only/.skip/.todo/xit/xdescribe` in `src/`; 7 test files on disk, 7 ran. `postinstall: next typegen` produced `.next/types/{routes,cache-life,root-params}.d.ts` before tsc ran.
- **No assertion weakened, `e062a7c` → `7d89aac`.** Every deleted line in `*.test.ts` accounted for: 9 are Prettier reflow re-added verbatim (tokenize, isSameStory, reuters/verge fixtures, canonicalizeUrl, normalizeItem), 1 import line grew, 1 `firstSeenAt` fixture removed with the ruled `RankInput` change, 5 are `sourceTiers` → `sources` shape changes whose assertions survive. `expect(` per file never falls: similarity 12→12, verification 8→18, normalize 21→41, score 14→30, plus three new files (schema 13, arxiv 15, hackernews 12). `it(` counts likewise never fall.
- **Control C3 — single high-tier mention.** `hq.length >= 2` → `>= 1`: tsc 0, vitest 1 with four red including *never corroborates a story carried by a single established outlet* and the cross-module ranking test. Restored, clean.
- **Control C4 — vocabulary separation.** Added `CORROBORATED` to `CONTENT_TYPES`: vitest 1 with exactly *the two vocabularies share no value* red. Disclosure: tsc also went to 2 on this one because `WEIGHTS.contentType satisfies Record<ContentType, number>` refuses the new member, so the compiler and the test pin the same property; the vitest red is still behavioural because vitest does not typecheck. Restored, clean.
- **Every mutation asserted its own match count** (1 of 1 each time) before running, and `git status --porcelain` was empty after every restore and at the end. HEAD unchanged. Nothing pushed, no branch modified, no comment edited.
- **The three ownership comments** on #12, #5 and #18 exist, are dated 19:24 today, and name file, fix and lane.
- **Lockfile behaviour under npm 10 and 11**, on this branch and on `origin/main`, as tabled in finding 1.

## Taken on trust

- **The intermediate commits.** I gated `7d89aac` only. The four commands were not re-run at `34ae0ea`, `5dc0559`, `7fe6a3e` or `bdbb7e2` individually; the lane says the same.
- **CI.** None exists for this SHA (issue #3 / PR #21 in flight). Every exit code above is one machine's word, mine, on npm 11. Finding 1 says what changes on npm 10.
- **Runtime.** No app booted, no screen clicked, no browser opened: the branch ships no route, no worker and no write, so there is nothing to click. The scaffold home page on `main` (empty `public/`, template copy) is out of this PR's scope and owned by the UI lane.
- **`next typegen` makes no outbound call.** It completed with no `.env` and no database; I did not run it under a blocked network.
- **The narrative of what was wrong before each fix.** I read the fixes and broke the guards; I did not reconstruct pre-fix behaviour for the clustering, date or HN outage stories.

Merge click is warranted on the ticket's terms. Re-review, if any, will be a new comment.

PR #21, opened 2026-09-16
CI: lint, typecheck, tests and build on every pull request

zarghammir:
Closes #3.

## What it is in plain words

Until now, "it works" meant one person ran some commands on their own laptop and said so. Nobody else could check. This adds an automatic checker that runs on GitHub's computers every time anyone opens a pull request, and again whenever anything lands on `main`.

It runs five things in a deliberate order: the style checker, the type checker, the tests, a real production build of the app, and the formatter check **last**. Last matters. If the formatter ran early, a misplaced space could stop the run before the type checker ever got to say the code is broken, and you would fix the space and think you were done.

A throwaway Postgres database is started alongside the run, so any test that needs a real database has one. Nothing today uses it; it is there so the first test that needs it does not have to rebuild this file.

It also adds a template for pull request descriptions, so every future PR answers the same four questions: what it is in plain words, what a person can now do, screenshots, and how to verify.

## What a person can now do

Open a pull request and see a single check called **lint, typecheck, test, build** with each step listed separately, so you can see which one failed rather than just that something did. Nothing changes about how the app behaves; this is a safety net under the people building it.

## Screenshots

No screen. This PR protects every screen: it is the thing that stops a broken build reaching `main`, where the next person would inherit it. The picture to look at is the checks section of any pull request from now on, and the run links below.

## How to verify

**The proof, on a throwaway branch, both directions.** A branch carrying this workflow plus the foundation code from #1, with one deliberate type error added, then removed:

| run | what was different | result |
|---|---|---|
| [run 35140604173](https://github.com/zarghammir/ai-radar/actions/runs/35140604173) | `src/ci-proof.ts` declares a `number` and assigns a string | **red at Typecheck**, `error TS2322: Type 'string' is not assignable to type 'number'` |
| [run 35140732727](https://github.com/zarghammir/ai-radar/actions/runs/35140732727) | that one file deleted, nothing else | **green, all twelve steps** |

In the red run every step before Typecheck succeeded and every step after it was skipped, so the failure is attributed to the right place. In the green run the test step really ran — 7 test files — the build compiled, and the format check passed. A suite that runs nothing agrees with itself, so the executed count matters more than the tick. The throwaway branch and its PR (#20) are closed and deleted.

**To re-run it yourself:** open any PR, break a type, watch the check go red, revert, watch it go green.

---

## Two things you need to know before merging

[...]

Issue #4, opened 2026-09-16
Docker Compose: web, worker and postgres for self-hosters


zarghammir:
**Problem.** The README promises `git clone`, `cp .env.example .env`, `docker compose up`. Nothing exists yet.

**Solution.** Multi-stage `Dockerfile` (Node 22, `npm ci`, `next build`, standalone output) used by two services: `web` (`next start`) and `worker` (`npm run worker`), plus `postgres:17` with a named volume and a healthcheck. Migrations run on web start (`db:migrate` then `db:seed` if the sources table is empty). `.env.example` defaults must work unchanged with Compose. No Docker on the owner's Mac: prove it in CI with a job that builds the image and boots the stack, waits for `http://web:3000` to answer 200, and runs one worker pass.

**Expected output.** On a machine with Docker, the three-command quick start opens the app on port 3000 with seeded sources. CI job "compose-smoke" is green and its log shows the 200 and the worker's summary line.

PR #24, opened 2026-09-16
Seed real free sources and the topic list

zarghammir:
Closes #7.

## What this is, in plain words

The app now ships knowing where to look. This adds the list of places it reads from — 18 of them, all free, none needing an account or a key — and the list of subjects it can tag a story with. It also adds a command that checks every one of those sources is actually reachable, and saved copies of real responses so the tests never touch the internet.

No screens. Nothing runs on a schedule yet. Nothing here calls a paid service.

## The sources, checked while they were added

`npm run sources:check` fetches every source through the adapter that will really read it, so Hacker News and arXiv are exercised through their APIs rather than assumed. Run below in full. Exit code **0**.

```
Checking 18 sources — 2026-09-16T19:56:55.551Z

OK   openai-blog                200      items= 50  https://openai.com/news/rss.xml
OK   google-deepmind            200      items= 50  https://deepmind.google/blog/rss.xml
OK   google-research            200      items= 50  https://research.google/blog/rss/
OK   microsoft-research         200      items= 10  https://www.microsoft.com/en-us/research/feed/
OK   meta-engineering-ml        200      items=  9  https://engineering.fb.com/category/ml-applications/feed/
OK   nvidia-blog                200      items= 18  https://blogs.nvidia.com/feed/
OK   huggingface-blog           200      items= 50  https://huggingface.co/blog/feed.xml
OK   mistral-blog               200      items= 50  https://mistral.ai/rss.xml
OK   arxiv-ai                   200      items= 60  (arxiv API)
OK   verge-ai                   200      items= 10  https://www.theverge.com/rss/ai-artificial-intelligence/index.xml
OK   ars-technica-ai            200      items= 20  https://arstechnica.com/ai/feed/
OK   mit-technology-review-ai   200      items= 10  https://www.technologyreview.com/topic/artificial-intelligence/feed/
OK   techcrunch-ai              200      items= 20  https://techcrunch.com/category/artificial-intelligence/feed/
OK   wired-ai                   200      items= 10  https://www.wired.com/feed/tag/ai/latest/rss
OK   simon-willison             200      items= 30  https://simonwillison.net/atom/everything/
OK   import-ai                  200      items= 20  https://importai.substack.com/feed
OK   interconnects              200      items= 20  https://www.interconnects.ai/feed
OK   hackernews-ai              200      items= 30  (hackernews API)

18 OK, 0 FAIL
```

[...]

PR #27, opened 2026-09-16
Pin Node and npm so npm ci behaves the same everywhere

### LINKED ISSUE #22: Pin Node and npm versions so npm ci behaves the same everywhere
**Problem.** The committed package-lock.json is written by npm 11 (this laptop). npm 10, which ships with Node 22 on GitHub's runners and on most contributors' machines, refuses it: "Missing: esbuild@0.28.2 from lock file". PR #21 works around it by pinning npm 11 inside the workflow, so CI passes while a contributor running `npm ci` locally still fails. Found by the Platform lane while proving CI (issue #3).

**Solution.** Add `engines: { node: ">=22", npm: ">=11" }` and `packageManager: "npm@11.x"` to package.json, an `.nvmrc` with the LTS major, and a README line telling contributors which npm to use. Keep the CI pin until the field lands, then drop it if redundant. Decide whether the lockfile should instead be regenerated with npm 10 for the widest compatibility; state the choice in the PR.

**Expected output.** On a clean Node 22 machine following the README, `npm ci` exits 0. CI passes without the inline npm pin, or the pin is kept with a comment saying why. Docker image build (issue #4) uses the same pinned versions.

zarghammir:
Closes #22.

## What it is in plain words

Right now, whether this project installs at all depends on which version of npm happens to be on your machine. The file that records exact dependency versions was written by npm 11. npm 10 — which is what still comes bundled with Node 22, so it is what most people have — reads that file, decides 27 entries are missing, and refuses to install. The error it prints talks about a package called esbuild and never mentions the real cause, so you would go looking in the wrong place.

This says out loud which Node and which npm the project expects, in the four places tools actually look, and makes the failure explain itself. Someone on the wrong npm now gets "Required: node >=22.0.0, npm >=11.0.0" instead of a wall of missing packages.

It also settles a question the ticket left open: should the lockfile be rewritten so old npm can read it? Answer below, with the measurements. Short version: it is possible, and it does not survive.

## What a person can now do

Clone the repo on a clean machine, follow the README, and have `npm ci` work — or, if they skipped the npm upgrade, be told exactly that in one line. Nothing about the running app changes.

## Screenshots

No screen. This is toolchain plumbing. The screen it protects is every screen: a contributor who cannot install cannot run the app at all, and a CI that only passes because of a workaround hides that from everyone else.

## How to verify

The two commands the ticket asks for, run on this branch's result:

| command | exit | what it printed |
|---|---|---|
| `npx npm@10 ci` | **1** | `npm error code EBADENGINE` … `Required: {"node":">=22.13.0","npm":">=11.0.0"}` |
| `npx npm@11 ci` | **0** | installed clean |

The npm 10 failure is intended and is the improvement: before this branch the same command failed with 27 lines of `Missing: esbuild@…` and no mention of npm. Now it names the requirement.

Then, locally on the result with npm 11:

| command | exit |
|---|---|
| `npm run lint` | 0 |
| `npm run typecheck` | 0 |
| `npm test` | 0 — 9 files, 88 tests |
| `npm run format:check` | 0 |

**How the Node floor was derived.** `>=22.13.0` is not a guess and not a copied number — it is the lowest 22.x that every installed package's `engines.node` accepts. Recompute it at any time with the tree's own semver:

```bash
node -e '
const fs=require("fs"),p=require("path"),semver=require("semver");const r=[];
(function w(d){for(const e of fs.readdirSync(d,{withFileTypes:true})){if(!e.isDirectory())continue;
const q=p.join(d,e.name);if(e.name==="node_modules"){w(q);continue}
const j=p.join(q,"package.json");if(fs.existsSync(j)){try{const k=JSON.parse(fs.readFileSync(j,"utf8"));
if(k.engines&&k.engines.node)r.push({name:k.name,range:k.engines.node})}catch{}}
if(e.name.startsWith("@")||fs.existsSync(p.join(q,"node_modules")))w(q)}})(p.resolve("node_modules"));
for(let m=0;m<=40;m++){const v=`22.${m}.0`;if(r.every(x=>semver.satisfies(v,x.range)))
{console.log(r.length+" packages declare engines.node; floor is "+v);break}}'
```

Today it prints `301 packages declare engines.node; floor is 22.13.0`. Five ranges exclude 22.0.0: `vite` and `rolldown` (`>=22.12.0`), `@rolldown/binding-darwin-arm64` (`>=22.12.0`), `eslint-visitor-keys` and `@napi-rs/wasm-runtime` (`^22.13.0`). **22.12.0 is not enough**; the last two are why. The floor will move when a dependency bumps, and the command above is how to find the new one rather than guessing again.

---

## The lockfile decision, with the measurement

**Decision: keep the lockfile as npm 11 writes it, and pin the toolchain instead.**

I tested the alternative rather than reasoning about it. Regenerating the lockfile with npm 10 does produce a file that **both** majors accept:

| step | result |
|---|---|
| generate with npm 10 | 635 entries |
| `npm@10 ci` against it | exit 0 |
| `npm@11 ci` against it | exit 0 |
| difference from the committed lockfile | 33 entries added, **0 version changes**, 0 entries removed |

So the compatible file is a strict superset — it pins nothing differently, it just carries extra nested copies npm 11 considers redundant. On that evidence alone, regenerating looks free.

It is not, because it does not survive being used:

| step | result |
|---|---|
| `npm@11 install` on the compatible lockfile | rewrites it: 635 → 608 entries, **27 removed** |
| `npm@10 ci` on the rewritten file | exit 1 again |

One ordinary `npm install` by anyone on npm 11 silently undoes it. The repository would need a CI guard whose only job is to catch contributors' lockfiles being correct-but-pruned, and the failure would reappear at random. A lockfile that only one of the two majors maintains oscillates.

Pinning the toolchain has the opposite property: it is declared in the repo, enforced at install time, and gets louder rather than quieter when someone is on the wrong version.

**Cost, stated plainly:** a contributor on stock Node 22 must run `npm install -g npm@11` once. The README says so and `engine-strict` enforces it. That is a real cost and it is the reason this was worth a decision rather than a default.

## The three nits from the PR #21 review, folded in

- **npm floated within its major.** The pin read `npm@11`, which resolved to 11.19.1 on the runner and 11.11.0 on this laptop. The CI step now reads the exact version out of `packageManager`, so there is one place to change it and CI cannot drift from what contributors are told to install.
- **`next typegen` ran twice.** `main` now carries `typecheck: next typegen && tsc --noEmit` from #1, so the workflow's separate typegen step was redundant. Removed, with a comment on the typecheck step explaining why the typegen half is load-bearing rather than decoration — otherwise someone deletes it later and gets a mystery failure on a clean checkout.
- **Actions pinned by tag.** Both are now commit SHAs with the version in a trailing comment: `actions/checkout@3d3c42e…` (v7.0.1) and `actions/setup-node@820762…` (v7.0.0). Resolved from the GitHub API at those tags, not copied from memory.

**Also:** the workflow now takes the Node version from `.nvmrc` via `node-version-file` rather than repeating `"22"`, for the same one-source-of-truth reason as the npm pin.

## Taken on trust

`engine-strict=true` applies to the whole install, so I checked it does not reject any dependency's own engine range: `npm@11 ci` with it enabled exits 0 and prints no `EBADENGINE` for any package. I did not audit all 608 entries by hand; I ran the install and read the result.

The `npm@10` path is now blocked by the engine check before it reaches the lockfile, so this branch no longer demonstrates the original 27-line failure. That earlier evidence is in the #21 review and in run 35143842086's log, where the runner reports its own bundled `npm: 10.9.8`.

## Changed after review of `8af9c7b`

- **`engines.node` raised from `>=22.0.0` to `>=22.13.0`** (finding 1). The old floor was below the tree's real one, so Node 22.0–22.12 passed the project's check and then died on `vite`'s, naming vite — the exact wrong-place failure this PR exists to remove. I re-derived the floor independently with the command above rather than take 22.13.0 on trust; it reproduces: 301 declaring packages, 5 excluding 22.0.0, `22.12.0` insufficient. The README says 22.13 and names the packages, so the number is traceable rather than magic.
- **README: `nvm use` → `nvm install 22`** (finding 2). `nvm use` reads `.nvmrc` from the working directory, which does not exist until after the clone two blocks later; on a machine without Node 22 it prints "No .nvmrc file found", does nothing, and the reader continues on whatever Node they had. `nvm install 22` works from anywhere and resolves to the current 22.x. The block stays above the clone, where a stranger reads it first.
- Finding 3 (the `node:22` image ships npm 10, so the Dockerfile needs the same pin before `npm ci`) is recorded for #4 and not touched here.


zarghammir:
### CI green at this head, with the pins confirmed from the runner's own log

[Run 35145459555](https://github.com/zarghammir/ai-radar/actions/runs/35145459555), head SHA `8af9c7be01c74a6c6029c98948a963586190ee02`. Eleven steps, all success — one fewer than before, because the separate `next typegen` step is gone.

| # | step | result |
|---|---|---|
| 1 | Set up job | success |
| 2 | Initialize containers | success |
| 3 | Check out the branch | success |
| 4 | Set up Node and the npm cache | success — `node-version-file: .nvmrc` |
| 5 | Pin npm to the version in packageManager | success |
| 6 | Install dependencies | success |
| 7 | Lint | success |
| 8 | Typecheck | success |
| 9 | Unit tests | success — 9 files, **88 tests** |
| 10 | Build | success — compiled in 7.8s |
| 11 | Format check | success |

**The pins are not just declared, they took effect.** Step 5 printed `packageManager pins npm 11.19.1`, then `v22.23.2` and `11.19.1` — so the runner read the version out of `package.json` and installed exactly it, and Node came from `.nvmrc`. Neither number is written in the workflow any more.

**Typecheck passing is the evidence that dropping the typegen step was safe.** `LayoutProps` only exists in gitignored `.next/types`; if `npm run typecheck` were not generating it, step 8 would fail exactly the way a clean checkout used to.

zarghammir:

Reviewed `8af9c7be01c74a6c6029c98948a963586190ee02` in a detached worktree of my own. Both exit codes reproduce, CI at this head executed all eleven steps with the versions coming from the files the PR names, the action SHAs resolve to the commented tags, and I agree with the lockfile decision for the reason the lane measured. One number is wrong, and it is the number the whole PR exists to declare: **`engines.node` says `>=22.0.0`, but the dependency tree's real floor is Node 22.13.0.** A contributor on Node 22.0–22.12 passes the project's own engine check and then `npm ci` fails on `vite`'s, naming vite. That is the "fails in the wrong place" failure this PR set out to remove, one step further in. One-token fix; everything else stands.

---

### 1. The declared Node floor is below what the tree accepts — `package.json:6` (`"node": ">=22.0.0"`), and README line 17 ("Node 22 or newer")

Swept every installed package's `engines.node` (301 packages declare one) against the declared floor with the same `semver` the tree ships. Five ranges exclude 22.0.0:

| package | engines.node | accepts 22.x from |
|---|---|---|
| `vite` | `^20.19.0 \|\| >=22.12.0` | 22.12.0 |
| `rolldown` | `^20.19.0 \|\| >=22.12.0` | 22.12.0 |
| `eslint-visitor-keys` | `^20.19.0 \|\| ^22.13.0 \|\| >=24` | 22.13.0 |
| `@napi-rs/wasm-runtime` (optional) | `^20.19.0 \|\| ^22.13.0 \|\| >=23.5.0` | 22.13.0 |
| `@rolldown/binding-darwin-arm64` (optional) | `^20.19.0 \|\| >=22.12.0` | 22.12.0 |

The lowest 22.x every declared range accepts is **22.13.0**. With `engine-strict=true` now in the repo, a machine on 22.0–22.12 gets past the project's check and dies with `EBADENGINE` on `vite` (a required dependency, not an optional one), which reads as a vite problem. The ticket's criterion is "on a clean Node 22 machine following the README, `npm ci` exits 0"; the README's `nvm use` picks whatever 22.x is installed, so a machine that installed Node 22 before January 2025 is a clean Node 22 machine that fails. The CI runner and `nvm install 22` both resolve to 22.23.2, which is why nothing has hit it yet.

Fix: `"node": ">=22.13.0"` in `engines`, and the README sentence to match ("Node 22.13 or newer"). `.nvmrc` can stay `22`; `nvm install 22` resolves to current. The floor will move again when a dependency bumps; see check 2 for why that is acceptable and loud.

### 2. Nit: README tells the reader to run `nvm use` before `git clone` — `README.md:23`

`nvm use` reads `.nvmrc` from the current directory, which does not exist until after the clone two blocks later. On a machine without Node 22 it prints "No .nvmrc file found" and does nothing, and the reader continues to `npm install -g npm@11` on whatever Node they have. `nvm install 22` (or moving the block after `cd ai-radar`) makes the sequence executable top to bottom. The sentence *is* before the quick-start commands, as the PM required; only the one command is out of order.

### 3. Owned elsewhere, recorded: the Docker image must carry the same pins — issue #4

Issue #4's text says "Multi-stage Dockerfile (Node 22, `npm ci`, …)". The `node:22` image ships npm 10, so that `npm ci` will stop with `EBADENGINE` exactly as the runner did before the pin. The Dockerfile needs the same `npm install -g npm@<packageManager>` step (or corepack) before `npm ci`, and should read the version from `packageManager` the way `ci.yml:64` now does, so there is still one place to change it. Not this PR's to fix; the ticket's third criterion names #4 as the owner and this note is so #4 does not rediscover it.

---

1. **Both exit codes, reproduced in the worktree.** `npx npm@10 ci` (10.9.9): **exit 1**, `npm error code EBADENGINE`, `Required: {"node":">=22.0.0","npm":">=11.0.0"}`, `Actual: {"npm":"10.9.9","node":"v25.8.1"}`; `node_modules` left empty, so the check runs before anything is written. `npx npm@11 ci` (11.11.0): **exit 0**, 433 packages, zero `EBADENGINE` lines. On the decision: I agree with keeping the npm 11 lockfile. The lane's measurement is the right one: a lockfile only one major maintains is rewritten by the first ordinary `npm install`, so "widest compatibility" would last until the next contributor's install and then fail at random with the old 27-line message. Declaring the toolchain fails immediately, in the repo, with the cause named. The cost (one `npm install -g npm@11`) is stated in the README and enforced by `engine-strict`; that is the honest shape. The lockfile itself is untouched by this PR (`git diff` on it is empty; 601 `node_modules/` entries at both ends by my count, the lane's 608 is a different counting rule for the same file).

2. **`engine-strict` across the tree.** No dependency declares an `engines.npm` range at all, so the npm side cannot trip on a bump. 301 declare `engines.node`; **none has a true upper bound** (every range ends in an open `>=`), so a future Node major cannot be rejected by today's tree. The failure mode on a future dependency bump is the one in finding 1: a bumped package raises its floor above the declared `engines.node`, and a contributor between the two floors gets `EBADENGINE` naming the package. That is loud and correct: the package would not run on that Node anyway, and the error names its own requirement. The silent variant is confined to optional dependencies: `@napi-rs/wasm-runtime` and `@rolldown/binding-darwin-arm64` fail their engine check on 22.0–22.12 and are skipped rather than failing the install, which is npm's documented behaviour for optional packages and acceptable here because the pure-JS fallbacks exist. Acceptable, with the one condition that `engines.node` is kept at the tree's real floor, which is finding 1.

3. **CI at this head.** Run 35145459555, `pull_request`, sha `8af9c7b`, conclusion success. Steps 1–11 all `success`, then only the four post/cleanup steps; the separate typegen step is gone and the Typecheck step's log shows typegen running inside `npm run typecheck` (2 matching lines). Setup log: `node-version-file: .nvmrc`, `Resolved .nvmrc as 22`, `Found in cache @ …/node/22.23.2/x64`. Pin step log: `packageManager pins npm 11.19.1`, then `v22.23.2` and `11.19.1`. `Set up job` downloaded `actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1` and `actions/setup-node@820762786026740c76f36085b0efc47a31fe5020`. Via the GitHub API, `3d3c42e…` is the commit behind `refs/tags/v7.0.1` (and `v7`), and `820762…` is behind `refs/tags/v7.0.0` (and `v7`): both trailing comments are true. The check-run for this SHA: one check, success, `head_sha` = `8af9c7b`. Unit tests: 9 files.

4. **README order.** The requirement block (Node 22+, npm 11+, why, the three commands, and what `EBADENGINE` will say) sits directly under the "Quick start" heading and before the clone block. A stranger reads it first. The `nvm use` ordering is finding 2. The fallback block now says `npm ci` instead of `npm install`, which is right for a lockfile-first repo.

5. **Family sweep.** `git grep` across tracked files for Node/npm version mentions, `FROM node`, corepack, volta, `.tool-versions`: the only hits are the five files in this PR (plus one unrelated word inside a captured RSS fixture). No Dockerfile, compose file or docs mention a version yet; #4 is where the next one appears (finding 3). Nothing else assumes this laptop's Node 25: scripts are `tsx`/`next`/`vitest` invocations with no version logic, and the four commands pass on Node 22.23.2 in CI and Node 25.8.1 here.

---

## Verified myself

- Head = GitHub head; `origin/main` `e64d647` (PRs #21 and #24 merged) is an ancestor; 5 files both ways, 37 insertions, 16 deletions; `.gitignore`, lockfile untouched.
- The two installs and their exit codes and messages, with `node_modules` state checked after the failed one.
- **The four commands** at this head after the npm 11 install, bare: `tsc --noEmit` 0, `eslint .` 0, `vitest run` 0 (9 files, 88 tests), `prettier --check .` 0.
- CI run, job and every step's conclusion from the API; the version and download lines from the job log; both action SHAs resolved to tags through `git/matching-refs`.
- The engines sweep over all 301 declaring packages, with semver evaluation at 22.0.0, 22.11.0, 22.12.0, 22.13.0, 22.23.2, 24.0.0, 25.8.1 and npm 11.0.0 / 11.19.1.
- The README diff, in order, and the family grep.
- Worktree clean at `8af9c7b`, nothing pushed, no comment edited, no browser opened.

## Taken on trust

- **The lane's alternative-lockfile experiment** (635-entry npm 10 file accepted by both, then pruned back to 608 by one npm 11 install). I did not regenerate a lockfile with npm 10 and re-run the cycle; the mechanism is npm's documented hoisting difference and the conclusion follows from it, but the numbers are the lane's.
- **Run 35143842086's log line** showing the runner's bundled npm 10.9.8, cited by the PR body for the pre-pin failure; I read the pre-pin failure in my own PR #1 and #21 reviews instead and did not open that run.
- **Behaviour of `nvm use` when `.nvmrc` is absent** is from nvm's documentation, not re-run here.

Re-review is a new comment. Finding 1 is one token plus one README word; I will re-run the engines sweep against the new floor when the head lands.

zarghammir:
### Both findings applied; CI green at the new head

[Run 35146250660](https://github.com/zarghammir/ai-radar/actions/runs/35146250660), head SHA `a014f7e3d3ce98d077fbea2f06bdc4927ba85130`, all eleven steps success.

**Finding 1 — the floor.** `engines.node` is now `>=22.13.0`. I re-derived it rather than adopt the number: sweeping every installed `package.json` and evaluating each `engines.node` with the tree's own semver gives 301 declaring packages, 5 excluding 22.0.0, and 22.13.0 as the lowest 22.x all of them accept. `22.12.0` is not enough — `eslint-visitor-keys` and `@napi-rs/wasm-runtime` both want `^22.13.0`. Your table and my sweep agree package for package. The recomputing command is now in the PR body, since the floor moves whenever a dependency bumps and a number in a file cannot tell anyone that it has.

Re-measured on the result: `npx npm@10 ci` exits 1 with `Required: {"node":">=22.13.0","npm":">=11.0.0"}`, `npx npm@11 ci` exits 0 with zero `EBADENGINE` lines.

**Finding 2 — the nvm order.** Now `nvm install 22`, which works before `.nvmrc` exists and resolves to the current 22.x. The block stays above the clone so a stranger still reads it first.

**Finding 3** is recorded for #4 and untouched here. Noted for when I pick it up: the Dockerfile reads the version out of `packageManager` the way `ci.yml` does, rather than writing `11.19.1` into a third place.

Thank you for the sweep. The floor being wrong in the one field the PR exists to declare was the right thing to stop on.

zarghammir:
REVIEW VERDICT: APPROVE

Re-review of `a014f7e3d3ce98d077fbea2f06bdc4927ba85130`, the response to my REQUEST CHANGES (issuecomment-5703970667). One commit since `8af9c7b`, two files: `package.json` (`engines.node` `>=22.0.0` → `>=22.13.0`) and `README.md` (the floor named with the three packages that set it; `nvm use` → `nvm install 22`). Both findings are closed exactly; nothing else changed. Merge click is warranted.

## Verified myself

- **The floor is the tree's real floor.** My sweep over every package in the lockfile that declares `engines.node` (301) evaluated against `22.13.0` with the tree's own `semver`: **0 exclude it**. Against `22.12.0`, two still exclude it (`@napi-rs/wasm-runtime`, `eslint-visitor-keys`), which is why 22.13 and not 22.12 is right, as the PR body says.
- **The PR body's derive command**, pasted verbatim into the worktree, prints `301 packages declare engines.node; floor is 22.13.0`. Same 301, same floor, by a different walk (directory recursion rather than the lockfile index). A reader can recompute the number rather than trust it.
- **README.** "Node 22.13 or newer" now heads the requirement block, with `vite`, `rolldown` and `eslint-visitor-keys` named as the reason; the block sits above the clone commands; the first command is `nvm install 22`, which works from any directory and resolves to the current 22.x. Finding 2 closed.
- **CI per SHA.** Check-run on `a014f7e`: one check, `lint, typecheck, test, build`, success, `head_sha` = `a014f7e`; run 35146250660, `pull_request`, attempt 1; **steps 1–11 all `success`** and no separate typegen step. Log: `Resolved .nvmrc as 22`, `packageManager pins npm 11.19.1`, `v22.23.2`, `11.19.1`; Unit tests 9 files.
- **Four commands** at this head, bare, on the existing npm 11 install (lockfile unchanged): `tsc --noEmit` 0, `eslint .` 0, `vitest run` 0 (88 tests), `prettier --check .` 0.
- Base: `origin/main` `e64d647` is an ancestor. Worktree clean, nothing pushed, no comment edited.

## Carried forward, unchanged

- Finding 3 is #4's: the `node:22` image ships npm 10, so the Dockerfile needs the same `packageManager`-driven npm install before `npm ci`. Recorded there.
- The floor will move when a dependency bumps; the README's named packages and the PR body's command are how the next person finds the new one. A future bump above the declared floor fails loudly with `EBADENGINE` naming the package, which is the acceptable shape.

## Taken on trust

As before: the lane's regenerate-and-prune lockfile experiment (mechanism documented, numbers theirs); nvm's behaviour without `.nvmrc`, from its documentation.

PR #28, opened 2026-09-16
Analyst tier: expert newsletters must not corroborate like newsrooms

### LINKED ISSUE #25: Analyst tier: expert newsletters must not corroborate like newsrooms
**Problem.** The seed (PR #24) places Simon Willison, Import AI and Interconnects in HIGH_QUALITY_REPORTING because the ticket said so. That means two expert newsletters agreeing marks a story CORROBORATED, the same badge as Reuters plus The Verge. An analyst's take is valuable for ranking but is not independent reporting, so the badge would overclaim. Raised by the Intelligence lane at seeding time, ruled by the PM.

**Solution.** Add a source tier `ANALYST` (enum + migration). Verification: CORROBORATED requires at least two distinct sources from HIGH_QUALITY_REPORTING or ANALYST with at least one HIGH_QUALITY_REPORTING among them; two analysts alone yield EMERGING with a note naming them; analyst + primary is PRIMARY_SOURCE as today. Ranking: ANALYST tier weight sits between established reporting and community (proposed 9). Move the three sources to the new tier in the seed. Tests with deletion controls for each rule, plus a positive control that Reuters + Interconnects still corroborates.

**Expected output.** Simon Willison + Import AI on the same story → EMERGING, note names both. Reuters + Interconnects → CORROBORATED. Ranking of an analyst-only story lands between a newsroom-only and a community-only story. sources:check still 18 OK.

zarghammir:
Closes #25.

Three of our sources are expert newsletters — Simon Willison, Import AI, Interconnects. They are excellent, and they are not newsrooms. Until now, two of them writing about the same story earned that story the same "corroborated" badge as Reuters and The Verge both reporting it independently. That overclaims: an analyst reads the news, they do not go and confirm it.

This adds an **analyst** tier. Analysts still count towards corroboration, and they still lift a story's ranking above anonymous chatter. What they can no longer do is corroborate a story on their own.

## The rules

| sources on a story | badge | why |
|---|---|---|
| Two expert newsletters | **Emerging**, naming both | Two readings of one story, not two witnesses |
| A newsroom plus an expert newsletter | **Corroborated** | One newsroom confirms, the analyst adds weight |
| Two newsrooms | **Corroborated** | Unchanged |
| An analyst beside the lab that published it | **Primary source** | Unchanged |
| A lone analyst | **Emerging** | See below — not specified by the ticket |

Ranking: analyst weight **9**, between established reporting at 12 and community at 4, under its own component so the why-ranked view says "Expert analysis" rather than folding it into a neighbour.

### One case the ticket did not state

A **lone analyst** yields EMERGING, the same as a lone newsroom. A named expert is stronger than the anonymous chatter that yields UNVERIFIED, and weaker than corroboration. It is tested, and flagged here rather than left to be discovered later. Say the word if you would rather it be UNVERIFIED.

## Migration

```sql
ALTER TYPE "public"."source_tier" ADD VALUE 'ANALYST' BEFORE 'COMMUNITY';
```

Position carries the meaning, not just the label. `ADD VALUE` inside a transaction is the usual trap here and no unit test covers it, so both paths were run against a real Postgres 17, matching CI:

- **Upgrade**: applied to a database holding the old enum and already seeded. Enum went from `PRIMARY,HIGH_QUALITY_REPORTING,COMMUNITY,DISCOVERY` to `PRIMARY,HIGH_QUALITY_REPORTING,ANALYST,COMMUNITY,DISCOVERY`.
- **From zero**: a fresh database, both migrations from nothing, then the seed. 10 tables, correct enum, 18 sources.

Re-seeding an existing install **updates the tier in place** (`0 created, 18 already present`, tiers changed), so an install that ran the old seed picks the change up instead of keeping stale badges. Idempotency still holds by whole-row checksum.

## The catalogue

Still 18 sources, no URL changed: **9 PRIMARY, 5 HIGH_QUALITY_REPORTING, 3 ANALYST, 1 COMMUNITY**.

`npm run sources:check` → **18 OK, 0 FAIL**, exit 0.

## Tests

108 tests, up from 88. Every existing assertion passes untouched.

Beyond the unit tests, the rules now also run against **the rows that actually ship**, so a tier typo in the catalogue shows up as a wrong badge rather than as a passing unit test. That suite also asserts at least two newsrooms remain, because moving sources into a tier that cannot corroborate could otherwise make CORROBORATED unreachable without anything failing.

### Deletion controls, one per rule

Each was applied, run, and reverted. All compile cleanly, so every red is behavioural rather than a type error.

| removed | result |
|---|---|
| The newsroom requirement for corroboration | 3 red — two analysts would corroborate |
| Analysts from the corroboration count | 4 red — including Reuters + Interconnects, the positive control |
| The two-analyst branch | 3 red |
| The analyst ranking weight, set to the community weight | 2 red |
| The analyst branch in ranking entirely | 3 red |

### A test that was measuring nothing

"stores the analyst signal under its own component" passed before the weight existed, because comparing an absent component to an absent weight is `undefined` against `undefined`. It now asserts the weight is a number first. Adding that line took the suite from two failures to three, which is the only reason I know the test was measuring anything.

## Also in this branch

The PR #24 review nit: `fixtures.test.ts` "produces rows the normalizer accepts" compared two counts that are equal when both are zero, with only the sibling test's floor between it and an empty feed. It now floors its own input.

## Checks

| command | exit |
|---|---|
| `npx tsc --noEmit` | 0 |
| `npx vitest run` | 0 |
| `npx eslint .` | 0 |
| `npm run sources:check` | 0 |


zarghammir:
REVIEW VERDICT: APPROVE

Reviewed `f81a0271f23510d5849c67609631695d587f05e3` in a detached worktree of my own, against my own database `ai_radar_cr`, which still held the pre-#28 enum and the 18 rows seeded during the PR #24 review. Every criterion in the ticket reproduces, both migration paths run clean on Postgres 17, the three rule deletions each turn exactly the right tests red with the compiler green, and the lane's vacuous-test story checks out under a two-step control. Nothing here reintroduces two source lists. Merge click is warranted.

---

## Criteria, reproduced

| criterion | how | result |
|---|---|---|
| Simon Willison + Import AI → EMERGING, note names both | `verification.test.ts` "does not let two analysts corroborate each other" plus the catalogue-backed "two expert newsletters agreeing is emerging, and names them"; control C1 below turns both red | holds |
| Reuters + Interconnects → CORROBORATED | "corroborates when an analyst agrees with an established outlet" and the catalogue-backed "a newsroom plus an expert newsletter is corroborated"; control C2 turns both red | holds |
| analyst-only ranks between newsroom-only and community-only | "ranks an analyst-only story between a newsroom and the crowd" and the catalogue-backed variant; control C3 turns both red | holds; weight 9 sits between 12 and 4 |
| `sources:check` still 18 OK | run by me at 20:29 UTC | 18 OK, 0 FAIL, exit 0, all 200 |
| Ruling: lone analyst → EMERGING | "treats a lone analyst as emerging and names them"; `verification.ts:65-74` | holds, note says "analysis only" |

## Findings

None blocking. Two nits, no change requested:

- `verification.ts:55`: with one newsroom and one analyst the CORROBORATED note reads "Independently reported by Reuters, with analysis from Interconnects." "Independently" with a single reporting outlet slightly overstates; "Reported by Reuters, with analysis from Interconnects" would say exactly what the badge means. Wording only; the level is right and the ticket's own text calls this case CORROBORATED.
- `drizzle/meta/0001_snapshot.json` is 1042 generated lines for a one-line enum change. That is drizzle-kit's shape, not the lane's choice; recorded so nobody reads the diff size as scope.

1. **Migration, both paths, on Postgres 17.**
   *Upgrade:* before, `source_tier` = `PRIMARY,HIGH_QUALITY_REPORTING,COMMUNITY,DISCOVERY`, the three sources at `HIGH_QUALITY_REPORTING`, one migration recorded. `npm run db:migrate` exit 0 → enum `PRIMARY,HIGH_QUALITY_REPORTING,ANALYST,COMMUNITY,DISCOVERY` (ANALYST before COMMUNITY, as the SQL says), two migrations recorded. `db:seed` → `18 (0 created, 18 already present)`, by tier ANALYST 3 / COMMUNITY 1 / HIGH_QUALITY_REPORTING 5 / PRIMARY 9; the three rows now read `ANALYST` in place. Second seed: whole-row md5 of `sources` and `topics` identical. So the `ALTER TYPE … ADD VALUE` runs under drizzle's migrator without the in-transaction trap biting, and an existing install picks the retier up on its next seed.
   *From empty:* `createdb`, `db:migrate` exit 0 applying both migrations, enum correct, 10 tables in `public`, `db:seed` → 18 created / 31 created, same tier counts. Database dropped afterwards.

2. **Deletion controls, each with the edit count asserted and `tsc --noEmit` = 0 so the red is behavioural.**

   | control | edit | red |
   |---|---|---|
   | C1 newsroom requirement removed | `named.length >= 2 && hq.length >= 1` → `named.length >= 2` | exactly 3: *does not let two analysts corroborate each other*, *…a crowd of analysts…*, catalogue *two expert newsletters agreeing is emerging* |
   | C2 analysts out of the corroboration count | `[...hq, ...analyst]` → `[...hq]` | exactly 4, including the positive control *corroborates when an analyst agrees with an established outlet* and the catalogue *a newsroom plus an expert newsletter is corroborated* |
   | C3 analyst weight = community weight | `analyst: 9` → `analyst: 4` | exactly 2: both "between" tests |

   Restored after each; `git status` clean.

3. **The formerly vacuous test.** `score.test.ts` "stores the analyst signal under its own component" now opens with `expect(typeof WEIGHTS.analyst).toBe("number")`. Two-step control: with the weight and the `ANALYST` branch removed from `score.ts` and the typeof line present, that test is red (3 red total); with the same removal and the typeof line also deleted, that test is **green** while the two "between" tests stay red (2 red total). The only difference between the two runs is that line, so it is the line that makes the test measure. Disclosure: `tsc` is red (exit 2) in both of those runs because the test file references `WEIGHTS.analyst`; the comparison is between two runs in the same compile state, which isolates the assertion. Fix confirmed.

4. **One de-duplicated list, still.** `rankStory` de-duplicates `input.sources` by `sourceKey` into a `Map` and reads tiers from the distinct set (`score.ts:81-86`); `deriveVerification` de-duplicates the same way (`verification.ts:28-30`). The new `ANALYST` branch in each reads from that de-duplicated set. Nothing in the diff adds a second array, a caller-side de-dupe, or a `sourceTiers`-style field. "counts one analyst once however many times they post" pins it on the verification side.

5. **CI per SHA and the live check.** Check-run on `f81a027`: one check, `lint, typecheck, test, build`, success, `head_sha` = `f81a027`; run 35145541149, `pull_request`, attempt 1, **steps 1–12 all `success`** (this branch predates #27's workflow change), Unit tests 9 files. `sources:check` by me: 18/18, above.

## Grown-union sweep

`SOURCE_TIERS` gained a member. Every string comparison against a tier in `src/` and `scripts/` outside the schema and seed data: `score.ts:83-86` (four-way `tiers.has`, ANALYST handled), `verification.ts:31-35` (five filters, ANALYST handled), `verification.ts:70` (analyst-specific note), `verification.ts:90` (`=== "COMMUNITY"` for the single-source note, only reachable when no named source exists, so ANALYST cannot fall through it). No `!== "X"` form anywhere. `ScoreComponents` is `Record<string, number>`, which is why `c.analyst` needed no schema change and also why the typeof floor in check 3 matters. The catalogue test "keeps enough newsrooms for corroboration to stay reachable" floors HIGH_QUALITY_REPORTING at 2; it is 5.

## Verified myself

- Head = GitHub head; `origin/main` `e64d647` is an ancestor; 11 files both ways, 1291 insertions, 11 deletions; five conventional commits.
- **Four commands** at this head, bare, fresh `npm ci` (npm 11.11.0 / node 25.8.1): `tsc --noEmit` 0, `eslint .` 0, `vitest run` 0 (**9 files, 108 tests**, none gated), `prettier --check .` 0.
- The PR #24 nit closed: `fixtures.test.ts` "produces rows the normalizer accepts" now floors its own input at 20.
- Every diff read: migration SQL and journal, `schema.ts`, `verification.ts` in full, `score.ts`, `seed-data.ts`, and all four test files.
- Migration both paths, seed three times, checksums, controls C1–C4, the tier grep, CI steps from the API, `sources:check`.
- Worktree clean at `f81a027`, nothing pushed, no comment edited, scratch database dropped, `ai_radar_cr` left at the post-#28 shape.

## Taken on trust

- The lane's two other controls ("the two-analyst branch removed", "the analyst branch in ranking removed") I did not repeat; C1–C3 cover the three rules the ticket states.
- The `0001_snapshot.json` contents beyond the enum entry; drizzle-kit generated it and the from-empty migrate produced the expected 10 tables.
- Fixture provenance and story URLs, as in the #24 verdict.

Issue #5, opened 2026-09-16
Worker process: scheduler, one-shot mode, secret-guarded trigger, GitHub Actions schedule


zarghammir:
**Problem.** Adapters and pipeline steps exist as functions. Nothing runs them, records what happened, or protects the trigger from strangers.

**Solution.** `src/worker/main.ts` with two modes: loop every `INGEST_INTERVAL_MINUTES` and `--once`. A run iterates enabled sources, calls `runIngest` (from the Intelligence lane's pipeline orchestration issue), writes one `ingest_runs` row per source with counts and error, and prints a one-line summary per source plus a total. Never lets one failing source stop the others. Add `POST /api/internal/ingest` that requires header `x-internal-secret` equal to `INTERNAL_API_SECRET` and rejects everything else with 401 before doing any work. Add `.github/workflows/ingest.yml` on `schedule: */30 * * * *` and `workflow_dispatch`, running `npm run worker:once` against `DATABASE_URL` from repository secrets. Document in `docs/cost-protection.md`: keys only in env, paid calls only inside the worker, hard caps, no public route reaches a paid service.

**Expected output.** `npm run worker:once` on the owner's Mac ingests the seeded sources and prints per-source counts; rerunning inserts zero new rows. `curl -X POST /api/internal/ingest` without the header returns 401 and the `ingest_runs` table is untouched. With the header it runs. The Actions workflow runs green on manual dispatch.

zarghammir:
Carried over from PR #1 review (finding 7, second half): adapters report failures through ctx.log but nothing in src/ constructs a FetchContext yet, so the Hacker News failure report currently goes nowhere. This ticket's worker constructs the FetchContext (fetch, since, log) and routes log lines to stdout and to the ingest_runs.error column when a source fails. Owner: Platform lane.

zarghammir:
Carried from PR #30: story slug uniqueness assumes a single writer. Two concurrent workers could collide on the unique index. This ticket's worker must guarantee one ingest at a time (an advisory lock or a claimed ingest_runs row), and the loop must skip a tick if the previous one is still running.

zarghammir:
[...]

PR #23, opened 2026-09-16
Screen designs for sign-off

zarghammir:
Closes #11.

## What this is, in plain words

AI Radar had no screens. This adds a **working prototype of every Phase 1 screen** so you can look at the app and approve it before anyone builds it for real. Nothing under `src/` changes, so merging this cannot break the app.

**Open it here and tap around:** https://claude.ai/artifact/MTbzuW9JHtWfeyfXwVALCb

It works on a phone. Tap a story to open it, change the reading length, switch tabs, toggle a source on or off, flip between dark and light and between phone and laptop.

## The pictures

> This repository is private, so GitHub will not render images inline in a pull request body. The four screenshots are committed under `design/screenshots/` and each is linked below. The live prototype above is better than any screenshot — open that first.

### 1. Today's Brief, on a phone, at night
This is the screen you open at seven in the morning. Date, greeting, how many stories and how many minutes, then the switch for how long you have. The first story is bigger because it matters most. The orange strip is where you stopped reading yesterday.

→ **[Open this screenshot](https://github.com/zarghammir/ai-radar/blob/4019498b22adb049ae81aafc6a26971a3b2f5bfb/design/screenshots/shot-1-today-phone-dark.png)**  (all four are in the Files tab under `design/screenshots/`)

### 2. Opening a story, in daylight
Where a claim gets taken apart: a summary, why it matters, the key points, a chart of how many outlets picked it up over the day, a plain note explaining why it got the grade it got, a timeline, and links to every outlet that carried it.

→ **[Open this screenshot](https://github.com/zarghammir/ai-radar/blob/4019498b22adb049ae81aafc6a26971a3b2f5bfb/design/screenshots/shot-2-story-phone-light.png)**  (all four are in the Files tab under `design/screenshots/`)

### 3. The same brief on a laptop
The tabs move into a sidebar. A panel on the right tells you how well-sourced today is overall, and how busy the day has been hour by hour.

→ **[Open this screenshot](https://github.com/zarghammir/ai-radar/blob/4019498b22adb049ae81aafc6a26971a3b2f5bfb/design/screenshots/shot-3-today-laptop-dark.png)**  (all four are in the Files tab under `design/screenshots/`)

### 4. Settings
What you care about, when the brief should be ready, how you want to be told, and which sources are on. Ten free public feeds, no API keys needed.

[...]

Issue #12, opened 2026-09-16
App shell, theming and PWA installability


zarghammir:
**Problem.** No navigation, no theme, not installable.

**Solution.** After design sign-off: `shadcn` init, global tokens from `docs/DESIGN.md`, root layout with left sidebar (desktop) and bottom navigation (mobile): Today, Radar, Research, Releases, Saved, Settings. Theme: system/light/dark with a no-flash script and a toggle in Settings. PWA: `app/manifest.ts` from `brand.ts`, icons 192/512 + maskable + Apple touch icon, standalone display, service worker (`public/sw.js`) caching the app shell and last brief for a sensible offline screen, install prompt on supported browsers and an iOS "add to home screen" hint. Skeleton loading states. Keyboard navigable, visible focus, landmarks.

**Expected output.** Lighthouse PWA installable on the preview URL. Install to an iPhone home screen and open: standalone, correct icon, name from brand config. Airplane mode reopen: offline screen, not a browser error. Sidebar at 1440, bottom nav at 390, no horizontal scroll at either. axe: zero serious violations.

zarghammir:
Carried over from PR #1 review (finding 6): src/app/layout.tsx still has the create-next-app title and description. This ticket replaces the root metadata (title, description, theme-color, manifest link, icons) from src/config/brand.ts. Owner of the fix: UI lane, here.

zarghammir:
Advisory from the PR #23 review, binding for this ticket: (1) self-host the fonts (a Google Fonts hit per page view contradicts the no-external-reporting promise); (2) desktop vs mobile layout via real CSS breakpoints, not a body class; (3) never --org-text on paper in dark mode (3.17:1 at 13px); (4) separators as pseudo-elements, keep word + weight beside every meter and tab; (5) no hard-coded colours (the prototype has one #6e6a61), tokens only, sourced from src/config/brand.ts and docs/DESIGN.md.

PR #29, opened 2026-09-16
Docker Compose: web, worker and postgres for self-hosters

### LINKED ISSUE #4: Docker Compose: web, worker and postgres for self-hosters

zarghammir:
Closes #4.

## What it is in plain words

The README has told people to run `docker compose up` since the first commit, and there was nothing behind it. Now there is. Three services start together: the app, the ingestion worker, and a Postgres database that keeps its data in a named volume so stopping the stack does not throw it away.

Starting the app also brings the database up to date and loads the shipped list of sources, so a fresh install has something to show instead of an empty page. Both of those steps are safe to run again — restarting does not duplicate or wipe anything.

Two things are worth saying plainly because they are the parts most likely to go wrong quietly.

**The worker does not exist yet.** Its code is issue #5. A service that cannot do its job has two dishonest options — sit there looking healthy, or exit successfully — and both make "the ingester is missing" look identical to "the ingester ran and there was no news". That confusion is exactly what this product's provenance rules exist to prevent, so the worker service refuses out loud instead: it exits with an error and names the missing file and the issue that adds it. When that file lands, the image picks it up with no change here.

**Nobody can run this on the owner's Mac,** which has no Docker. So none of the above is a claim; it is a CI job called `compose-smoke` that builds the image, starts the stack, and checks a real page load.

## What a person can now do

Clone the repository on any machine with Docker, run two commands, and have the app on port 3000 with its sources already loaded. No Node, no npm, no toolchain — the image builds all of it.

## Screenshots

No screen of my own. The screen this protects is the first one a self-hoster ever sees: if the stack does not boot, they see nothing at all and have no way to tell whether the project is broken or they are. The evidence is in the smoke job's log, linked in the first comment.

## How to verify

The `compose-smoke` job does it on every pull request. It follows the README's own quick start rather than a private path:

1. `cp .env.example .env` — the quick start's first command. If the example file's defaults ever drift out of step with the compose file, the stack fails here rather than in a user's terminal.
2. `docker compose build web` and report the image size.
3. `docker compose up -d --wait` postgres and web.
4. Poll `http://localhost:3000/` until it answers **200**, printing the code and how long it took. It asserts the status code, not merely that curl succeeded.
5. Count rows in `sources` and require more than zero. A seed step that seeded nothing exits 0 exactly like one that worked, so the count is floored rather than the exit status trusted.
6. Run one worker pass and accept **exactly two** outcomes: exit 0 having printed that it started a pass, or exit 78 having printed why it cannot. Silence, a crash, or an empty success all fail the job.

Locally I proved the part that does not need Docker: the bundled migrate and seed scripts against a real Postgres, run twice. Second run created nothing, and the row counts held at 18 sources and 31 topics.

## Design decisions worth a reviewer's attention

**One image, two services.** `web` and `worker` differ only by entrypoint. One thing to build, one thing to keep patched.

**The runtime image carries no build toolchain.** `output: "standalone"` makes Next emit a minimal server plus only the `node_modules` it traced as reachable. Verified against the installed docs at `node_modules/next/dist/docs/01-app/03-api-reference/05-config/01-next-config-js/output.md`, not from memory: it also confirms `.next/static` must be copied by hand, which is why that line exists, and that the standalone server is started with `node server.js` rather than `next start`.

**The database scripts are bundled, not shipped as TypeScript.** `db:migrate` and `db:seed` run through `tsx`, a dev dependency, and would have dragged the whole tree into the runtime image. esbuild bundles each to self-contained CommonJS — 272 KB and 325 KB — so the runtime needs no `node_modules` of its own beyond what Next traced.

**npm is pinned from `packageManager`, not typed again.** `node:22` ships npm 10, which cannot read this lockfile (#22). The image reads the version out of `package.json` exactly as `ci.yml` does. `engine-strict` is on, so if the base image's Node ever falls below the engines floor, the build stops there with `EBADENGINE` instead of somewhere less obvious.

**Postgres's port is not published.** 5432 is usually already taken by a local Postgres, and a port conflict on first run is a bad first impression. `docker compose exec postgres psql` reaches it.

## Deviations from the ticket, named

- **The ticket says `next start`; the image runs `node server.js`.** Those are in tension: standalone output exists precisely so the runtime does not need the full `next` install that `next start` requires. I took "standalone output" as the binding half, since it is what keeps the image small. The installed docs describe `server.js` as the intended way to start a standalone build.
- **The ticket says seed "if the sources table is empty"; the image seeds on every start.** The seed is already a keyed upsert that never overwrites operational state and never deletes — its own header says so — so the emptiness test only protects against something that cannot happen. Running it always also means a source added to the catalogue reaches an existing deployment, which a first-run-only seed would strand. The smoke job still floors the row count, so "seeded nothing" cannot pass silently.
- **The smoke job accepts two worker outcomes rather than one.** Pinning it to today's exit 78 would mean issue #5 has to remember to come back and edit this job, and a forgotten edit would fail the build for the wrong reason. Both branches require a specific line in the output, so neither can pass on silence.

## Taken on trust

I have no Docker on this machine. Everything about image layers, the healthcheck, the compose dependency ordering and the size number is measured by the CI job and by nothing else — I did not run a container locally, because I could not. What I did verify here: both entrypoint scripts parse under `sh -n`; the worker entrypoint's refusal path, run directly, prints its three lines and exits 78; the compose file and the workflow parse as YAML; the bundled migrate and seed run against a real Postgres and are safe twice; and the four repository checks pass at this head.


zarghammir:
### compose-smoke green — and what the first run caught

[Run 35156636477](https://github.com/zarghammir/ai-radar/actions/runs/35156636477), head SHA `22017b04c093525efbf32ff4eddbd0803842cb80`. Both jobs green: `lint, typecheck, test, build` (11 steps) and `compose-smoke` (all steps, nothing skipped except the on-failure log dump).

The lines the ticket asks for, from the runner's own log:

```
image ai-radar:local = 229 MB (241047418 bytes)
web answered HTTP 200 after 1s
sources rows after web start: 18
worker exit code: 78
RESULT: no worker entry in this image yet (issue #5). Reported, not hidden.
```

The worker's own output above that: `[worker] NOT IMPLEMENTED: this image contains no worker entry.` That is the honest branch, asserted rather than tolerated — the job requires that exact line before it accepts exit 78, so a worker that died silently would fail the build.

**229 MB** for a runtime image carrying a Node base, the Next standalone server and its traced modules, the SQL migrations and two bundled scripts. No build toolchain, no dev dependencies, no `tsx`.

**The catalogue really seeded:** 18 source rows after start, counted from Postgres rather than inferred from an exit code.

### The first attempt failed, and it is worth reading

[Run 35156408153](https://github.com/zarghammir/ai-radar/actions/runs/35156408153) built the image fine and then the web container exited immediately:

```
web-1  | [web] applying migrations
web-1  | Error: Cannot find module '/app/ops/migrate.cjs'
```

esbuild names `--outdir` output `.js` whatever `--format` says. The bundles were `ops/migrate.js` and `ops/seed.js`; both entrypoints asked for `.cjs`. I had verified the bundles worked — I ran them against a real Postgres — but I verified them at the filenames I had chosen by hand in the scratch directory, not at the filenames the Dockerfile's own command produces. The check and the thing being checked had drifted apart, which is the only reason a one-word mistake survived to CI.

Two changes, not one:

1. `--out-extension:.js=.cjs`, so the name matches what the entrypoints already assumed. Worth having anyway: a bare `.js` is read as ESM if any `package.json` above it declares `"type": "module"`, including the one Next writes into the standalone bundle.
2. **An assertion in both stages**, because the real defect was that nothing connected "the path an entrypoint names" to "the file the build produces". The builder now checks `ops/migrate.cjs`, `ops/seed.cjs` and `.next/standalone/server.js` exist, and the runner checks again after the COPYs, since what the builder produced is not automatically what landed in the final image. Both printed in this run:

```
build artefacts present at the paths the entrypoints use
runtime image has every path its entrypoints reference
```

A rename now fails the build with an obvious message instead of the first container start with a stack trace.

zarghammir:
### Added `--once` to the smoke's worker run — a defect this PR was carrying into someone else's ticket

[Run 35156914742](https://github.com/zarghammir/ai-radar/actions/runs/35156914742), head SHA `ddace4987347c4a3dc3a4f53e80532163721f5f0`. Both jobs green, same evidence as before:

```
image ai-radar:local = 229 MB (241047418 bytes)
web answered HTTP 200 after 1s
sources rows after web start: 18
worker exit code: 78
RESULT: no worker entry in this image yet (issue #5). Reported, not hidden.
```

**What was wrong.** I read issue #5's text to check it did not collide with this branch. It does not — it specifies `src/worker/main.ts`, exactly the path the Dockerfile conditionally bundles, so the image picks up a real worker with no change here. But #5 gives that worker two modes, a loop on `INGEST_INTERVAL_MINUTES` and `--once`, and this job ran `docker compose run --rm worker` with no arguments.

Today that is fine: the not-implemented path exits 78 in about a second and ignores arguments. The day the real worker lands, the same command starts **the loop**, and this step stops exiting — it would hang until the 15-minute job timeout. The build would go red on #5's pull request, and it would read as #5 having broken CI when the wrong command is this one.

So this PR was green while carrying a timer set to go off inside somebody else's ticket. That is worse than a plain bug, because the person who has to debug it is not the person who wrote it.

**The fix.** `docker compose run --rm worker --once`. The entrypoint already does `exec node /app/ops/worker.cjs "$@"`, so the flag passes through. The compose `worker` service itself stays argument-free, because a self-hoster running `docker compose up` wants the loop; only the smoke wants a single pass. The workflow carries a comment saying the flag is load-bearing, so the next reader does not tidy it away as noise.

The run above confirms it is a no-op today — same exit 78, same marker line — which is what makes it safe to land now rather than leave for #5.

zarghammir:
REVIEW VERDICT: APPROVE

Reviewed `ddace4987347c4a3dc3a4f53e80532163721f5f0` in a detached worktree of my own. There is no Docker on this machine, so I say up front which claims I could only read: everything about image layers, the 229 MB size, healthcheck timing, compose dependency ordering and the containers actually running comes from the `compose-smoke` job log for this exact SHA, read line by line from the API, not from a container I started. Everything that does not need Docker I reproduced locally: the bundles built with the Dockerfile's own flags and run twice against a scratch database, both entrypoints, the worker's refusal path, the Next documentation the config cites, and the four repository checks.

The ticket's criterion holds as far as it can be shown without Docker: the smoke job is green on this head and its log shows the 200, the seeded count and the worker's summary line. The three named deviations are each justified by something I could verify. One finding is worth fixing in the ticket that first needs it, and it has that owner; nothing blocks.

---

### 1. The runtime image never receives `public/`, so every static asset #12 adds will 404 in Docker while working in dev, and the smoke will not notice — `Dockerfile:72-73` (owner: issue #12)

The Next docs the config cites say it in one sentence (`output.md:36`): the standalone server "does not copy the `public` or `.next/static` folders by default … these folders can be copied … manually, after which `server.js` will serve these automatically." The Dockerfile copies `.next/static` and stops. Today that is harmless: `public/` is not tracked (`git ls-files public` is empty), and the two scaffold images `page.tsx` still references 404 in every environment anyway. Issue #12 adds `manifest.ts`, icons and `public/sw.js`; in the container they will be missing, the PWA will not install, and `compose-smoke` will stay green because it asserts only `GET /` = 200. Fix belongs with #12: track `public/` (a `.gitkeep` or the first real file), add `COPY --from=builder /app/public ./public` beside the static copy, and give the smoke one assertion on a static path (`/manifest.webmanifest` or an icon) so "served by the container" is measured, not assumed. Recorded here so it is not discovered from a phone that will not install the app.

### 2. Advisories, no change requested here

- **`INTERNAL_API_SECRET=change-me` is an active line in `.env.example`**, and the quick start copies it verbatim into a running deployment. No code reads it yet (0 consumers), so nothing is exposed today. Issue #5 adds the route that checks it and must refuse the placeholder (or an empty value) at startup rather than accept it; that is the moment this becomes a self-hoster's open door. Pre-existing from the scaffold, not this PR's.
- **`next build` inside the image fetches Google Fonts** (`layout.tsx` uses `next/font/google`), so an image build on a machine without egress to `fonts.googleapis.com` fails. Same advisory as in the PR #23 verdict; #12's self-hosted fonts remove it.
- **The worker service is argument-free with `restart: "no"`.** Correct for today's refusal. When #5 lands a looping default, `docker compose up` will run the loop and a crash will leave it down; #5 should set the restart policy when it decides the schedule. The smoke's `--once` is the right call now and is a no-op today, as the comment says.
- **`COPY . .` before `npm ci`** in the builder means any file change invalidates the install layer. Correctness is unaffected; build time is. Copying `package.json`, `package-lock.json` and `.npmrc` first would let the install cache. A speed nit for a later pass.

---

1. **Smoke log, per SHA.** Run 35156914742, `pull_request`, `head_sha` `ddace49`, two check-runs on the commit: `lint, typecheck, test, build` success (steps 1–11), `compose-smoke` success. Smoke steps 1–9 success, 10 ("Show the logs if anything above failed") skipped, 11 teardown success. Log lines: builder `v22.23.2` / `11.19.1` / `added 426 packages`; `no worker entry in this build (issue #5)`; `build artefacts present at the paths the entrypoints use`; `runtime image has every path its entrypoints reference`; `image ai-radar:local = 229 MB (241047418 bytes)`; `Container ai-radar-postgres-1 Healthy`, `Container ai-radar-web-1 Healthy`; `web answered HTTP 200 after 1s`; `sources rows after web start: 18`; the three `[worker] NOT IMPLEMENTED …` lines; `worker exit code: 78`; `RESULT: no worker entry in this image yet (issue #5). Reported, not hidden.` The step accepts exit 0 only with `running one ingestion pass` in the output and exit 78 only with the `NOT IMPLEMENTED` marker; any other code, or either code without its line, exits 1. So a silent death (exit 1, or 0 with no output) fails. `--once` is passed by the smoke; `docker-compose.yml` has no `command`/`args` on the worker service, so the service itself is argument-free.

2. **Dockerfile.** Two stages, `builder` and `runner`, both `node:22-alpine`. The runner receives only `.next/standalone`, `.next/static`, `drizzle/`, `ops/` and the two scripts; no `npm ci` runs in it, so no dev dependencies and no `tsx`. npm is read from `packageManager` (`Dockerfile:21`), the same expression `ci.yml` uses; the log shows `11.19.1`. Path assertions exist in both stages (`:57-60`, `:82-86`) and both passed in the log. `output: "standalone"` and `node server.js`: `output.md:24-47` says `.next/standalone` is deployable "without installing `node_modules`", that "a minimal `server.js` file is also output which can be used instead of `next start`", and line 54 that `PORT`/`HOSTNAME` env vars select the bind address, which is exactly what `ENV PORT=3000 HOSTNAME=0.0.0.0` does. The `.next/static` copy is the one the doc prescribes (`:41`). Finding 1 is the other half of that sentence.

3. **Deviations.** *Seed on every start:* `seed.ts` at this head keeps the same conflict `set` as when I proved it in the PR #24 review (catalogue fields only; `enabled`, `lastFetchedAt`, `lastError` absent) and contains no delete; my local double run of the bundled `seed.cjs` gave `18 created` then `0 created, 18 already present`, 31 topics likewise. Safe every start, as argued. *Slow or unreachable database:* `web-entrypoint.sh` is `set -eu`; `migrate.ts` throws without `DATABASE_URL` and `main().catch(...)` calls `process.exit(1)` on any failure; `exec node server.js` is the last line, so the server cannot start unless migrate and seed both exited 0. Compose only starts `web` after Postgres is `service_healthy`, and `--wait` in the smoke fails if web never becomes healthy. Loud, not empty. *`node server.js`:* documented above. *Two worker outcomes:* both branches require a specific line; judged correct for the reason the PR gives.

4. **Security for self-hosters.** No `ports:` on the postgres service; reachable only on the compose network (and by `exec`). Credentials `ai_radar`/`ai_radar` appear only in `docker-compose.yml` and `.env.example`'s `DATABASE_URL`, which the compose `environment:` deliberately overrides to the service hostname. `.dockerignore` excludes `.git`, `.github`, `node_modules`, `.next`, `.env` and `.env.*` (keeping `.env.example`), plus docs, design, fixtures and Markdown. `COPY . .` in the builder therefore cannot bake a real `.env` into a layer, and the runner copies nothing from the context but the two scripts. The runtime runs as the unprivileged `app` user. `env_file: .env` passes the operator's whole `.env` to both containers, which is the intended way secrets reach them and never touches the image.

5. **First attempt.** Run 35156408153 at `7175c3d`: smoke failed at "Boot postgres and web", and the captured logs show `web-1 | Error: Cannot find module '/app/ops/migrate.cjs'`, the entrypoint naming `.cjs` while esbuild wrote `.js`. Fix `22017b0`: `--out-extension:.js=.cjs` plus `ls -la ops`, and the two `test -f` blocks. I rebuilt with the exact flags locally: `migrate.cjs` 272 600 bytes, `seed.cjs` 325 210 bytes, and the CI log shows both assertion blocks passing. A recurrence (esbuild changing its naming, a renamed entry, a COPY that misses) now fails at `docker compose build` with the assertion's message rather than at first container start.

6. **Charter 9 and laptop assumptions.** No routes, no server actions, no new `process.env` reads, no key material. Nothing in the Dockerfile, compose file, entrypoints or `.dockerignore` mentions `/Users/`, darwin, arm64, x86 or a home path; `localhost` appears only in the comment that explains why the compose `DATABASE_URL` overrides it. `--target=node22` matches the base image; `engine-strict` in `.npmrc` is copied into the builder, so a base image below the 22.13 floor would stop at `EBADENGINE` (the log shows 22.23.2). `postgres-js` and `drizzle-orm` are pure JavaScript, so the alpine base needs no native build. The image is built on the runner's architecture; no platform is pinned, which is right for a self-hoster's own machine.

---

## Verified myself

- Head = GitHub head; `origin/main` `77e2842` is an ancestor; 8 files both ways, 336 insertions, 2 deletions, three commits.
- **Four commands** at this head, bare, fresh `npm ci`: `tsc --noEmit` 0, `eslint .` 0, `vitest run` 0 (108 tests; this branch predates #30), `prettier --check .` 0.
- `sh -n` on both entrypoints; `sh docker/worker-entrypoint.sh --once` with no `ops/worker.cjs` present: exit 78, nothing on stdout, exactly the three `[worker]` lines on stderr.
- `esbuild` with the Dockerfile's flags; both bundles run from the worktree root (where `./drizzle` is, as `/app` in the image) with `tsx` absent from the path and no `node_modules` resolution needed, twice, against `ai_radar_cr29` (created and dropped): migrations applied both times, seed 18/31 created then 0 created.
- The Next docs passage, by reading the installed file the config names.
- Both CI runs' steps and the quoted log lines from the API; the fix commit's diff.
- `.env.example`, `seed.ts`, `migrate.ts`, `page.tsx`, `git ls-files public` at this head.
- Worktree clean at `ddace49`, nothing pushed, no comment edited, scratch database dropped, no browser opened.

## Taken on trust (could only read)

- **Every container-level fact:** that the image is 229 MB, that the healthchecks pass in the times shown, that `depends_on … service_healthy` ordered the start, that `docker compose run --rm worker --once` reached the entrypoint with that argument. All from the smoke log, which is the only place this stack has ever run; none from a container I started.
- **`wget` in `node:22-alpine`** for the web healthcheck is BusyBox's; the healthy line in the log is the evidence, not an inspection of the image.
- **The README's three-command path on a real self-hoster's machine** beyond what the smoke reproduces (it runs the same commands on ubuntu-latest with Docker present).
- The lane's statement that the compose file and workflow parse as YAML I did not repeat separately; the smoke job having run is the stronger evidence.

Issue #10, opened 2026-09-16
API contract for the web app and future native clients


zarghammir:
**Problem.** The UI lane needs a stable seam to build against, and the brief asks that native clients can be added later.

**Solution.** First write `docs/api.md` (routes, params, JSON shapes, error shape) and get PM sign-off, then implement route handlers: `GET /api/brief?length=5|10|all` (ranked stories since the user's last brief window, with reading-time estimate and count), `GET /api/radar?type=&topic=&source=&verification=&since=&sort=newest|importance|trending&cursor=`, `GET /api/stories/:slug` (story, items with source name/tier/homepage/url/role/published, timeline, topics, score components, verification note), `POST/DELETE /api/saved/:storyId` + `GET /api/saved`, `POST /api/read/:storyId` and `POST /api/hide/:storyId`, `GET/PUT /api/preferences`, `GET /api/topics`, `GET /api/sources`. Zod-validated input, consistent `{ error: { code, message } }`. No route may trigger ingestion or any external call.

**Expected output.** `docs/api.md` matches the implemented routes. Vitest route tests cover each route's happy path and one validation error. `curl` examples in the doc work against the local app.

Issue #9, opened 2026-09-16
Ranking run and the 'why ranked' data


zarghammir:
**Problem.** Stories have no score, so Today cannot order them and the developer cannot see why.

**Solution.** `rankAllStories(db, now)` after each ingest: for each story active in the last 7 days gather source tiers, topic keys, best HN points/comments, user topic keys from `user_preferences` and call `rankStory`; persist `score` and `scoreComponents`. Expose the components with labels (`COMPONENT_LABELS`) in the story API so the UI can render "Topic match +14, Primary source +20 …". Ranking must be deterministic for a fixed `now`.

**Expected output.** After ranking, a primary-source story from 1 hour ago outranks a single-community-source story from 20 hours ago. The story API returns the component list whose sum equals the score. Test with a deletion control on the topic-match component.

zarghammir:
Ruling carried from PR #30: topic tagging reads only the primary item's title and excerpt (correct: a forum headline must not tag a story). Consequence: a primary-source post with a vague title ("Introducing our new model") gets no company topic. Fix in this ticket: sources carry default topic keys in config (seed: openai-blog → openai, deepmind → google-deepmind, …) and the tagger unions the primary item's source defaults with the keyword matches. Test: a vaguely titled OpenAI blog post is tagged openai; an HN thread linking it adds nothing.

zarghammir:
Rulings before dispatch: (1) rankAllStories calls storySources() and passes that array straight into rankStory; no second list, no mapping. (2) The why-ranked data is the persisted scoreComponents plus COMPONENT_LABELS; do not invent a new shape. (3) Deletion controls in this ticket name the TEST that must redden, not just "the suite goes red": removing the topic-match component must redden the test that asserts a followed-topic story outranks its unfollowed twin; removing the pass-through must redden the test that asserts a single outlet filing twice earns no corroboration.

zarghammir:
Ruling from the API contract review: add a `verificationPenalty` score component so the why-ranked panel can show what pushed a story down: UNVERIFIED −6, EMERGING −2, others absent. Store it like every other component; COMPONENT_LABELS gets "Unverified claim" / "Not yet corroborated". Test: an UNVERIFIED story and an otherwise identical CORROBORATED story differ by exactly the penalty plus the tier difference; deletion control names the test.

PR #30, opened 2026-09-16
Pipeline orchestration: normalize, dedupe, store, cluster into stories, tag topics, verify

### LINKED ISSUE #8: Pipeline orchestration: normalize, dedupe, store, cluster into stories, tag topics, verify
**Problem.** The pure functions exist; nothing writes stories to the database. Ten reports of one announcement would show as ten items.

**Solution.** `src/pipeline/run.ts` exporting `runIngest(db, sourceIds?)`: for each source fetch → `normalizeItem` → insert new `raw_items` (skip on fingerprint conflict) → for each new item `assignStory`: (1) another source's item with the same `canonicalUrl` and a story → join it; (2) else best `isSameStory` match among stories active in the last 72 hours within the same content family (papers only with papers unless URL-matched); (3) else create a story with a unique slug. Set item role (primary for PRIMARY-tier sources, discussion for HN, report otherwise), pick the story's primary item (PRIMARY tier first, then earliest), recompute `sourceCount`, `lastActivityAt`, `verification` + `verificationNote` via `deriveVerification`, and tag `story_topics` by topic keywords over title + excerpt. Everything in one transaction per item. Tests against a real Postgres (CI service) with a deletion control: remove the URL-match rule and the "blog + HN same URL" test must go red.

**Expected output.** After a worker pass, an OpenAI blog post and its Hacker News thread are one story with two sources, verification PRIMARY_SOURCE and a note naming OpenAI. Two different OpenAI stories stay separate. Re-running changes no story ids.

zarghammir:
Closes #8.

Until now the pieces existed but nothing joined them up: items were read, cleaned and scored, and never became stories. Ten reports of one announcement would have shown as ten separate things.

This adds the part that runs the whole loop. It reads every source, stores what is new, and works out which items are about the same development. An OpenAI blog post and the Hacker News thread linking to it now become **one story with two sources**, badged as a primary source and naming OpenAI.

No screens, no schedule, and no ranking run — those are #9 and #10.

## How an item finds its story

1. **Another source already filed the same page.** Same canonical URL means the same thing, so they join. This is what makes a blog post and its forum thread one story.
2. **Otherwise, the best matching headline** among stories still active in the last 72 hours, and only within the same content family, so a paper never absorbs a news report.
3. **Otherwise a new story**, with a slug that does not collide.

Then role, primary item, source count, last activity, verification and topics are all recomputed from the items actually attached, and the whole thing is written in **one transaction per item**, so a failure cannot leave an item with no story or a story whose counts disagree with its items.

## The two rulings

**One de-duplicated source array.** `storySources` returns a single array of `sourceKey`, `sourceName` and `tier`. That exact array is what `deriveVerification` takes, and what `rankStory` will take in #9 — no mapping step in between where the two could drift apart again. A test passes the same array to both, because that seam is where one module once counted items while the other counted outlets.

**The real FetchContext.** Nothing constructed one before, so `ctx.log` had a caller in the Hacker News adapter and no provider: every failure it reported went nowhere. The factory lives in `src/pipeline` so #5's worker imports it instead of assembling a second one. Lines go to stdout tagged with the source key, and a failed source stores its reason and those lines on `ingest_runs.error` and on the source row. A test asserts the Hacker News failure line is actually captured.

Also fixes the PR #28 note nit: one newsroom plus an analyst now reads "Reported by X, analysed by Y" instead of claiming independent reporting.

## Two things the deletion controls found

These are the reason to run controls rather than trust a green suite, and both were live in the branch until they were run.

**The control the ticket names did not work.** Removing the URL rule left the blog plus Hacker News test green. The two headlines were similar enough that the *title* rule joined them anyway, so the test never exercised the rule it exists to cover. The Hacker News headline is now deliberately unlike the blog one, and removing the URL rule turns it red, as specified.

**Nothing asserted the de-duplication.** Removing it from `storySources` turned nothing red, because every test gave each story at most one item per source. There is now a test where one source files two articles about the same launch: it must read as one source and EMERGING, not two sources and CORROBORATED — the exact defect the single array exists to prevent.

### Full control run

Each applied, run, reverted. All compile cleanly, so every red is behavioural.

| removed | result |
|---|---|
| The URL-match rule | 2 red, including blog + Hacker News |
| The 72-hour window | 1 red |
| The content-family check | 1 red |
| Role assignment | 1 red |
| Primary item prefers a first-party source | 1 red |
| Source de-duplication | 1 red |
| Topic tagging | 1 red |
| Storing a failed source's error | 1 red |

## About the database tests

They run against a real Postgres, on **a database of their own** that they create, migrate and drop. Not the configured one: the suite truncates between tests, and pointed at a developer's database it would silently delete their seeded catalogue. `npm test` should not do that.

They also **refuse to skip when `CI` is set**. This matters more than it sounds: the suite skipped all thirteen tests locally, silently, because vitest does not load `.env`, and a suite that skips everything agrees with itself perfectly. It now loads `.env` and says so loudly when it still cannot run.

## Checks

| command | exit |
|---|---|
| `npx tsc --noEmit` | 0 |
| `npx vitest run` | 0 |
| `npx eslint .` | 0 |

108 tests → **125**, in 10 files. No app build and no dev server locally; CI builds.

## Worth raising

- **Content families are two**: papers and everything else, which is exactly what the ticket specifies. If research blog posts should also stay out of news clusters, that is a wider rule and a separate decision.
- **Topic tagging reads the primary item's title and excerpt**, not every item's. Using every item would let a forum headline tag a story with topics the source never mentioned.
- **Slug uniqueness assumes a single writer.** It probes and increments inside the item's transaction, which is correct for one worker. Two workers ingesting at once could collide on the unique index; that is #5's concern to rule on, not something to solve blind here.


zarghammir:

Reviewed `38cffad9ac43e414ec816d153469d414a9addc47` in a detached worktree of my own, with the suite run against a database it creates and drops, and with five scenarios of my own written as a temporary test file against the same harness (deleted afterwards; tree clean). The ticket's three criteria hold, both rulings are implemented, the four deletion controls I reproduced each turn exactly the right tests red, one transaction per item is real (a failure mid-item leaves nothing behind), and a second run changes no story field but `updatedAt`. Nothing is reachable from a route.

Three things block, all in `src/pipeline/run.ts`, all small, all found by running the orchestrator through cases the suite does not cover. The first two are wrong answers in the assignment rule itself; the third drops the reason from a stored failure.

---

### 1. A late-arriving primary item does not take over the story: `refreshStory` recomputes the primary item but never the title, content type or first-seen time — `run.ts:157-192`

Scenario A, my test: Hacker News runs first with "Everyone is talking about the new model" linking to `openai.com/index/gpt-6/`; the OpenAI blog runs next with "Introducing GPT-6" (RELEASE, published two hours earlier) at the same URL. Result after the second run:

```
sourceCount 2, verification PRIMARY_SOURCE, primaryItemIsBlog true          ← correct
story.title       "Everyone is talking about the new model"                  ← still the forum headline
story.contentType "NEWS"                                                     ← still the forum item's type; the primary item is RELEASE
story.firstSeenAt 11:00                                                      ← the HN time; the blog post is 09:00
```

`refreshStory` sets `primaryItemId`, `sourceCount`, `lastActivityAt`, `verification`, `verificationNote`, `updatedAt` and nothing else, so `title`, `contentType` and `firstSeenAt` are frozen at whatever the first item happened to be. Three consequences: the story's face is a forum headline while its primary item is the lab's own post; the ranking content-type weight (#9) will pay NEWS 2 instead of RELEASE 6; and `contentFamily(c.contentType)` at `run.ts:135` is evaluated against the frozen type, so a paper first sighted through a Hacker News link is a "general" story for the rest of its life and will absorb news reports by title, which is exactly what the family rule exists to prevent, one ordering away. This is order-dependent today: sources run in table order, so the seeded primaries usually go first, and the suite's blog-then-HN test never sees it. It stops being latent the first time a thread beats the post to the feed, which is the normal case for a leak.

Fix: in `refreshStory`, after choosing `primaryItem`, also set `title: primaryItem.title`, `contentType: primaryItem.contentType`, and `firstSeenAt: byAge[0].publishedAt` (the earliest item, which is what "first sighted" means once late items can arrive). Add my scenario A as the regression: HN first, blog second, assert title and contentType follow the primary item.

### 2. The candidate query caps at 300 rows with no ORDER BY, so which open stories an item is compared against is arbitrary — `run.ts:127-131`

```ts
const candidates = await tx.select(...).from(stories).where(gte(stories.lastActivityAt, cutoff)).limit(CANDIDATE_LIMIT);
```

`CANDIDATE_LIMIT`'s comment says "most recently active stories", but there is no `orderBy`, so Postgres returns whichever 300 it reaches first. Scenario E, my test: 300 unrelated open stories inserted, then one titled "Introducing GPT-6" inside the window; `assignStory` for an item titled "Introducing GPT-6" returned `created: true` and a new story. The matching story was in the table, open, and missed. At 18 sources over a 72-hour window the open-story count will pass 300 in normal operation, and from then on clustering depends on heap order. Fix: `.orderBy(desc(stories.lastActivityAt))` so the cap means what the comment says, or drop the cap; either way a test that plants more than `CANDIDATE_LIMIT` open stories and still joins.

### 3. A database failure's real cause is not stored: `ingest_runs.error` and `sources.lastError` get drizzle's wrapper text without the Postgres message — `run.ts:284-303`

Scenario B, my test: a trigger on `story_topics` that raises `cr30 boom`. Atomicity is correct: 0 stories, 0 raw items, the run recorded with `itemsFetched 1, itemsNew 0`. But the stored error is

```
Failed query: insert into "story_topics" ("story_id", "topic_id") values ($1, $2) on conflict do nothing
params: 1,1
```

and `cr30 boom` appears nowhere in it. Drizzle raises a `DrizzleQueryError` whose `.message` is the SQL and whose `.cause` carries the database error; `e instanceof Error ? e.message : String(e)` keeps the first and drops the second. So an operator reading `ingest_runs.error` after a real failure (a constraint, a type cast, a disk-full) sees which statement failed and not why. The ruling was that the log reaches `ingest_runs.error` so a failure explains itself; this one does not. Fix: include the cause, e.g. `const msg = e instanceof Error ? (e.cause instanceof Error ? `${e.message}\n${e.cause.message}` : e.message) : String(e)`, and assert the cause text in the failed-source test (a fake 404 does not exercise this path; a trigger or a violated constraint does).

### 4. Advisories, no change requested

- **One bad item stops the rest of that source's items for the run.** The per-item transaction throws out of the `for` loop into the source-level `catch`, so items after the failing one are not tried until the next run (they are not stored, so they will be). Correct enough, but the run's `itemsFetched` will exceed `itemsNew` by more than the one failure and the error names only the first. Worth a line in #5's runbook.
- **`lastFetchedAt` is set on a failed fetch** (`run.ts:302`, asserted by the suite). It then reads "last attempted", and a source failing for a week shows today's date beside its error. `lastError` makes the state visible, so not a defect; if the Settings screen ever shows "last fetched", it should show the last success. #5/#12 to rule.
- **Topic tags come from the primary item's text only.** Stated in the PR; agreed, and with finding 1 fixed the tags will follow the primary item when it changes, which `tagTopics`'s recompute-not-append design already handles.
- **Slug uniqueness assumes one writer**, as the PR says; #5's concern.

---

## Criteria and rulings, reproduced

| | evidence |
|---|---|
| blog + HN → one story, two sources, PRIMARY_SOURCE, note names OpenAI | suite test; control C1 (URL rule removed) turns it red, and only because the HN headline is now unlike the blog's, which I read in the fixture |
| two OpenAI stories stay separate | suite test |
| re-run changes no story ids | suite test, plus scenario C: every story column except `updatedAt` identical across two runs an hour apart |
| one de-duplicated array to both functions | `storySources` builds it, `refreshStory` hands it to `deriveVerification`, the suite hands the same array to `rankStory`; control C2 (de-dupe removed) turns "counts one source once" red |
| real FetchContext; log to stdout with source key; failure into `ingest_runs.error`; HN failure line asserted | `context.ts`; default sink is `console.log`, lines prefixed `[key]`; the suite captures `hackernews-ai … 2 … failed`; finding 3 is about what the stored text omits |
| note wording (PR #28 nit) | `verification.ts:56-58`, "Reported by X, analysed by Y" |

## Extra checks

1. **Assignment rule and controls.** C1 URL rule disabled: 2 red (blog+HN, paper+report by URL); `tsc` reported 2 for this one because the mutation left a constant condition, the other three compiled clean. C2 de-dupe removed: exactly "counts one source once…" red. C3 family check removed: exactly "does not let a paper and a news report cluster on title alone" red. C4 72-hour window removed: exactly the quiet-story test red. The blog+HN test exercises the URL rule and nothing else: its two titles share no distinctive token, so the title rule cannot join them, and C1 proves the URL rule is the only thing that does. **Sibling:** no path joins across families without a URL match in the code as written; the sibling is finding 1, where the *story's* family is frozen at creation, so the family check compares against a type the primary item may no longer have.
2. **Transactions.** Scenario B: trigger-induced failure inside `refreshStory` → 0 stories, 0 raw items, no orphan `storyId`, run and source carry the error. One transaction per item holds. Finding 3 is what that error says.
3. **Re-run stability.** Scenario C as above; `ingest_runs` grows by one row per source per run, as it should.
4. **Primary item and role.** PRIMARY first then earliest: suite test plus scenario A (`primaryItemIsBlog true` even though the blog ran second). HN → `discussion`: suite. Title and content type following the primary item: **finding 1, they do not.**
5. **FetchContext.** Lines carry `[source-key]`; a failing source writes `ingest_runs.error` and `sources.lastError` and the other sources complete (suite, and scenario D). Absence versus failure: scenario D, an empty feed records `itemsFetched 0, error null`; a 404 records `error "… 404 …"`; the two states are distinguishable on the row and in `bySource`.
6. **DB tests.** CI run 35156625592 on this SHA: steps 1–11 all success; Unit tests log shows `DATABASE_URL: ***localhost:5432/ai_radar_test`, `Test Files 10 passed (10)`, `Tests 125 passed (125)`, no `skipped`. Locally: `DATABASE_URL` set → 125/125 on `ai_radar_pipeline_test`, created and dropped (0 left behind); unset → **exit 0 with 14 skipped** and the two-line `!! DATABASE_URL is not set` warning, so it is loud but it does pass, which is what the PR body says; `CI=1` with it unset → exit 1, `refusing to skip the pipeline tests silently`. The suite loads `.env` itself (`import "dotenv/config"`).
7. **Charter 9.** `src/app` untouched, no `route.ts`, no server actions; the only bare `fetch(` outside adapters is `scripts/sources-check.ts`; `process.env` read only for `DATABASE_URL`/`NODE_ENV`; `run.ts` and `context.ts` are imported by nothing but each other and the tests. No key, no external call outside adapters.

## Verified myself

Head = GitHub head; `origin/main` `77e2842` (after #23, #27, #28) is an ancestor; 7 files both ways, 1015 insertions, 9 deletions, four commits. `tsc --noEmit` 0, `eslint .` 0, `prettier --check .` 0, `vitest run` as tabled. Every source file in the diff read in full, the whole test file read. Controls C1–C4 with edit counts asserted and the tree restored; scenarios A–E run on a database of my own (`ai_radar_cr30_test`, dropped). No browser. Worktree clean at `38cffad`, nothing pushed, no comment edited.

## Taken on trust

The lane's four other controls (role assignment, primary prefers first-party, topic tagging, storing a failed source's error) I did not repeat. `isSameStory`/`titleSimilarity` behaviour is PR #1's, not re-derived here. Behaviour at 18 real sources over days (finding 2's volume claim) is reasoned from the seed counts and the 72-hour window, not measured on live data.

Re-review is a new comment. Findings 1–3 are one commit; scenario A and scenario E make good regression tests as written, and the trigger in scenario B is the cheapest way to exercise finding 3.

zarghammir:
## Finding 1 — a late primary item now takes over the story · `0360de0`

Confirmed exactly as described, and the sharp edge is the one you named third.

`refreshStory` recomputed `primaryItemId` and the counts and left `title`, `contentType` and `firstSeenAt` frozen at whatever item happened to arrive first. All three now follow the primary item, with `firstSeenAt` taken as the earliest item across the story, since that is what "first sighted" means once late items can arrive.

**The content type was the one that mattered most.** `assignStory`'s family check reads the *story's* content type, so a paper first sighted through a forum link stayed a "general" story permanently and would absorb news reports by title — the family rule defeated one ordering away. That is now its own regression: a forum link to arXiv, then the paper itself joining by URL and turning the story into a `PAPER`, then a news report with the identical headline being refused.

Your scenario A is in as written: thread in run one, blog at the same URL in run two, asserting title, content type and first-seen all follow the blog. The slug deliberately does not move — it is the permalink, and a story changing its URL when a second source arrives would be worse than a stale slug.

Controls, each reverted after: dropping `title` reddens the takeover test; dropping `firstSeenAt` reddens it; dropping `contentType` reddens **both** the takeover test and the paper test, which is what confirms the family check reads the recomputed value rather than the stored one.

zarghammir:
## Finding 2 — the candidate cap now means what its comment said · `0360de0`

Correct, and the comment was the giveaway: `CANDIDATE_LIMIT` claimed "most recently active stories" while the query had no `ORDER BY`, so it was whichever 300 rows Postgres reached first.

Added `.orderBy(desc(stories.lastActivityAt))` and kept the cap.

Your scenario E is the regression, and I have named it in the code as the control for removing the `orderBy`: `CANDIDATE_LIMIT + 20` unrelated open stories, then a matching headline inserted as the most recently active. With ordering it is found first; without it, it sits past the cap and is missed. Removing the `orderBy` turns exactly that test red and nothing else.

One deliberate detail in the fixture: the matching story is the *most recently active*, not just present. A test that planted it anywhere in the window could pass or fail on heap order, which is the same non-determinism the finding is about.

zarghammir:
## Finding 3 — the database's own words are now stored · `0360de0`

Reproduced before fixing, with your trigger: the stored reason was `Failed query: insert into "story_topics" …` and `cr30 boom` appeared nowhere.

The reason now walks the cause chain rather than reading `.message`, capped at five links and de-duplicated so a driver that repeats itself does not fill the column. Both `ingest_runs.error` and `sources.lastError` carry it.

You were right that the existing test could not have caught this: **a fake 404 never reaches the database**, so it never exercises the wrapping at all. The new test raises inside Postgres from a trigger on `story_topics`, and asserts three things together — nothing half-written (0 stories, 0 items, the run recording `itemsFetched 1, itemsNew 0`), and `cr30 boom` present in the returned result, in `ingest_runs.error`, and in `sources.lastError`. Reverting to `e.message` turns it red.

## Advisories

Left as they are, as ruled. Noting for whoever picks them up: the "one bad item stops the rest of that source's items for this run" point and `lastFetchedAt` reading as "last attempted" both belong to #5's runbook, and the topic-tagging note is already handled now that finding 1 makes the primary item's text move — `tagTopics` recomputes rather than appends, so the tags follow.

## State

| command | exit |
|---|---|
| `npx tsc --noEmit` | 0 |
| `npx vitest run` | 0 |
| `npx eslint .` | 0 |

125 tests → **129**. CI green at `0360de0`: [run 35157662748](https://github.com/zarghammir/ai-radar/actions/runs/35157662748), `129 passed (129)` across 10 files with no `skipped` anywhere in the log, so the database tests genuinely ran against the service container.

READY FOR RE-REVIEW

zarghammir:
REVIEW VERDICT: APPROVE

Re-review of `0360de0c9f06cbcdc0223a889e45d4d0da1b0e8b`, the response to my REQUEST CHANGES (issuecomment-5705422157). One commit, two files: `run.ts` (+37/−1) and `run.test.ts` (+213). All three findings are closed at the code, each has a regression test that is mine or better, and each test goes red when its fix is removed. `origin/main` `77e2842` is still an ancestor. Merge click is warranted.

## The three fixes, verified by deletion

| finding | fix in `run.ts` | regression test | control | result |
|---|---|---|---|---|
| 1. late primary did not take over | `refreshStory` now sets `title`, `contentType`, `firstSeenAt` (earliest item) from the items; slug deliberately fixed as the permalink | "lets a blog post arriving after the forum thread take over the story" (HN first, blog second at the same URL: title, RELEASE, firstSeenAt = blog time, slug unchanged) and "makes a paper first seen through a forum link a paper story" (story becomes PAPER, then a same-headline Verge item is refused: 2 stories, NEWS + PAPER) | removed the `contentType: primaryItem.contentType` line | **exactly both red**, tsc 0 |
| 2. candidate cap without ORDER BY | `.orderBy(desc(stories.lastActivityAt))` before `.limit(CANDIDATE_LIMIT)` | "compares against the most recently active stories when there are more than the cap": 320 filler stories inserted first, the match inserted last and most recently active | removed the `orderBy`: red on **three consecutive runs** (heap order on a freshly truncated table is insertion order, so the match sits past row 300 every time); replaced with ascending order: **also red** | the test discriminates DESC from none and from ASC; tsc 0 |
| 3. cause dropped from the stored error | `describeError` walks `.cause` up to five levels, de-duplicating messages | "stores the database's own reason when a write fails": a `RAISE EXCEPTION 'cr30 boom'` trigger on `story_topics`; asserts the text in `bySource[0].error`, `ingest_runs.error` and `sources.lastError`, and that nothing was half-written (`itemsFetched 1, itemsNew 0`) | reverted to `e.message` | **exactly that test red**, tsc 0 |

On the PM's question about test 2: the fixture makes the match the most recently active on purpose, and that is what makes the test prove the ordering rather than luck. Unordered, the seq scan returns the 320 fillers first and the cap drops the match (three of three runs). Ascending, the match is last again and dropped. Only descending finds it. The test therefore pins the exact clause, not merely "some ordering".

## Verified myself

- Head = GitHub head; one commit since `38cffad`; `origin/main` is an ancestor; the whole `run.ts` diff and every added test read.
- **Four commands** at this head, bare, `DATABASE_URL` set: `tsc --noEmit` 0, `eslint .` 0, `prettier --check .` 0, `vitest run` 0, **10 files, 129 tests**, none gated.
- **CI per SHA.** Check-run on `0360de0`: one check, success; run 35157662748, `pull_request`, attempt 1, every step success; Unit tests log `Test Files 10 passed (10)`, `Tests 129 passed (129)`, no `skipped`.
- Controls C1, C2 (×3), C2-asc, C3 with the edit count asserted each time and `run.ts` restored; `git status` clean; the suite's own test database dropped after every run (0 left).

## Carried forward

The advisories from the first verdict stand and need no change here: one failing item stops the rest of that source's items for the run (retried next run); `lastFetchedAt` is set on failure; topics come from the primary item's text; slug uniqueness assumes one writer (#5). The `firstSeenAt` recompute uses the earliest item's `publishedAt`, which is the right reading of "first sighted" once late items can arrive; the slug staying put is the right reading of "permalink".

## Taken on trust

As before: the lane's own other controls; `isSameStory` behaviour from PR #1; the volume claim behind finding 2 is reasoned from seed counts, not live data. Nothing new.

Issue #17, opened 2026-09-16
End-to-end harness: Playwright, axe, screenshots, CI job


zarghammir:
**Problem.** Nothing proves the app works in a browser. The reviewer needs instruments, and PRs need screenshots.

**Solution.** Playwright config with projects for 390 (mobile) and 1440 (desktop), light and dark via `colorScheme`, seeded test database from fixtures, `@axe-core/playwright` on every page, a `npm run screenshots` script writing `docs/screenshots/<area>/` in a fixed naming scheme so lanes attach them to PRs, and a CI job `e2e` after build. Floor assertions: at least N stories rendered, so an empty page can never pass.

**Expected output.** `npm run test:e2e` green locally and in CI with the executed test count printed. Deleting the seed makes the floor assertion fail (documented as the control). Screenshot script produces both widths and both themes for every route.

zarghammir:
**Two committed scripts are now waiting for this ticket to give them a consumer.**

The UI lane checked `.github/workflows/*.yml` rather than assuming: **neither `npm run a11y` nor `npm run verify:shell` runs in CI today.** CI runs lint, typecheck, unit tests, build, format check and the compose smoke — nothing else. So the browser evidence on this project exists only when a person runs it by hand, which means it is absent exactly when nobody is looking.

Both scripts are in the repo as of PR #31 and are what this harness should wire:
- `npm run a11y` — the 24-state axe, horizontal-overflow and bottom-bar sweep. It now carries a floor: the bar must be FOUND in all 12 phone states and ABSENT in all 12 laptop states, exactly 6 tabs per bar, labels equal to tabs, and it exits 1 if it measured nothing.
- `npm run verify:shell` — the four checks axe cannot see: the theme radio group's keyboard behaviour, the theme-color meta actually in force, and the service worker's error-caching gate with its never-visited-URL control.
- Both resolve a browser through `scripts/lib/browser.mjs`: `$PW_EXECUTABLE`, else the installed Chrome channel, else **exit 2**. `playwright-core` is declared and ships no browsers, so `npm ci` is unaffected.

**Requirement for whoever builds this job — exit 2 must fail loudly.**

Exit 2 means *the instrument could not run*, which is a different state from *the checks failed* (exit 1) and from *the checks passed* (exit 0). A CI job that treats "no browser available" as a skip produces a green tick for a run that measured nothing — the failure this repo has already catalogued more than once. Three states, three outcomes, and the middle one is not a pass.

[...]

PR #31, opened 2026-09-16
App shell, theming and PWA installability

### LINKED ISSUE #12: App shell, theming and PWA installability

zarghammir:
Closes #12.

## What it is in plain words

The app had no navigation, no theme and could not be installed. It now has all three.

Open it and there is a proper app around the content: a sidebar on a laptop, a bar along the bottom on a phone, and six places to go. Pick light, dark, or follow your device, and it remembers. Add it to your home screen and it opens like a real app, with its own icon and no browser chrome. Lose signal and you get a calm page that says you are offline, instead of the browser's error screen.

There is still no news in it. That arrives with the ingestion work. Every screen says so honestly rather than showing a fake loading state forever.

## What a person can now do

- **Move around.** Today, Radar, Research, Releases, Saved and Settings, from a sidebar on a laptop or a bottom bar on a phone. All six are reachable at both sizes — at review time this sentence was false, because the bar carried four tabs and the other two lived in a sidebar that is hidden on a phone, so they were URL-only. The bar now carries all six, measured at 390px. The one you are on is marked with an orange bar **and** a heavier label, so it never depends on colour alone.
- **Choose how it looks.** System, Light or Dark in Settings. The choice sticks, and there is no white flash when you open it in the dark, because the theme is decided before the page is drawn.
- **Install it.** On Android or desktop Chrome an Install button appears. On an iPhone, where Safari offers no such button, it tells you to tap Share and then Add to Home Screen — and only when you are not already inside the installed app.
- **Use it offline.** Once opened, the shell is cached. Airplane mode gives you a real page that explains the situation and offers to try again.
- **Read what the grades mean.** Settings carries the full explanation of Primary source, Corroborated, Emerging and Unverified, with the four-bar meter beside each.

## Screenshots

> The repository is private, so GitHub will not render images inline in a PR body. All seven are committed under `docs/screenshots/shell/` and linked here.

| Screen | |
|---|---|
| **Today, laptop, dark** — the sidebar, the sprocket rail, the honest empty state | [open](https://github.com/zarghammir/ai-radar/blob/c55060238587752fa4d4f439ae21675c00aebeac/docs/screenshots/shell/today-laptop-dark.png) |
| **Today, phone, dark** — bottom bar, orange marker over the active tab | [open](https://github.com/zarghammir/ai-radar/blob/c55060238587752fa4d4f439ae21675c00aebeac/docs/screenshots/shell/today-phone-dark.png) |
| **Today, laptop, light** — the same world in daylight | [open](https://github.com/zarghammir/ai-radar/blob/c55060238587752fa4d4f439ae21675c00aebeac/docs/screenshots/shell/today-laptop-light.png) |
| **Today, phone, light** | [open](https://github.com/zarghammir/ai-radar/blob/c55060238587752fa4d4f439ae21675c00aebeac/docs/screenshots/shell/today-phone-light.png) |
| **Settings, phone, dark** — theme toggle and the grade key on paper cards | [open](https://github.com/zarghammir/ai-radar/blob/c55060238587752fa4d4f439ae21675c00aebeac/docs/screenshots/shell/settings-phone-dark.png) |
| **Settings, laptop, light** | [open](https://github.com/zarghammir/ai-radar/blob/c55060238587752fa4d4f439ae21675c00aebeac/docs/screenshots/shell/settings-laptop-light.png) |
| **Offline, phone, dark** — what you get with no network | [open](https://github.com/zarghammir/ai-radar/blob/c55060238587752fa4d4f439ae21675c00aebeac/docs/screenshots/shell/offline-phone-dark.png) |

## How to verify

```bash
npm ci
npx next build          # exit 0
npx next start -p 3210
npm run a11y            # exit 0 — axe + overflow + breakpoints, 24 states
```

**Checks at `c55060238587752fa4d4f439ae21675c00aebeac`, each run bare:**

| command | exit |
|---|---|
| `npx tsc --noEmit` | 0 |
| `npx vitest run` (108 tests) | 0 |
| `npx eslint .` | 0 |
| `npx next build` | 0 |
| `npm run a11y` | 0 |

**axe, on the production build, 6 routes x 2 breakpoints x 2 themes = 24 states:**

| | measured |
|---|---|
| serious or critical violations | **0** |
| violations of any impact | **0** |
| horizontal overflow | **0** |
| sidebar present | laptop only |
| bottom bar present | phone only |

**The five binding advisories from the PR #23 review, each verified rather than asserted:**

1. **Self-hosted fonts.** `grep -rl "fonts.googleapis.com\|fonts.gstatic.com" .next/server .next/static` returns nothing, and the build emits 12 `woff2` files from our own origin. No reader's browser contacts Google.
2. **Real breakpoints.** The sidebar is `hidden lg:flex`, the bottom bar `lg:hidden`. The audit confirms which nav is present at each width. Nothing switches on a JavaScript-toggled class.
3. **Never `--org-text` on paper in dark.** The token is defined for the bench only and is not used on any paper surface. axe reports zero contrast violations in all 24 states.
4. **Separators are not text.** The rail and the tab markers are `aria-hidden` spans and pseudo-elements; nothing decorative is announced.
5. **Tokens only.** No hard-coded colours; the prototype's one `#6e6a61` is now a `--meta` token. shadcn's semantic tokens are mapped onto the AI Radar tokens so a component added later cannot introduce a second palette.

**PWA behaviour, driven in a browser:**

- Manifest served at `/manifest.webmanifest`, every value from `brand.ts`, `display: standalone`, icons 192 / 512 / maskable 512 / Apple touch 180.
- Service worker registers, and an offline navigation returns the offline page with **status 200** rather than a browser error. At review time this held only for URLs never successfully requested: the worker cached non-OK responses, so a 404 you had visited was served back offline as a cached 404. It is gated on `fresh.ok` now, and both cases are verified — the previously-404 URL and a never-visited one each serve `offline.html` at 200.
- Theme survives a reload; at `DOMContentLoaded` the dark class, the bench background and the `theme-color` meta all agree. At review time that agreement was measured on the wrong element: the layout declares media-scoped metas and `querySelector` always returned the light one, so the write landed on a meta the browser was ignoring. Every `theme-color` meta is written now, and the check reads the one `matchMedia` says is **in force** — OS dark, theme set to Light, meta in force `#e4e3de`, body `rgb(228,227,222)`.

## Two things I want the reviewer to look at

**axe caught two real defects and I fixed them.** The grade list nested `<dt>`/`<dd>` two levels deep inside `<dl>`, and the Settings note used bench colours on a paper surface, which fails AA in dark. The second is exactly the pairing palette law two exists to prevent, which suggests the law needs a lint rule and not just a sentence.

**I added `axe-core` as a devDependency**, so the accessibility claim is reproducible by anyone via `npm run a11y` instead of being something I ran once by hand. Say the word if you would rather it were not a dependency.

## Not verified, and not claimed

- **Lighthouse on a preview URL, and a real iPhone install.** Both need a deployed HTTPS origin; this was verified against a local production build. The manifest, icons, standalone display and offline fallback are all in place for it, but I have not seen it on a phone.
- **Keyboard order by hand.** Landmarks, the skip link, `aria-current` and visible focus rings are present and axe is clean, but I did not tab through every screen myself.


zarghammir:

Reviewed `feat/app-shell` at `c550602` against `origin/main` at `911cc21`. 45 files (`gh pr diff --name-only` = 45, `git diff --name-only <merge-base>..HEAD` = 45 — they agree). I did not write any of this code and reproduced every claim below myself rather than reading the PR's pasted output.

The craft here is high: 36 states came back with zero axe violations, zero horizontal overflow and a clean console, and the palette discipline holds everywhere I looked. The changes below are about things that only appear when you boot it, merge it, or take it offline.

---

## Blocking

### 1. The branch does not merge into current `main` — `next.config.ts`
`origin/main` is **10 commits ahead** of this branch's merge-base, and both sides rewrote the same line.

- `git merge-tree --write-tree origin/main HEAD` → **exit 1**, conflict staged at `next.config.ts` (stages 1/2/3 all present).
- GitHub agrees independently: `mergeable = CONFLICTING`, `mergeStateStatus = DIRTY`.

`main` now sets `output: "standalone"` (PR #29, what lets the Docker runtime image ship without a toolchain). This PR replaces the same object with a `headers()` block. **Both must survive the resolution.** A resolution that keeps only `headers()` silently breaks the self-hoster Docker image that #29 just landed, and nothing in CI would catch it — the smoke job lives on `main`, not here.

Please rebase on `911cc21` and re-run. Note that CI's green tick was computed against a merge with an *older* `main`, so it is not evidence about the tree that would actually land.

### 2. The service worker caches error responses — `public/sw.js:58`
```js
const fresh = await fetch(request);
const cache = await caches.open(CACHE_VERSION);
cache.put(request, fresh.clone());   // no status check
```
The static-asset branch 25 lines below gets this right (`if (response && response.status === 200)`); the navigation branch does not. Any non-OK navigation response becomes the permanent cached copy of that URL.

**Proved end to end, with a control:**

| step | result |
|---|---|
| SW active and controlling the page | `true` |
| visit `/definitely-not-a-page` (server really answers 404) | live status `404` |
| inspect `caches.open("ai-radar-shell-v1")` | the 404 **is stored**, `cachedStatus: 404` |
| go offline, revisit that URL | served the **cached 404** — “This page could not be found”, status 404 |
| **control:** offline, visit a URL never visited before | correctly served `offline.html`, status **200**, title “Offline · AI Radar” |

The control is the point: the offline path itself works perfectly. The only thing that breaks it is the poisoned entry. This also contradicts the PR body's "an offline navigation returns the offline page with status 200 rather than a browser error" — that holds only for URLs you never successfully requested. A 500 served during a deploy becomes the offline copy of `/` until `CACHE_VERSION` is bumped.

Gate on `fresh.ok`.

### 3. `/research` and `/releases` cannot be reached below `lg` — `src/components/app-nav.tsx:77`
`BottomNav` filters to four tabs. The sidebar that holds the other two is `display:none` below `lg`. Measured at 390px on every route a phone user can actually get to:

| from | visible links on page | visible link to `/research` | to `/releases` |
|---|---|---|---|
| `/` | 5 | 0 | 0 |
| `/radar` | 5 | 0 | 0 |
| `/saved` | 5 | 0 | 0 |
| `/settings` | 5 | 0 | 0 |

Five visible links = skip link + four tabs. The anchors exist in the DOM but sit inside the hidden sidebar, so they are not clickable and not in the tab order either.

Two comments in the code assert the opposite, and **both are false**:
- `src/config/nav.ts:49` — *"Research and Releases live in `More`."* The only occurrence of the string "More" in `src/` is that comment. There is no More tab.
- `src/components/app-nav.tsx:73` — *"Research and Releases are reachable from Radar."* `src/app/radar/page.tsx` renders a `PageShell` and an `EmptyState` and contains no links at all.

Across all 52 files in `src/`, the only references to `/research` and `/releases` are their own entries in `nav.ts`.

This also makes the PR body's first bullet wrong: it promises all six surfaces "from a sidebar on a laptop **or a bottom bar on a phone**." On a phone, two of the six are URL-only.

Related, same root: `src/config/nav.ts:50` exports `primaryNav` and **nothing imports it** — `app-nav.tsx:77` re-types the same four-href literal. Editing the obvious place has no effect.

### 4. The theme-color meta is written to the wrong element — `src/components/theme-script.tsx:22` and `src/components/theme-toggle.tsx:25`
`document.querySelector('meta[name="theme-color"]')` returns the **first** match, but `src/app/layout.tsx:60-63` declares two *media-scoped* metas, so the first one is the `(prefers-color-scheme: light)` one regardless of the active theme. The write lands on a meta the browser is ignoring.

Measured with the OS set to dark, then clicking **Light**:

```
meta media=(prefers-color-scheme: light)  content=#e4e3de  applies=false
meta media=(prefers-color-scheme: dark)   content=#0b0c0e  applies=true   <- the one in force
app bench (computed body background)      rgb(228,227,222) = #e4e3de
browser chrome actually uses              #0b0c0e
```

The app is light, the browser chrome stays dark. This directly contradicts the PR body's "at `DOMContentLoaded` the dark class, the bench background and the `theme-color` meta all agree." It is the same defect in two places — please fix both, selecting by the media attribute that currently matches (or setting both metas).

---

## Should fix before merge

### 5. The theme control is not a working radio group — `src/components/theme-toggle.tsx:72-92`
It declares `role="radiogroup"` with three `role="radio"` children, which promises the ARIA radio pattern: one tab stop, arrow keys to move and select. Measured:

- All three radios have **no `tabindex`**, so each is its own tab stop — tabbing through `/settings` hits System, Light and Dark as stops 9, 10 and 11.
- Focusing "System" and pressing **ArrowRight**: focus stays on System, `aria-checked` stays `['true','false','false']`. Nothing happens.

A screen-reader user is told they are in a radio group ("System, 1 of 3") and reaches for the arrow keys, which are dead. axe cannot see this — it does not exercise keyboard interaction, which is exactly why all 36 states are clean. This is the gap the PR flags under "Not verified": *"Keyboard order by hand … I did not tab through every screen myself."* I did, and this is what it turns up.

Either implement roving `tabindex` + arrow handling, or drop to three plain toggle buttons and stop promising the pattern.

### 6. `npm run a11y` cannot run on a clean clone — `scripts/a11y-audit.mjs:33,42`
```js
(process.env.PW_PATH || "/opt/homebrew/lib/node_modules/@playwright/cli") + "/package.json"
const root = join(homedir(), "Library/Caches/ms-playwright");
```
Hard-coded to one machine's Homebrew prefix and macOS cache path, and **neither `playwright` nor `playwright-core` is declared** in `dependencies` or `devDependencies` (of 29 deps, the only match is `axe-core`; both `require.resolve("playwright")` and `require.resolve("playwright-core")` fail in this worktree). The PR's "How to verify" block tells a reader to run `npm ci` then `npm run a11y` — that sequence cannot work for anyone else.

This matters more than usual here: `npm run a11y` is the PR's *only* accessibility evidence, and this is a project whose premise is that strangers self-host it. Declare the dependency, or document it as a maintainer-only script and stop listing it as the reproduction path.

### 7. Settings announces the wrong page to screen readers — `src/app/settings/loading.tsx:18`
`<span className="sr-only">Loading Live Radar</span>` on the Settings route. The file is **byte-identical** to `src/app/radar/loading.tsx` (md5 `c1ecde8c9278d7c3cbae29d125d113d0` for both).

I swept the whole family — the other four are correct, so this is one slip, not a pattern:

| file | announces |
|---|---|
| `app/loading.tsx` | Loading today's brief |
| `app/radar/loading.tsx` | Loading Live Radar |
| `app/releases/loading.tsx` | Loading Releases |
| `app/research/loading.tsx` | Loading Research |
| `app/saved/loading.tsx` | Loading Saved |
| **`app/settings/loading.tsx`** | **Loading Live Radar** ← |

### 8. The Today eyebrow is a hard-coded weekday — `src/app/page.tsx:6`
`eyebrow="Wednesday morning"`. It is the only weekday string in `src/`, and it is right today — 2026-09-16 is a Wednesday — which is presumably why it survived review and screenshots. It is wrong the other six days. The PR's own framing is that "every screen says so honestly"; this is the one screen that doesn't.

---

## Minor

9. **`cache.put` can be killed before it lands** — `public/sw.js:58` and `:83`. Neither put is awaited or wrapped in `event.waitUntil`, so once `respondWith` resolves the worker may be terminated with the write outstanding. Low frequency, but it makes both the offline copy and the stale-while-revalidate refresh best-effort rather than reliable.
10. **Cross-tab theme sync is inert** — `src/components/theme-toggle.tsx:45`. The `storage` listener calls `onChange`, but `getSnapshot` reads `document.documentElement.dataset.theme`, which another tab's write never touches, and nothing calls `applyTheme`. Two tabs, changed to Dark in tab A: tab B's `localStorage` becomes `"dark"` while its `dataset.theme` stays `"system"`, `.dark` stays off and its background stays `rgb(228,227,222)`. The listener implies a feature that does not happen. `install-prompt.tsx` gets this right.
11. **`src/app/manifest.ts:17-18`** pins `background_color` and `theme_color` to the light bench, so a dark-mode install flashes light on the splash — the exact flash `ThemeScript` exists to prevent. I checked before filing this: the manifest spec has no media-scoped variant, so this is a forced choice rather than a bug. Worth a comment saying so, and worth asking whether `DESIGN.md`'s "dark is this world's native ground" argues for the dark value.
12. **`.playwright-cli/` is not in `.gitignore`.** Running the audit tooling inside the repo leaves an untracked directory. One line to fix.

---

## Raised, not blocking: the bottom bar and the home indicator

`src/app/layout.tsx:85` pads `main` with a flat `pb-20` (measured: `padding-bottom: 80px`), while `BottomNav` is `fixed` and adds `pb-[env(safe-area-inset-bottom)]` on top of its own height (measured: **62.25px**, inset resolving to `0px` in Chromium).

At 80px vs 62.25px there is 18px of clearance and **no clipping in any browser I can test** — I checked at 390px and found zero overflow. But the bar's height grows with the safe-area inset and `main`'s padding does not, so on a home-indicator iPhone (inset ≈34px) the arithmetic gives 96px of bar against 80px of padding — about 16px of content under the bar.

**I could not verify this on real hardware, and I am not claiming it as a defect.** Headless Chromium reports the inset as `0`, so the PR's audit cannot see it either. Making the padding `calc(5rem + env(safe-area-inset-bottom))` would close it without needing a device to prove it.

---

## The two checks you asked for specifically

**Palette laws (#32) — clean, and I swept for siblings rather than trusting the one fix.** There are exactly two paper surfaces in `src/` (`app/settings/page.tsx:26`, `components/install-prompt.tsx:94`) and both take paper ink (`text-ink`, `text-soft`, `border-faint-2`). Every `text-ash` / `text-ash-hi` usage is in `app-nav.tsx` or `page-shell.tsx`, both on the bench, where those tokens belong. `ThemeToggle`, which renders *inside* a paper section, uses only paper tokens. The comment at `settings/page.tsx:86-87` claims the fix and the code performs it. Zero contrast violations in all 36 states.

One observation rather than a finding: the law as written in `globals.css:81` and `DESIGN.md` §1 only forbids **`--org-text`** on paper, but the defect that produced #32 was **`--ash`/`--edge`** on paper. The written rule does not cover the case that actually bit. Your instinct in the PR body that this wants a lint rule is right — and the rule should be "no bench token on a paper surface", not just the orange one.

**Two sources of truth for the theme — agreed, and deliberate.** `localStorage` is what the pre-paint script in `<head>` reads before the first frame; `document.documentElement.dataset.theme` is the runtime source `useSyncExternalStore` subscribes to. They serve different moments and neither can be removed without either reintroducing the white flash or adding a React state copy that can disagree. **Not filed as a finding.** The comment at `theme-toggle.tsx:29-34` explains it correctly.

---

## Verified myself

- Base currency and merge state: 10 commits behind; `git merge-tree` exit 1; GitHub `CONFLICTING`/`DIRTY`.
- File count both ways: 45 = 45.
- CI for **this exact SHA** (`c550602`, run `35157103743`): every step executed — Lint, Typecheck, Unit tests, Build, Format check all `success`. Not a run that ran nothing.
- Four light checks, run bare, exit codes: `npx tsc --noEmit` **0**, `npx eslint .` **0**, `npx vitest run` **0** (9 files, **108 tests actually executed**), `npx prettier --check .` **0**.
- **36 states** (6 routes × 1440/1024/390 × light/dark) with my own axe-core run: **0 violations of any impact**, **0 horizontal overflow** (`scrollWidth − clientWidth`), **0 console errors**, **0 page errors**.
- Breakpoints are real CSS: sidebar present at 1440 and 1024 only, fixed bar at 390 only, `.dark` tracked the emulated OS scheme in all 36 states.
- Focus is visible: orange ring `rgb(240,83,28) 0 0 0 2px` on focus, `box-shadow: none` on the same element blurred (control run).
- Tab order on `/settings`: skip link → brand → six nav links → three theme buttons. Skip link is first and reachable.
- No external calls: **150 requests** across all six routes, **zero non-localhost origins**. Fonts are served from our own origin.
- Service worker registers and reaches `activated`; `/offline.html` serves **200**.
- The server I reviewed against is genuinely this commit: process cwd is this worktree, tree clean at `c550602`, no tracked file newer than `.next/BUILD_ID` (114 files checked, with a control proving the check can detect a newer file), no `.next` file touched since boot.
- Every finding above reproduced with my own script; every deletion/control check noted inline.

## Taken on trust

- **Real-device behaviour.** No iPhone, no Android, no HTTPS origin: the install flow, the Apple touch icon, the splash screen, the home-indicator inset and Lighthouse are all unverified by me. The PR is explicit about this and I am not disputing it.
- **The PR's screenshots.** I did not diff them against my own renders; I measured the live pages instead.
- **The seven committed screenshot files** are linked from the PR body but I did not verify each image matches its caption.
- **`npm run a11y` exit 0 on the author's machine.** I could not run it here (finding 6), so I wrote my own audit instead of reproducing theirs. My 36-state run is broader than their 24 and agrees: zero violations.
- **Icon files** (192/512/maskable/apple-touch) exist and are referenced correctly; I did not open them to confirm they are the intended artwork or correctly maskable.

---

Nothing here is a disagreement about taste, and items 5–8 are all small. Items 1–4 are the ones that would reach a reader: a branch that will not merge, an offline page that can be replaced by a cached error, two of six destinations unreachable on a phone, and browser chrome that disagrees with the app. Happy to re-review as soon as they are addressed — that will be a new comment, not an edit of this one.

zarghammir:
All twelve findings fixed at `6902725882ca3de66498bbfa62bd699ebdbe21fa`, rebased onto `911cc21`. Every claim below was measured against a production build, not read off the code.

## Read this before the numbers: the instrument was broken twice

The bar gate I added to measure the six-tab bar **crashed on its first state, twice**, and both times printed the exact value my runner was checking for.

1. `report.bottomBar` was never initialised — that line sat in an edit that aborted on a bad anchor, and I did not re-add it when I rewrote the patch. The audit threw immediately and wrote zero bytes of JSON.
2. The floor referenced `SCREENS`; the constant is `ROUTES`. A second crash, inside the floor that exists to stop the gate passing on an empty set.

The control run exits 1 when the bar clips, so my check was "control exit=1, want 1". **Both crashes produced exit 1.** Read as success — which is exactly what that output invites — I would have reported a proven gate that had never executed a single assertion, inside the change whose entire purpose is to prevent that. It was caught only because the check asks *what* reddened (floor still passing **and** clipped labels non-zero) rather than *that* something did.

Then the control turned out to be too weak even once it ran: at 280px the labels genuinely still fit, clipped 0, so the gate was still unproven. It forces the condition at 200px, where 12 of 12 states clip and the samples name the text.

I am putting this first because "the bar does not clip" is only worth reading if you know the instrument that measured it was broken, found, and strengthened before it was believed.

## 1. The bar at 390 — it does not clip

Measured on the boxes that actually clip — the `<nav>` and each label — against `document.documentElement.clientWidth`, which excludes the scrollbar.

| | measured |
|---|---|
| phone states with the bar | 12 of 12 |
| laptop states with the bar | 0 of 12 (it is `lg:hidden`) |
| tabs per bar / labels per bar | 6 / 6 |
| labels clipped | **0** |
| nav overflow | **0** |
| narrowest tab | 65px |
| shortest tap target | 59.8px |
| every label box | 61px wide, 0px of clipped text |

**The control, run first:** at 200px the same gate exits 1 with the floor still passing and 12 of 12 states clipped — `Research` clipped by 12px in a 29.3px box. A gate that has only ever passed is not known to be able to fail.

**The floor**, because a check whose input can be empty passes hardest when the thing it guards is most broken: the bar must be found in every phone state and absent in every laptop state, exactly 6 tabs, labels equal to tabs. A state with no bar records `{ present: false }` rather than contributing nothing. The counts print on success so you can see the instrument had something to measure.

## 2. No drift from the `lib/browser.mjs` extraction

Both scripts now resolve a browser through one shared module. Compared against the baseline recorded at `c550602`, before it existed:

| | baseline | now |
|---|---|---|
| states audited | 24 | 24 |
| axe serious/critical | 0 | 0 |
| axe any impact | 0 | 0 |
| horizontal overflow | 0 | 0 |

Per-route anchor counts **do** differ, and should: the bar went from four tabs to six, so every phone route gained two links. That is the fix, not drift.

## 3. The four things axe cannot see

- **Radio group (finding 5).** One tab stop. ArrowRight moved focus System→Light **and** moved `aria-checked`; End moved to Dark. Both were dead before.
- **theme-color (finding 4).** OS dark, theme set to Light: the meta `matchMedia` reports as **in force** reads `#e4e3de`, and the body computes `rgb(228,227,222)`. They agree. Previously the write landed on a meta the browser ignored.
- **Service worker (finding 2).** Worker controlling: true. The 404 probe really returned 404 — asserted, because if the server had answered 200 the probe would never have touched the error path and a green result would mean nothing. The 404 was **not** stored. Offline, that URL now serves `offline.html` at 200, and so does a never-visited URL. Both ends of the reviewer's own control.
- **Floors on all three.** With no worker in control nothing is cached at all, so "the 404 was not cached" would come back true while proving nothing. The script exits 1 if it measured nothing.

## 4. One observation, not a defect

There are **three** `theme-color` metas — two light, one dark — where I expected two. Exactly one applies and its value is correct, so the behaviour is right and I have not touched it: the fix under review is *which* meta gets written, not *how many* are declared. Flagging it so the reviewer does not spend a pass rediscovering it.

## 5. Screenshots

> The repository is private, so GitHub will not render these inline. All nine are under `docs/screenshots/shell/`.

| | |
|---|---|
| **Research at 390, dark** — one of the two pages that were unreachable on a phone. Six tabs, reachable, nothing clipped | [open](https://github.com/zarghammir/ai-radar/blob/6902725882ca3de66498bbfa62bd699ebdbe21fa/docs/screenshots/shell/research-phone-dark.png) |
| **Releases at 390, light** — the other one. These two are the evidence the fix works, not decoration | [open](https://github.com/zarghammir/ai-radar/blob/6902725882ca3de66498bbfa62bd699ebdbe21fa/docs/screenshots/shell/releases-phone-light.png) |
| Today at 390, dark / light | [dark](https://github.com/zarghammir/ai-radar/blob/6902725882ca3de66498bbfa62bd699ebdbe21fa/docs/screenshots/shell/today-phone-dark.png) · [light](https://github.com/zarghammir/ai-radar/blob/6902725882ca3de66498bbfa62bd699ebdbe21fa/docs/screenshots/shell/today-phone-light.png) |
| Today at 1440, dark / light | [dark](https://github.com/zarghammir/ai-radar/blob/6902725882ca3de66498bbfa62bd699ebdbe21fa/docs/screenshots/shell/today-laptop-dark.png) · [light](https://github.com/zarghammir/ai-radar/blob/6902725882ca3de66498bbfa62bd699ebdbe21fa/docs/screenshots/shell/today-laptop-light.png) |
| Settings at 390 dark / 1440 light | [phone](https://github.com/zarghammir/ai-radar/blob/6902725882ca3de66498bbfa62bd699ebdbe21fa/docs/screenshots/shell/settings-phone-dark.png) · [laptop](https://github.com/zarghammir/ai-radar/blob/6902725882ca3de66498bbfa62bd699ebdbe21fa/docs/screenshots/shell/settings-laptop-light.png) |
| Offline at 390 | [open](https://github.com/zarghammir/ai-radar/blob/6902725882ca3de66498bbfa62bd699ebdbe21fa/docs/screenshots/shell/offline-phone-dark.png) |

## 6. The merge, and one correction

Rebased onto `911cc21`; `next.config.ts` conflicted because both sides rewrote the same object. **Both survive**: `output: "standalone"` from #29 and the `headers()` block. The reviewed commit is kept on a local ref so the rebase is reversible.

One correction to finding 1, offered as fact rather than argument: **the compose smoke does run on this branch.** CI run `35162222681` ran both `lint, typecheck, test, build` and `compose-smoke`, both success — it built the image, booted postgres and web, waited for a 200, confirmed the catalogue seeded and ran a worker pass. So the danger you identified was real and the resolution is independently proven, because dropping `standalone` would have failed that image build. Only the claim about CI coverage was stale.

## 7. Still not verified

- **Lighthouse on a deployed preview, and a real iPhone install.** Both need an HTTPS origin; this was a local production build.
- **The safe-area padding.** `main` now uses `calc(5rem + env(safe-area-inset-bottom))`. Headless Chromium reports the inset as 0, so nothing here can tell the difference. Unverified on hardware.
- **`npm run verify:shell` and `npm run a11y` do not run in CI.** I checked the workflows rather than assuming. Exit 2 means *the instrument could not run* and is non-zero, so Actions fails the step by default — but nothing wires these scripts today, so that three-state signal has no consumer. Filed against the QA harness ticket (#17) rather than reaching into Platform's workflow.

zarghammir:
REVIEW VERDICT: APPROVE

Reviewed at **`6902725`**. 51 files, counts agree both ways, base carries current `main`, `MERGEABLE`/`CLEAN`, both CI jobs green with every real step executed. Light checks at this head, run bare: `tsc --noEmit` **0**, `eslint .` **0**, `prettier --check .` **0**, `vitest run` **0** — 129 passed, 10 files, 0 skipped.

**All twelve items from my first review are fixed, and the four things this PR said it could not yet claim are now measured — by me, in a browser, not read off the report.** I have no open items on this PR.

---

## The blocker is gone, and the resolution kept both halves

`git merge-tree --write-tree origin/main HEAD` exits **0** and GitHub says `CLEAN`, where both said conflict before. `next.config.ts` carries `output: "standalone"` **and** the `headers()` block — the resolution I was worried about getting wrong. And `compose-smoke` now runs on this branch and passes, so a dropped `standalone` would fail the image build rather than being caught by nobody.

**A correction to my own first verdict, since the owner reads these.** I wrote that "nothing in CI would catch it — the smoke job lives on main, not here." That was **true of `c550602`**, where `ci.yml` on that tree declared exactly one job; the rebase is what brought `compose-smoke` onto the branch. The workflow file is branch content, so CI coverage is a property of the tree at a SHA, not of the repository. A claim of that shape needs its SHA attached, and mine did not.

## The four claims I could not check before

| | measured |
|---|---|
| phone states with the bar | **12 / 12**, six tabs and six labels in every one |
| narrowest tab | **65px** · shortest tap target **60px** (clears WCAG 2.5.8's 24px and 2.5.5's 44px) |
| every label box | **61px**, **0 clipped** in all 12 |
| nav overflow · document overflow · page errors | **0 · 0 · none** |
| laptop states showing the bar | **0 / 12** |
| roving tabindex | `['0','-1','-1']` — one tab stop in a 14-key walk |
| ArrowRight | focus **System → Light**, `aria-checked` moved with it · **End → Dark** |
| service worker | controlling; live probe really returned **404**; that URL's cache entry is **null** — the 404 was not stored |
| offline, the previously-404 URL | **200**, `Offline · AI Radar` |
| **my own control**, a URL never visited | **200**, `Offline · AI Radar` |
| theme-color, OS dark + "Light" chosen | meta **in force** `#e4e3de` against a bench of `rgb(228,227,222)` — they agree |

The service-worker rows are the same probe that failed in my first review, now passing in both directions, with the never-visited control beside it so a pass cannot come from nothing being cached at all.

The arrow-key row is the one axe structurally cannot produce: axe does not exercise keyboard interaction, which is exactly why 36 clean states told me nothing about it the first time.

**Three `theme-color` metas exist where two are declared, and that is not a defect.** The fix writes to *every* meta rather than the first, so the count cannot matter: whichever one the browser applies now carries the resolved colour. I verified that directly rather than accepting the disclosure — all three read `#e4e3de`, and the one whose media query matches is the one I measured against the bench.

## The instruments, attacked rather than reproduced

- **The committed audit passes honestly**: 24 states, 0 serious, 0 violations of any impact, 0 overflow, floors satisfied, 12/12 bars, 65px narrowest — matching my independent numbers exactly.
- **The 200px control engages and forces the condition it exists to force**: exit 1, viewport 200, labels clipped in **12 of 12** states, and the floor still intact at 12/12 bars measured, so the control did not break the instrument to get its red. The move from 280px was right: a control that cannot force the condition is decoration.
- **I proved the floors can fail**, because a floor only ever seen to pass is the same problem this PR keeps solving elsewhere. I renamed the bottom bar's landmark so the audit's selector misses it, rebuilt, and re-ran: **0 clipped labels, 0 nav overflow — which without the floor reads as a clean pass** — and the audit exited 1 with *"bar found in 0 of 12 phone states — that is the selector or the render, not the layout"*. That is precisely the vacuity the floor was written for, and it catches it.
- **The crash the PR body discloses is not in what ships.** The string `control exit=1, want 1` appears in no committed file, so that happened in an uncommitted harness. The committed `a11y-audit.mjs` does its browser work inside `try`/`finally` with the close in the `finally`, so a crash propagates and the success JSON is never printed. Both things are true and they mean different things: the disclosure is honest about what happened to the author, and **what ships is stronger than what bit them**. A lesson learned in a harness nobody can read dies with the session — if that runner still exists outside the repo, it is worth committing.

## The twelve, each verified in the code

`sw.js` gates on `if (fresh.ok)` and both `cache.put` calls are inside `event.waitUntil` · reachability is solved structurally, `bottomNavItems = navItems`, six tabs rather than a "More" link that would have needed writing · `theme-color` uses `querySelectorAll` in **both** places · full ARIA radiogroup with `Home`/`End` · `playwright-core` is a declared dependency and `scripts/lib/browser.mjs` resolves `PW_EXECUTABLE` then Chrome's channel, exiting 2 with a usable message · Settings announces "Loading settings" · the hard-coded weekday is now `LocalDate` · `primaryNav` is gone · the `storage` listener calls `applyTheme` · `.playwright-cli/` is ignored · `manifest.ts` takes the dark bench.

**And the one I raised as device-conditional is closed structurally, without needing a device:** `pb-[calc(5rem+env(safe-area-inset-bottom))]`. I could never have proved that one on hardware I do not have; making the padding track the inset removes the need to.

## Two notes, neither blocking

- **`AUDIT_CONTROL` is documented only inside the script.** I ran `CONTROL=narrow` first and got a silent normal-mode pass — `control=None`, viewport 390, exit 0. That is a control that looks like it ran. One line in the README, or a refusal when an unrecognised `CONTROL`-ish variable is set, would stop the next person making my mistake.
- **`local-date.tsx:41`** renders `<time>` with no `dateTime` attribute, and "Wednesday, 16 September" is not a machine-readable date, so the element is technically invalid HTML. The hydration handling around it is right — `getServerSnapshot` returns `""`, so the server emits nothing and the markup cannot disagree.

## Verified myself

Everything in the two tables above, from my own scripts against a production build I made at this SHA; the audit in both modes and the floor control; the twelve fixes in the source; base currency, merge state, file counts both ways, and every step of both CI jobs.

One number I checked before reporting: my first local run said "111 passed, 18 skipped". The 18 were the pipeline's database file skipping because my shell had no `DATABASE_URL` — not a gap in this PR. With a database it is 129 passed, 10 files, 0 skipped.

## Taken on trust

- **Real-device behaviour** — no iPhone, no Android, no HTTPS origin. The install flow, the Apple touch icon, the splash screen, the home-indicator inset and Lighthouse are all still unverified by me, and the PR is explicit about that. The inset is the one that stopped mattering, because the padding now tracks it rather than assuming it.
- **The seven committed screenshots** — I measured the live pages instead of diffing images.
- **Icon artwork** — referenced correctly; I did not open them to confirm they are the intended images or correctly maskable.

The merge click is the owner's, and this one has earned it.

Issue #11, opened 2026-09-16
Screen designs for sign-off: Today, Radar, Story, Saved, Settings


zarghammir:
**Problem.** The owner is visual and approves on seeing. No screen exists. Building UI before sign-off risks rebuilding it.

**Solution.** Load `impeccable` (run `/impeccable init`), `frontend-design`, `apple-design`, `emil-design-eng`. Produce a static, framework-free prototype (`html-prototyper`) of: Today (header with greeting, date, count, reading time, 5/10/all switch; story cards with category badge, title, summary, why it matters, primary source, also-reported-by, detected time, verification chip, actions), Radar (filters, sort, chronological list), Story detail (executive summary, why it matters, key points, original source, timeline, sources, verification explanation, why-ranked panel), Saved, Settings with onboarding (interests chips, brief time, timezone, notification choice). Mobile (390) with bottom nav and desktop (1440) with left sidebar, light and dark. Calm: no gradients, no hero, no infinite scroll. Use real story content from the seeded feeds, not lorem. Write `docs/DESIGN.md` with tokens (type scale, colors, spacing, badge system). Publish as one page for the owner.

**Expected output.** One link the owner can open on phone and laptop showing every screen in both themes. PM posts the owner's sign-off on this issue. Nothing in `src/app` changes in this issue.

zarghammir:
Owner sign-off, 2026-09-16 13:15: "I approved the design. It's very cool." Prototype: https://claude.ai/artifact/MTbzuW9JHtWfeyfXwVALCb · direction round: https://claude.ai/artifact/8rTQSaC1UTi9g3Y5WWvqQy · PR #23. Tokens move into src/config/brand.ts and the app shell under issue #12.

Issue #7, opened 2026-09-16
Seed real free sources and the topic list


zarghammir:
**Problem.** The app must ship with useful real content and zero paid sources. No source rows or topics exist.

**Solution.** `src/db/seed.ts` (idempotent upsert by key) with: PRIMARY tier feeds (OpenAI, Anthropic, Google DeepMind, Meta AI, Microsoft Research, NVIDIA, Hugging Face blog, Mistral, xAI where a feed exists), HIGH_QUALITY_REPORTING (The Verge AI, Ars Technica AI, MIT Technology Review AI, TechCrunch AI, VentureBeat AI, Wired AI, Simon Willison, Import AI, Interconnects), COMMUNITY (Hacker News top with AI filter), PRIMARY arXiv (cs.AI, cs.LG, cs.CL, cs.CV, cs.RO, stat.ML). Every feed URL must be fetched and parsed successfully at seed-authoring time; record the check in the PR. Topics from the brief (companies, domains, fields) with keyword lists for tagging. Also `fixtures/` with one saved response per adapter for offline tests.

**Expected output.** `npm run db:seed` prints the count of sources and topics; running it twice changes nothing. A script `npm run sources:check` fetches every enabled feed and reports OK/FAIL per source; all OK at merge time. Adapter tests run against fixtures without network.

PR #33, opened 2026-09-16
API contract for review

zarghammir:
Part one of #10. **Document only** — no routes, no code, nothing to run.

## What this is

The seam the UI lane builds against, written before any route exists so the shape is agreed rather than discovered halfway through two implementations.

`docs/api.md` covers every route the ticket lists: method, path, params, exact JSON shapes for the story card, story detail (items, timeline, topics, score components with labels, verification note), the brief envelope with count and reading minutes, the radar page with cursor, preferences, saved, topics and sources. Plus the error shape, pagination and the three different time windows.

## Every field is backed by something that exists

I cross-checked all **48 columns** the document names against the live database. None is invented.

Fields the schema holds but nothing yet writes are listed with the value they return until their ticket lands, rather than quietly promised:

| field | until then | filled by |
|---|---|---|
| `summary`, `whyItMatters`, `keyPoints` | `null`, `null`, `[]` | AI summariser, Phase 2 |
| `score`, `scoreComponents` | `0`, `[]` | #9, ranking run |

The document states twice that a client rendering `summary` without falling back to `excerpt` shows empty cards on every story in Phase 1. That is the one place this contract could mislead, so it is said where it cannot be missed.

## Three decisions, not transcriptions

These need sign-off rather than review, because each could reasonably go the other way:

1. **`length` is reading minutes, not a story count.** The product brief says five-minute and ten-minute modes, so a five-minute brief is however many stories fit in five minutes. "Top 5 stories" is the other reading and the two diverge quickly.
2. **`trending` is sources-attached-in-24h**, computed per request from each item's `fetchedAt`, because nothing stores a rate of change. Engagement velocity would need a stored signal and a ticket.
3. **No brief delivery is recorded.** The window is derived from `briefTime` and `timezone` on every request, so a brief cannot be "already read" and reloading does not use it up. Remembering one needs a new column.

## Two value sets that are not enforced anywhere yet

Called out rather than left looking like schema facts: `theme` and `topics.group` are unconstrained text columns. `light`/`dark`/`system` and `company`/`domain`/`field` are proposed here and would be held only by route validation. `briefLength` and `notificationChannel` are TypeScript-enforced but plain text in the database. Only `contentType`, `verification`, `tier`, `role` and `kind` are real Postgres enums where a wrong value cannot be stored at all.

## The rule the ticket asked to be explicit

> No route triggers ingestion, and no route makes an external call.

Stated at the top, with the consequence spelled out: the only tables any route writes are `saved_items`, `read_state` and `user_preferences`. No route writes `raw_items` or `stories`.

## Checks

Document only, so nothing changed behaviourally. `tsc`, `vitest`, `prettier --check` and `eslint` all exit 0 on this branch; 108 tests, unchanged from main.

Branched from `main` at `77e2842` with no dependency on #30, so it can be read and signed off in parallel.

Part two, the route handlers, follows once this is signed off and #30 merges.


zarghammir:
PM sign-off on the four open questions (UI lane's consumer review still to come before the contract is final):

1. **`length` is reading minutes.** Confirmed. A five-minute brief is however many stories fit in five minutes, at least one. "Top 5 stories" is not the product.
2. **`trending` = distinct sources attached in the last 24 h.** Confirmed for Phase 1. Engagement velocity is a Phase 3 signal with its own stored column and ticket; when it lands, `trending` may switch to it and the contract notes the change.
3. **No brief delivery recorded.** Confirmed for Phase 1. Phase 2's scheduled Morning Brief adds `digests`/`digest_items` and a delivery timestamp; until then the window is derived on every request. Per-story `read_state` is what "already seen" means today.
4. **7-day window for `storyCount` on topics and sources.** Confirmed.

Also binding: the ground rules section (no route triggers ingestion or an external call; routes write only saved_items, read_state, user_preferences) is the product's cost-protection promise and is a review finding if any implementation breaks it.

zarghammir:
Read as the consumer, against the prototype the owner signed off on #11 (`design/prototype/`, live at the link in PR #23) and the shell in #31. **No new backend work proposed.** Every gap below is marked *prototype adjusts* or *contract must add*, with the reason in one line.

Headline: **this contract fits the prototype better than I expected.** `verificationNote` answers the story page's "how this was graded" panel almost word for word, `timeline[]` is exactly what the pickup curve needs, `scoreComponents[]` with a server-owned `label` is precisely the why-ranked panel, and `sources[]` being the same de-duplicated list the badge is derived from means a card can explain its own chip without a second request. Four things I expected to have to ask for are already here.

Three things genuinely do not fit, and one of them is the only item I would hold the contract for.

---

## Today

| prototype element | field |
|---|---|
| Greeting, date | client clock; `window.to` if we want the server's day |
| "8 stories" | `count` |
| "11 minutes" | `readingMinutes` |
| 5 / 10 / Everything switch | `length` param; default from `briefLength` |
| Entry number 01, 02 | array index |
| Content-type badge | `contentType` |
| Verification chip word | `verification` |
| Headline | `title` |
| Summary | `summary ?? excerpt` |
| Why it matters | `whyItMatters` |
| Source name | `primarySource.name` |
| "+3 others" | `sourceCount - 1` |
| Time | `publishedAt`, rendered in `preferences.timezone` |
| "3 min" | `readingMinutes` |
| "Read at X ↗" | `url` + `primarySource.name` |

**Gaps**

- **The four-bar grade meter** — *prototype adjusts.* The meter is a rendering of the `verification` enum (4/3/2/1), not data; the client owns that map, and putting a number in the contract would let the picture and the word drift apart.
- **"2 unread from yesterday"** — *prototype adjusts.* `read` is per story, but there is no previous-window concept (your open question 3). It becomes "N unread" counted from `read` in the current brief, and "from yesterday" is dropped.
- **The pickup curve on a card** — *prototype adjusts.* `timeline[]` is `StoryDetail` only, and putting per-item timestamps on every row of a list is N×M for one sparkline. It moves to the story page, where `timeline[]` already supplies it exactly. On the card it becomes the true thing the card already carries: `sourceCount` and the span `firstSeenAt` → `lastActivityAt`.
- **The orange "you stopped here" flag** — *prototype adjusts.* Confirms your open question 3 from the consumer side: with no delivery record, the flag stops meaning "where you stopped last time". It becomes the boundary after the last consecutive `read: true` story, which needs no new column and is still honest.

---

## Radar

| prototype element | field |
|---|---|
| Type filter chips | `type`, repeatable |
| Verification filter | `verification`, repeatable |
| Topic filter | `topic`, repeatable |
| Sort | `sort` |
| "showing …" summary | `appliedFilters` |
| Row time | `lastActivityAt` (matches `sort=newest`) |
| Row headline / type / grade / outlet | `title`, `contentType`, `verification`, `primarySource.name` |
| Load more | `cursor`, `hasMore` |

**Gaps**

- **Arrivals-by-hour chart (24 columns, current hour in the reserved orange)** — **contract must add.** This is the one I would hold for. It cannot be derived from a paginated list: `limit` is 30 and `hasMore` is normal, so a client can only ever draw the page it fetched, and a histogram of one page is not wrong-ish, it is wrong. Fetching every story in 24 hours to draw 24 bars is the wrong shape for both sides. The smallest thing that works is a count per hour over the same filters — a `GET /api/radar/activity` or an `activity: [{hour, count}]` block on the radar response. It is a `GROUP BY date_trunc('hour', …)` over rows you already index for `since`. **If it is not added, the prototype adjusts by dropping the chart in Phase 1** — but the owner asked for the charts specifically and approved them, so I would rather you say no explicitly than have me quietly delete it.
- **"61 items today"** — *prototype adjusts.* No total is returned, and adding one would cost a second count query on every page. It becomes "showing N" from the returned array, which `appliedFilters` already supports.
- **"Last checked 3 minutes ago"** — no gap: `max(sources[].lastFetchedAt)` from `/api/sources`, one extra request on a screen that can afford it.

---

## Story

| prototype element | field |
|---|---|
| Executive summary | `summary ?? excerpt` |
| Why it matters | `whyItMatters` |
| Key points | `keyPoints` |
| Pickup over the day | derived from `timeline[]`: distinct sources cumulative by `publishedAt` |
| **"How this was graded"** | **`verificationNote`** — a direct hit, no derivation |
| Timeline | `timeline[]` (`at`, `sourceName`, `role`, `title`, `url`) |
| Where to read it | `items[]` / `timeline[]` |
| "first" marker on the earliest outlet | ordering of `timeline[]` |
| Why this ranked | `scoreComponents[]` (`label` from the server, `value` for the bar) |

**Gaps**

- **"differs" annotation** on an outlet reporting a different figure — *prototype adjusts.* That is an editorial judgement nothing computes; it was true of the one real story I used and is not generalisable. Dropped.
- **Phase 1 emptiness** — *prototype adjusts.* `summary` null, `whyItMatters` null, `keyPoints` `[]`, `scoreComponents` `[]`. Every one of those sections is **hidden when empty**, never rendered as a labelled empty block. I will fall back to `excerpt` everywhere `summary` is used; your doc says it twice and it is the single easiest way to ship blank cards.

**One clarification, not a gap.** Can `scoreComponents[].value` be negative? The prototype draws a penalty (unverified, emerging) as a hatched bar going the other way, because "why it ranked" is only honest if it shows what pushed a story **down**. If values are always positive I will render magnitudes only; if penalties arrive as negatives the panel works as designed. One sentence in the contract settles it.

---

## Saved

| prototype element | field |
|---|---|
| Grouping by Today / Yesterday / Earlier | `savedAt` |
| Card contents | `StoryCard` as elsewhere |
| Note, tags | `note`, `tags` — the prototype does not show these yet; they are an opportunity, not a gap |
| Empty state | no field needed |
| Save / unsave | `POST` / `DELETE /api/saved/:storyId` |

**Gap**

- **"4 on the pins · 12 minutes"** — *prototype adjusts.* Saved is paginated with no total, so the count describes what is loaded, or the badge goes. Idempotent save is the right call for a phone; a double tap on the bin must not be an error.

---

## Settings and onboarding

| prototype element | field |
|---|---|
| Interest chips | `GET /api/topics`, ordered by `storyCount`; written as `topicKeys` |
| Brief time | `briefTime` |
| Time zone | `timezone` |
| Default length | `briefLength` |
| Source list, names and homepages | `GET /api/sources` |
| Grade key definitions | static copy |
| First-run completion | `onboardedAt` |

**Gaps**

- **Source on/off toggles** — **contract must add**, or I make the list read-only. `GET /api/sources` returns `enabled`, but nothing writes it. A toggle with no endpoint behind it is a dead control, which is worse than not offering it. A `PUT /api/sources/:key { enabled }` is the smallest fix; it is a user preference about an existing row, not a write to `stories` or `raw_items`, so I do not think it crosses your "only the worker writes those" line — but that is your call, not mine.
- **Notification choice is a different shape** — *prototype adjusts.* The prototype offers "one push when the brief is ready" / "that, plus anything major" / "nothing", which is about **what triggers** a notification. `notificationChannel` is `push | email | none`, which is about **where it goes**. They are not the same axis. I will ship the contract's axis, since the channel is the thing that actually has a delivery mechanism behind it, and drop "plus anything major" until something stores that rule.
- **Theme now has two sources of truth** — *flagging, not asking.* #31 stores theme in `localStorage` so the no-flash script can read it before first paint; the contract also has `preferences.theme`. A server round trip cannot happen before paint, so localStorage has to stay authoritative for rendering. My proposal: the server field is the cross-device default, localStorage is the device override, and Settings writes both. Worth a line in the contract so nobody later "fixes" the duplication by removing the local copy and reintroducing the white flash.

---

## Your open questions, from the consumer side

1. **`length` is reading minutes.** Confirmed, and it matches what the prototype already shows: "8 stories · 11 minutes" reads as a time budget, not a top-N. Top-5 would make the header contradict itself the moment two long stories crowd out three short ones.
2. **`trending` as sources-added-in-24h.** Fine. It is also the only definition that is honest about what is stored, and the UI does not need to say the word "velocity".
3. **No brief delivery recorded.** Fine, and it costs exactly one prototype element (the flag, above). I would rather that than a column whose only job is to make a decoration true.
4. **`storyCount` over a fixed 7 days.** Fine for ordering chips. It is a hint, not a statistic, and nothing in the UI presents it as one.

---

## Summary

**Two "contract must add":** the radar activity histogram, and a write for `sources[].enabled`. Everything else the prototype needs either already exists or is mine to adjust — nine adjustments, all listed above, none of which touches what the owner signed off on visually except the Radar chart, which is why I have not quietly dropped it.

**One clarification:** whether `scoreComponents[].value` can be negative.

I can build Today, Story and Saved against this contract as written today.

zarghammir:
PM rulings on the consumer read (UI lane's comment above). Intelligence lane: fold these in, then the contract is final.

1. **Arrivals-by-hour histogram: ADD.** The owner approved the chart. New route `GET /api/radar/histogram` taking the same filters as `/api/radar` (type, topic, source, verification, since) and returning `{ from, to, buckets: [{ hour: ISO, count }] }`, 24 buckets by default, one GROUP BY over `lastActivityAt` (the same key `newest` sorts on). Read-only, cheap, no pagination.
2. **`PUT /api/sources/:key { enabled }`: ADD.** It is a preference about a catalogue row, not a write to stories or raw_items; the seed already leaves `enabled` alone. Add `sources` to the list of tables routes may write, with that one column only.
3. **`scoreComponents[].value` may be negative.** Say so. Issue #9 adds a `verificationPenalty` component (UNVERIFIED negative, EMERGING slightly negative) so the why-ranked panel can show what pushed a story down; until #9 lands no negative value is produced.
4. **Theme: two sources of truth, by design.** Write the UI lane's proposal into the contract: `preferences.theme` is the cross-device default, localStorage is the device override that the no-flash script reads before first paint, Settings writes both, and the local copy must never be removed.

Also fold in the two wording upgrades from my sign-off (trending is dated to Phase 3; per-story read_state is "already seen" today).

zarghammir:
## Final — `2df49e1`

Status line now reads **final, signed off by PM and UI lane on 2026-09-16**. All four rulings and both wording upgrades are in. Docs only: one file changed against `main`, and `tsc`, `vitest`, `prettier --check` and `eslint` all exit 0.

**Two routes added**, both because the consumer read found a real gap rather than a preference.

`GET /api/radar/histogram` — same filters as the list so the chart and the rows below always describe the same set, grouped on `lastActivityAt` which is the same key `sort=newest` orders by, and **every hour present including empty ones with `count: 0`**. Omitting empty hours would put silent gaps exactly where the chart should show a quiet period. Capped at 168 hourly buckets, and a longer window is a validation error rather than a silently truncated chart.

`PUT /api/sources/:key` — `sources` joins the writable tables for that one column, and the ground rules say so. Worth knowing: the seed already never overwrites `enabled`, so re-seeding cannot switch a source back on behind the reader. That was designed in #7 for a different reason and happens to be exactly what this needs.

**Three other rulings.** Score component values may be negative, with `verificationPenalty` named as the reason, because a bar chart assuming non-negative values renders wrongly the day #9 lands rather than the day it is written. Theme's two sources of truth are written in as deliberate, with the reason stated as the white flash rather than left as an apparent duplication. Trending is dated to Phase 1 rather than described as permanent.

**Also folded in.** The open-questions section is now the decisions actually taken, keeping the reasoning rather than only the answers. And three things the consumer read raised and did not need are recorded as deliberate absences — the radar total count, the notification trigger axis, and a numeric grade on a story — so they are not re-proposed later as oversights.

To the UI lane: thank you for the read. Four things you flagged as adjustments were ones I would have guessed wrong on alone, and the histogram argument was correct in a way I had not considered — I had treated it as a client-side derivation without noticing that pagination makes that impossible rather than merely awkward.

zarghammir:

Reviewed `feat/api-contract` at `c1ef067` — one file, `docs/api.md`, 45 lines of net change across three commits. Counts agree both ways (`gh pr diff --name-only` = 1, `git diff --name-only <merge-base>..HEAD` = 1). `git merge-tree --write-tree origin/main HEAD` exits **0** and GitHub reports `MERGEABLE`/`CLEAN`, so unlike #31 this one lands as-is.

Two changes, both small, both in the same subject the amendment already touched. I would approve this immediately after them — please read the length of this comment as thoroughness about a document that other people will implement against, not as a pile of objections.

I checked the amendment against the implementation rather than against its own reasoning: I fetched PR #36 and read `src/pipeline/ranking/score.ts` and `rank-all.ts` directly.

**The amendment is correct, and it was worth making.** The code really does emit exactly these two keys with exactly these two labels:

```
score.ts:163   unverifiedPenalty: "Unverified claim",
score.ts:164   emergingPenalty:   "Not yet corroborated",
score.ts:129   if (input.verification === "UNVERIFIED") c.unverifiedPenalty = -WEIGHTS.unverifiedPenalty;
score.ts:130   else if (input.verification === "EMERGING") c.emergingPenalty = -WEIGHTS.emergingPenalty;
```

Both are negative (`-6` and `-2`). `verificationPenalty` appears **zero** times in all 699 lines. Both sites were amended — the prose at 246 and decision-log item 7 at 687. Had this shipped unamended it would have named a key the code never emits, and the panel implementer would have found it.

---

## 1. The amendment now implies both penalties can appear on one story. At most one ever can. — `docs/api.md:246-251`

> "Issue #9 adds two penalty components so the why-ranked panel can show what pushed a story _down_ … `unverifiedPenalty` … and `emergingPenalty`"

`score.ts:129-130` is an `if` / `else if` on a single `verification` value. A story is `UNVERIFIED` **or** `EMERGING`, never both, so **exactly one penalty component — or neither — appears on any given story.** Never two.

Read plainly, "adds two penalty components" describes the component *catalogue*, which is true. But this is the paragraph a panel implementer reads to decide what a story's `scoreComponents` array can contain, and nothing here tells them the two are mutually exclusive. Someone laying out the negative side of a diverging bar chart, or writing a fixture, will reasonably plan for both.

This is a new imprecision created by the amendment — the old singular wording could not have suggested it — so I am filing it rather than mentioning it. One clause fixes it: *"A story carries at most one of these, since it is `UNVERIFIED` or `EMERGING`, never both."*

## 2. `label` is guaranteed unconditionally in two places, but the code falls back to the raw key — `docs/api.md:243-244` and `docs/api.md:564`

The contract states the guarantee twice, with no qualification:

- line 243-244: *"`label` comes from the pipeline's own label map, so the wording in the UI cannot drift from the weights."*
- line 564, provenance table: `` `scoreComponents[].label` `` → `` `COMPONENT_LABELS` in `src/pipeline/ranking/score.ts` ``

The implementation is `rank-all.ts:37`:

```js
label: COMPONENT_LABELS[key] ?? key,
```

When a component key is not in the map, **the label is the raw key**. This is not hypothetical or incidental — it is deliberate, reasoned in a comment at `rank-all.ts:30-32` ("an unlabelled bar is a visible gap someone will fix, a missing one silently stops the parts adding up to the whole"), and pinned by its own test at `rank-all.test.ts:341-346`, which asserts `scoreComponentList({ somethingNew: 3 })` yields `label: "somethingNew"`.

**The code's behaviour is right. The contract's description of it is not.** And the amendment sharpens the consequence: the new sentence at 254-256 tells clients to *"read the `label` the server sends rather than matching on the key."* That instruction is correct and I want it kept — but a client that follows it verbatim, on the strength of a guarantee that the label is always prose from the map, will render `somethingNew` to a reader the first time a component is added to the model ahead of its label. Which is exactly when it happens: the two keys in this very amendment were added to `WEIGHTS` and `COMPONENT_LABELS` in the same change, but nothing enforces that pairing.

So, to answer the question directly: **the "read the label, don't branch on the key" guidance holds and should stay.** It is not undermined by two keys being named in prose — the added sentence draws that line well. It is incomplete for a different reason, one that predates the amendment but sits in the lines you flagged: it does not say what `label` contains when the map has no entry.

One sentence at 243-244 covers it: *"A component with no entry in the map carries its key as its label, so an unlabelled component is visible rather than dropped."*

## 3. Minor — the example has no negative component to copy — `docs/api.md:223-227`

```json
"scoreComponents": [
  { "key": "recency", "label": "Recently active", "value": 13.1 },
  { "key": "primarySource", "label": "Primary source", "value": 20 },
  { "key": "corroboration", "label": "Corroborating sources", "value": 8 }
]
```

Three lines below, the contract warns that "a bar chart that assumes non-negative values will render wrongly the day it does." The example is the thing people copy into a fixture; the paragraph is the thing they skim. Adding `{ "key": "unverifiedPenalty", "label": "Unverified claim", "value": -6 }` to this array would do more to prevent that chart than the paragraph does, and it would carry finding 1 for free by showing one penalty rather than two. Not blocking on its own.

---

## Checked and cleared — not findings

- **Nothing else implies a single penalty component.** `verificationPenalty`: 0 occurrences in 699 lines. Every remaining mention of "penalty" or "negative" (lines 246, 248, 251, 252, 256, 687, 688, 689) is plural or key-specific. Both original sites were amended, not just the prose one.
- **The provenance table honours its own rule.** It opens with "A field cannot be added to this contract without filling a row here", so I tested the document against it. I initially thought `timeline` and `engagement` had no rows — they do (`items[].engagement` and `timeline[]`), my first read of the table was truncated. Every field in the `StoryCard`/`StoryDetail` examples resolves to a row or to a composite row that covers it.
- **Base staleness has no consequence here.** The branch is 10 commits behind `origin/main`, but it merges clean and I checked whether main moved anything this document describes. The only candidate was `b5345da` ("stop calling one newsroom plus an analyst independent reporting"), which changed the *wording* `deriveVerification` produces for `verificationNote`. The contract only states that field's provenance — "written by `deriveVerification` in the pipeline" — which is still exactly true. No drift. I would not hold this PR for a rebase.
- **The four ruled decisions are internally consistent**, and I did not re-litigate any of them:
  - `GET /api/radar/histogram` — takes the same filters as `/api/radar`, groups on `lastActivityAt` (the same key `sort=newest` orders by, so a bar and a row agree), fills empty hours with `count: 0` rather than leaving silent gaps, and caps at 168 buckets with a `VALIDATION_ERROR` instead of a silently truncated chart. The stated reason — a histogram of one 30-item page is not approximately right, it is wrong — is sound and the route's shape matches it.
  - `PUT /api/sources/:key` — writes `enabled` and nothing else, agrees with the provenance row ("the only column any route writes on that table") and with the "not here" row excluding `sources[].config`/`url`. The note that re-seeding never overwrites `enabled` closes the loop a toggle would otherwise leave open.
  - `scoreComponents[].value` may be negative — verified against the real weights; both penalties are genuinely emitted negative.
  - `theme` keeps two sources of truth — consistent with what I verified in #31, where `localStorage` is read pre-paint and `dataset.theme` is the runtime source. Correctly framed here as the fix for the white flash rather than duplication.
- **No route triggers ingestion or an external call.** The "not here" table states it as a rule ("No route may cause an external call; the worker owns that"), and no documented route contradicts it. 14 routes, each with its own heading, no orphans.

## Verified myself

- Head SHA is `c1ef067` per GitHub, matching what I was asked to review; the earlier `2df49e1` is superseded.
- Merge state: `merge-tree` exit 0, GitHub `MERGEABLE`/`CLEAN`. File count 1 = 1.
- Every key, label and sign in the amendment, read from PR #36's `score.ts` and `rank-all.ts` — not from this PR's description of them.
- The mutual exclusivity in finding 1 and the `?? key` fallback in finding 2, both read from the implementation and its tests.
- `verificationPenalty` absence, and the full inventory of penalty/negative mentions.

## Taken on trust

- **That PR #36 is what actually lands for #9.** It is open and unmerged; if its keys or labels change before merge, findings 1 and 2 change with them. This contract is describing code that does not exist on `main` yet, which is the nature of the document, not a defect in it.
- **The prose I did not machine-check.** I verified every claim that names a key, a route, a column or a file. I read but did not independently re-derive the rationale paragraphs.
- **Routes against implementations that do not exist yet.** Most of these 14 routes are unimplemented; I checked the contract for internal consistency and against the pipeline code that does exist, not against handlers.

---

This is a genuinely good contract — the provenance table with its own admission rule, and the "not here" table with a reason per omission, are doing real work that will save arguments later. Both findings are one sentence each, in lines the amendment already touched. Send them and I will approve in a new comment.

zarghammir:

Re-reviewed at `0f47c10`. One file, `docs/api.md`, +23/−9, `MERGEABLE`/`CLEAN`.

I said I would approve on sight and I cannot, for one line. **Findings 1 and 2 are fixed better than I asked.** Finding 3's fix introduced a new instance of exactly the thing this document exists to prevent, and it is a one-line change to put right.

## Fixed, and verified against the code rather than read

**Mutual exclusivity (finding 1) — fixed, and sharpened.** *"A story carries at most one of them, because a story has exactly one verification level… Nothing that lays out the negative side of a bar should reserve room for two."* That last sentence names the consequence for the reader who has to build the thing, which is more than I asked for.

**The label fallback (finding 2) — fixed in both places.** The new paragraph at `:247-252` states it, and the provenance row at `:578` now reads `COMPONENT_LABELS … falling back to the key if unmapped`. Both sites carried the same unconditional claim and both are corrected.

The wording is also careful in a way worth pointing out: *"If a component ever reaches the API without an entry in that map"* — **reaches**, not *is emitted*. That is the accurate framing, because since #36 an unlabelled component cannot be emitted at all (it fails to compile in either spelling — I proved both), while `scoreComponentList` reads `stories.score_components` out of the database, where a key persisted by an older build and since removed from the map really can arrive unlabelled. The sentence covers the reachable case and not the impossible one.

## The one line — `docs/api.md:227`

```json
{ "key": "emergingPenalty", "label": "Not yet corroborated", "value": -2 }
```

That component cannot appear on this story. **Line 139 of the same JSON object is `"verification": "PRIMARY_SOURCE"`**, and:

- `src/pipeline/clustering/verification.ts:42` — `if (primary.length > 0) return { level: "PRIMARY_SOURCE", … }` is the **first** branch. The example's `sources[]` leads with `openai-blog` at `tier: "PRIMARY"`, so this story is `PRIMARY_SOURCE` and can never be `EMERGING`.
- `src/pipeline/ranking/score.ts:129-130` — `emergingPenalty` is set **only** when `verification === "EMERGING"`.

So the example shows a value the ranker will never produce for it, eighty-eight lines below the field that rules it out, and two paragraphs above the sentence that explains the rule it breaks: *a story has exactly one verification level.* `PRIMARY_SOURCE` yields **neither** penalty, not one of them.

I do not want this read as a gotcha — the intent was right and my own suggestion invited it. I asked for a negative value in the example and this is the only example in the document that has a verification level to contradict, so the instruction and the constraint collided. Three ways out, in the order I would pick them:

1. **Move the negative into the prose.** A two-line snippet inside the *"`value` may be negative"* paragraph, showing an `EMERGING` story's components, demonstrates the shape without touching a worked example that is correct as it stands. It also puts the example next to the sentence explaining it.
2. **Add a second short `scoreComponents` block** elsewhere, labelled as belonging to an `EMERGING` story.
3. **Drop the line and leave the example as it was.** The prose now carries finding 1 and finding 2 properly, so the example is doing less work than it was when I raised this.

What I would avoid is swapping in `unverifiedPenalty` — `UNVERIFIED` is equally impossible for a story a lab published itself.

`corroboration: 8` on the same story is fine and I checked it rather than assuming: `sourceCount` is 3 and `sources[]` carries `openai-blog` plus `verge-ai`, so `distinctSources > 1` holds and the component is earned.

## Verified myself

- The delta is one file and nothing else moved.
- `verification.ts`'s branch order and `score.ts`'s penalty condition, read from `origin/main` and from #36's head respectively — the two facts the finding rests on.
- That `emergingPenalty`'s magnitude in the new line (`-2`) does match `WEIGHTS.emergingPenalty: 2` applied negative, and its label matches `COMPONENT_LABELS` exactly. The line is wrong about *which story*, not about the component.
- The example's `corroboration` against its own `sourceCount` and `sources[]`.
- That line 139 is the document's only example carrying a verification level, which is why this is one line and not a pattern.

## Taken on trust

- Unchanged from my first pass: that #36 is what lands for issue #9 — it is approved but not yet merged, so the keys and labels this contract names are still describing a tree that is not on `main`.
- The routes with no implementation yet: I checked internal consistency and the pipeline code that exists, not handlers.

Fix that line however you prefer and I will approve immediately — this is the only thing standing between the contract and the owner's click, and I am sorry it is a second round rather than the approval I promised.

zarghammir:
REVIEW VERDICT: APPROVE

Re-reviewed at `64d0451`. One file, `docs/api.md`, +82/−4, merges clean into current `main`.

The line I blocked on is fixed, and four more were found that I had not reported. I want to be precise about that, because it changes what this approval means: **I audited the field I had been arguing about and stopped; the lane audited the whole example and found four further contradictions in it.** `sourceCount: 3` beside a two-source list, an `openai-blog` item carrying Hacker News engagement, `firstSeenAt` taken from the wrong field, and a `score` that did not equal the sum of its own components — that last being the one property a reader copying the example into a fixture actually leans on. My finding was correct and too narrow.

## I re-audited it independently rather than accepting the audit

The lane reports verifying fourteen properties on a document it had itself contradicted twice. That is exactly the claim not to take on trust, so I parsed every JSON block out of the document and checked the merged `StoryCard` + `StoryDetail` story against the rules as the **merged** code states them. **22 properties, 0 failures:**

| | |
|---|---|
| `sourceCount` equals `sources.length` | ✅ |
| `verification` is `PRIMARY_SOURCE` iff a PRIMARY-tier source exists | ✅ |
| no penalty component on a non-`EMERGING`/`UNVERIFIED` story — **the line I blocked on** | ✅ |
| `primarySource` component present iff a PRIMARY-tier source exists | ✅ |
| `corroboration` present iff more than one distinct source | ✅ |
| `score` equals the sum of `scoreComponents` | ✅ |
| every source in `sources[]` has an item | ✅ |
| `timeline` length and source keys mirror `items` | ✅ |
| engagement only on COMMUNITY-tier items | ✅ |
| `primarySource`, `url`, `publishedAt` all match the one item with `role: "primary"` | ✅ |
| `firstSeenAt` equals the earliest item's `publishedAt` | ✅ |
| exactly one item has `role: "primary"` | ✅ |
| all 15 JSON blocks in the document parse | ✅ |

One check failed on my first pass and it was **my instrument, not the document**: I compared `primarySource` against source *names* when it is a `SourceRef` object. Corrected, it passes — the field is the PRIMARY-tier source and matches the primary item.

I also checked the generated text, which is the easiest thing for a document to get subtly wrong. `"Published directly by OpenAI and picked up by 2 other sources."` matches `deriveVerification`'s template exactly, including the plural: one primary plus two others, and `others > 1` selects `sources`. That sentence is now true of real output rather than plausible.

## The fix is better than the fix I asked for

You took option 1 — the snippet moved into the prose beside the exclusivity paragraph — and then added the sentence I did not think to ask for:

> *It is here rather than in the worked example above because that story is `PRIMARY_SOURCE`, and neither penalty can occur on one: a lab publishing its own announcement is neither unverified nor merely emerging.*

That records **why** the negative cannot live in the worked example. The next person who wants to illustrate a negative value reads the constraint instead of rediscovering it, which is the difference between fixing an instance and closing the hole. Not swapping in `unverifiedPenalty` was right for the same reason.

## `whyItMatters` on the card

Checked as asked, and it is **nullable rather than optional**: `"whyItMatters": null` appears literally in the `StoryCard` example, so a client sees a key with a null value, not an absent key. Those are different values and the distinction matters to anyone destructuring. The provenance row and the "value until then" table both carry it.

And nothing else grew: comparing the card block against `0f47c10`, the added key set is exactly `{whyItMatters}` and **nothing was removed**. `scoreComponents` is still detail-only, as ruled — I verified it is absent from the card block rather than taking the ruling's word.

## My last trust item is now closed by `main` moving

Both previous rounds ended with the same caveat: this contract described a tree that was not on `main`. **#36 has merged**, so I re-verified against `origin/main` rather than an open PR:

- `unverifiedPenalty: 6` / `emergingPenalty: 2`, applied negative — so the `-2` in the snippet is the real value.
- `"Unverified claim"` / `"Not yet corroborated"` — both labels byte-identical to the contract.
- `if (verification === "UNVERIFIED") … else if (… "EMERGING")` — the mutual exclusivity the prose now states.
- `labelFor(key)` with its `?? key` fallback — the behaviour the new paragraph documents.

Every key, label, sign and rule this contract names is now verifiable against merged code.

## The thing worth keeping from this review

**The contradiction survived two rounds because it sat eighty-eight lines from the field that ruled it out.** `"verification": "PRIMARY_SOURCE"` at line 139, the impossible component at line 227. A reader checking either end sees nothing wrong; only holding both in mind at once catches it, and no one reads a 700-line contract that way. That is why this document produced two self-contradictions in two rounds, and why the thing that closed it was an **audit of the whole example** rather than a fix to the reported line.

If this document grows a third worked example, that is a third thing that has to stay true against every rule in the other 700 lines — which is the argument for keeping exactly one, and for the checks above being a script somebody can run rather than a review somebody has to remember to do. Worth an issue, not worth holding this.

## Verified myself

- The delta is one file; base merges clean into current `main`; file counts agree both ways (1 = 1).
- All 22 properties above, computed from the document's own JSON rather than read.
- Every key, label, sign and exclusivity rule against **merged `main`**.
- That the card gained only `whyItMatters` and lost nothing, by diffing the parsed blocks across the two heads.

## Taken on trust

- **The formatting gate is CI's, not mine.** My local `prettier` could not run in this worktree — a stale `node_modules` missing `prettier-plugin-tailwindcss`, which is my environment and not the branch. I waited for CI instead of guessing: both checks are now **green** at this SHA (`lint, typecheck, test, build` and `compose-smoke`), so the gate I could not run myself has run.
- **The prose I did not machine-check.** I verified every sentence that names a key, a route, a column, a file or a generated string. The rationale paragraphs I read but did not re-derive.
- **Routes with no implementation.** Most of the 14 are still unimplemented; I checked internal consistency and the pipeline code that exists, not handlers.

Approved. This is the last thing between the contract and the owner's click.

zarghammir:
REVIEW VERDICT: APPROVE

Re-baselined on `50846b7`. **This supersedes my approval of `64d0451`** — that approval now points at a head the owner should not click.

The delta is the staleness fix and nothing else: one file, +17/−16, and I checked it is confined to the eight statements that stopped being true when #36 merged.

## The class of defect here is worth naming, because nothing catches it

None of those eight sentences was wrong when it was written. The document said no negative value is produced until #9 lands, that every score is `0` so the `importance` sort degenerates to id order, and listed `score` and `scoreComponents` as waiting on a ticket. All true at the time. **#36 merged and they became false without the file changing.**

No diff shows this. No test covers it. `git log` on `docs/api.md` shows nothing between the sentence being true and being false, because the change happened in a different file. It is only findable by re-reading the document cold against the world as it now is — which is what the lane did, unprompted, after I had already approved it.

The fix is the right shape, and better than correcting the eight sentences would have been: **the claims are now written about the state rather than about the ticket.** "A story the ranking run has not reached yet scores `0`" survives the next merge; "until issue #9 lands, every score is 0" could not. **Zero references to `#9` remain in the document** — I grepped, and the count is zero.

One sentence is worth singling out because it now says something the old one did not:

> *"…so a database that has never been ranked orders by id here; that is the absence of a score, not a ranking of zero."*

That names the distinction between an absent value and a real one, which is the family this project keeps being bitten by. The old wording described the same behaviour and taught nothing.

## Verified against merged `main`, not inferred

| new claim | checked against |
|---|---|
| a story the ranking run has not reached scores `0` | `schema.ts:166` — `real("score").notNull().default(0)` |
| `scoreComponents` is empty for such a story | `schema.ts:167` — `jsonb(...).notNull().default({})`, which `scoreComponentList` turns into `[]` |
| "the ranker's own suite asserts every component it can emit has a label" | `score.test.ts:314` *emits exactly the components the label map names* and `:323` *labels every one of them* — both on `main` |
| a chart assuming non-negative values renders wrongly | both penalties are applied negative on `main`, and the ranking run is merged, so this is now present-tense rather than predictive |

That first row is also, incidentally, why `stories.score` being `NOT NULL DEFAULT 0` makes "every story has a non-null score" an assertion that cannot fail.

## The audit again, because the file changed

I re-ran the property audit rather than assuming an unrelated edit left it intact: all 15 JSON blocks still parse, and **15 properties re-checked, 0 failures** — `sourceCount` against `sources`, verification against tiers, no penalty on a `PRIMARY_SOURCE` story, `score` equal to the sum of its components, timeline mirroring items, engagement only on COMMUNITY items, `firstSeenAt` as the earliest `publishedAt`, `whyItMatters` nullable on the card, `scoreComponents` still detail-only.

## Verified myself

- The delta is one file and confined to the staleness fix.
- Zero `#9` references remain.
- Every new claim against merged `main` as tabled above.
- The full property audit, re-run on the changed file.
- Base merges clean into current `main`; file counts agree both ways.

## Taken on trust

- **CI**, which was still in progress when I wrote this and which I waited for rather than guessed; the formatting gate is CI's, since my `prettier` cannot run in this worktree (stale `node_modules`, my environment not the branch).
- **The prose I did not machine-check** — I verified every sentence naming a key, column, file, route or generated string.
- **Routes with no implementation yet**, unchanged from previous rounds.

Approved at `50846b7`. Click this one, not the last.

PR #36, opened 2026-09-16
Ranking run and the why-ranked data

### LINKED ISSUE #9: Ranking run and the 'why ranked' data

zarghammir:
Closes #9.

Stories had no score, so Today could not order them and nobody could see why one story beat another. This scores every story from the last seven days and records the reasoning alongside the number, so the "why ranked" panel shows "Primary source +20, Corroborating sources +8, Unverified claim −6" rather than a bare figure nobody can argue with.

Three things land together: the penalty that lets a weak claim push a story *down*, the fix for first-party posts that named nobody, and the ranking run itself.

## `rankAllStories(db, now)` is exported and **nothing calls it yet**

Deliberate, and worth saying plainly so it does not read as dead code. Nothing in TypeScript calls `runIngest` today either — the worker is a compose service whose real loop is **#5**, and that ticket now carries calling this after each ingest as an explicit acceptance item, in both modes, with an end-to-end test asserting stories come out with a non-null score.

Without that, both halves stay green while the product does nothing: stories get ingested, never scored, and Today stays empty.

## The three pieces

**A weak claim now costs a story its place.** `unverifiedPenalty` at −6 and `emergingPenalty` at −2, stored as negative values. The sign is the behaviour, not a presentation detail: the approved screen draws these as a bar going the other way, and a clamp at zero would remove the only thing in the model that pushes down while every ordering test still passed.

Two keys rather than the single `verificationPenalty` the ruling named, agreed before building. `COMPONENT_LABELS` is a static key-to-label map, so one key carries one wording, and "Unverified claim" printed on a story that is merely *emerging* is a false statement in front of the reader. The API contract was amended on its own branch, so no document now names a key the code does not emit.

**First-party posts that name nobody now get tagged.** Tagging reads the primary item's text only, which is right — a forum headline must not tag a story — but it left a post titled "Introducing our new model" matching no keyword and carrying no company topic. Eight first-party sources carry default topic keys, and tagging unions those with the keyword matches. Defaults come from the **primary item's** source, so a thread joining later cannot bring its own. Only sources speaking for one organisation get them: an outlet reports on everyone.

**The ranking run.** One transaction per story. `storySources()` goes **straight** into `rankStory` with no second list — load-bearing, not incidental, since building a separate list is exactly how one module came to count items while the other counted outlets. `updatedAt` is deliberately untouched, because re-scoring changes no content and bumping it would make every run look like an edit and break determinism.

Engagement takes points and comments from the **same item**. A maximum of each across different rows would describe a conversation that never happened.

## Controls, including two that did not bite

Every control compiles, so each red is behavioural rather than a type error.

| removed | reddens |
|---|---|
| The penalty branch, or clamping it positive | all four penalty tests |
| Levelling the two penalty magnitudes | the emerging test alone |
| Collapsing the two labels | the label test alone |
| The topic-default union | the vague-title test |
| Defaults taken from every item, not the primary one's source | the forum test |
| The topic-match component | "ranks a followed-topic story above its unfollowed twin" |
| The 7-day window | the out-of-window test |
| Bumping `updatedAt` | the determinism test |
| Dropping unlabelled components | the fallback test |
| Genuine best-of-each engagement | the same-item test |

**Two did not bite, and both are recorded rather than quietly fixed.**

**The named control for the pass-through does not work alone.** "Removing the pass-through reddens the single-outlet test" — it does not. Removing `storySources`' de-duplication leaves it green, and so does removing `rankStory`'s own. Only removing **both** reddens it, because `rankStory` de-duplicates by key internally (PR #30, finding 1), so either layer alone is sufficient. That is defence in depth rather than a hole — but **anyone later deleting one layer as redundant needs to know the other is carrying it**, and that the test will not tell them.

**A quantity constant across the defect.** The engagement test could not tell same-item from best-of-each, because its strongest item led in both fields at once. It now uses two items whose best numbers live on different rows.

**Two mutations were invalid before they were useful**: one did not compile, and one was inert because it sat behind a condition that never fired. Both looked like passing controls. Typecheck is asserted on every mutation now, and a green control gets investigated rather than logged — which is the only reason either of the findings above exists.

## Also fixed

`RankInput` gained `verification`, which broke two existing call sites' typecheck. Both fixed rather than loosened: the catalogue test holds the level constant so it still measures the tier difference alone, and the pass-through test now uses the story's **stored** level, which makes it more real than the constant it had.

## Tests

125 → **153**, in 11 files. The ranking suite has its own database **name**, because vitest runs files in parallel and two files creating and dropping the same database race each other. Both database suites create, migrate and drop their own, and refuse to skip under `CI`.

| command | exit |
|---|---|
| `npx tsc --noEmit` | 0 |
| `npx vitest run` | 0 |
| `npx eslint .` | 0 |


zarghammir:

Reviewed `feat/ranking-run` at **`595de41`** — I re-baselined onto the new head rather than finishing at `89d4142`. The added commit is 34 lines in `score.test.ts` and touches no production file, so the three deletion controls I had already run still hold; I re-ran the whole suite at `595de41` anyway and attacked the new test separately.

Base is current (`main` ahead by 0), `MERGEABLE`/`CLEAN`, 8 files, counts agree both ways. Both CI jobs ran every real step at this exact SHA — `lint, typecheck, test, build` and `compose-smoke`, including the image build, the 200 wait and a worker pass. My own bare run: **155 passed, 11 files, 0 skipped, exit 0**, with `DATABASE_URL` set so the database suites genuinely executed rather than warning and skipping.

**The ranking logic itself is correct and it is well defended.** Every mutation I made was caught by something, usually by more than one thing. The four items below are a false comment, a test that does slightly less than it promises, and two small robustness points. None of them reaches a reader. I would approve on the fixes.

---

## Two of the three claims I was asked to test came back true. Here is the evidence.

### The single-outlet control: the claim is right, and the comment above it is wrong — `rank-all.test.ts:195`

The comment reads `// NAMED CONTROL for removing the storySources pass-through.` I ran the three mutations separately, each gated so a red could not come from code that fails to compile:

| mutation | new tsc errors | suite | which tests went red |
|---|---|---|---|
| **A** — remove the de-dup in `storySources` (`run.ts`) only | 0 | 1 failed / 152 passed | `run.test.ts` → *counts one source once when it files two items on the same story* |
| **B** — remove the de-dup inside `rankStory` (`score.ts`) only | 0 | 3 failed / 150 passed | `score.test.ts` → *pays no corroboration for one outlet repeating itself*, *scores seven items from one outlet exactly as one item from it*, *does not rank an emerging story above a corroborated one* |
| **C** — remove both | 0 | 5 failed / 148 passed | all four above **plus** *gives a single outlet filing twice no corroboration* |

So the claim in the PR body is confirmed: **that test only reddens when both layers are removed.** It is an end-to-end guard on "de-duplication happens somewhere", not a control for the pass-through — and the comment says otherwise, which is the one thing a reader will believe without re-deriving it.

**The reassuring half, which matters more than the finding:** each layer *is* individually defended, just not there. Mutation A is caught by `run.test.ts:378`, mutation B by `score.test.ts:114` and `:120`. There is no coverage hole here, only a mislabelled test. Please point the comment at the tests that actually control for each layer, or move the claim to them.

I swept the other one before filing this: `rank-all.test.ts:172`, `// NAMED CONTROL for removing the topic-match component`, **does** bite — deleting the `topicMatch` block reddens *ranks a followed-topic story above its unfollowed twin* along with its unit twin. One mislabelled control, not a pattern.

### The engagement test: confirmed, and its reasoning about the older test is confirmed too

Mutating `bestEngagement` to take the maximum of each field across different items reddened **exactly one test** — *takes points and comments from the same item, not the best of each* — while *uses the best engagement signal on the story* stayed green, precisely as the comment at `rank-all.test.ts:266-269` predicts. The quantity is no longer constant across the defect: same-item scores ≈6.8, best-of-each ≈10.6, and the assertion pins ≈6.8. **Not vacuous. This one is a good test.**

### `rankAllStories` has no production caller — confirmed

`rankAllStories`, `scoreComponentList`, `RANKING_WINDOW_HOURS` and `LabelledComponent` are referenced only by `rank-all.ts` itself and `rank-all.test.ts`. Nothing outside `src/` touches them either. Nothing needed it and nothing is broken by its absence; #5 picking it up is consistent with `score` staying `0` until then, which is what `docs/api.md` already tells clients.

---

## 1. The new label test is defeated by a different spelling — `score.test.ts:297`

```js
body.matchAll(/\bc\.([A-Za-z][A-Za-z0-9]*)\s*=/g)
```

The derivation is real — it finds all 11 keys the ranker emits, all 11 are labelled, and there are no orphans in either direction. The floor works: a regex matching nothing gives 0 and fails `>= 10`. So the vacuous case you were worried about is handled.

The gap is narrower and it is the rot case the test exists to prevent. I ran it as a pair:

| injected into `rankStory`, no label added | *labels all of them* |
|---|---|
| `c.freshness = 5;` | **RED** ✅ |
| `c["freshness"] = 5;` | **GREEN** ❌ — slipped straight through |

Identical behaviour, different spelling, and the second one emits an unlabelled component that would render `freshness` in front of a reader. (A second test, *sums correctly when the penalty outweighs everything else*, reddens in both cases — but only because `+5` changes the arithmetic. It is not a label check and would not name the problem.)

The doc comment at `:291` claims "Reading the assignments out of the source cannot drift from the source." It cannot drift **for one spelling**. Either widen the pattern to the bracket form, or derive at runtime from `Object.keys(components)` over inputs that exercise every branch.

**Also worth one line while you are there:** the floor is `>= 10` against an actual 11, so a future change that makes the pattern silently miss one key still passes. `expect(keys.length).toBeGreaterThanOrEqual(Object.keys(COMPONENT_LABELS).length)` ties the two together — a missed key then turns into a red instead of a shrug, and it needs no maintenance when a component is added.

## 2. `bestEngagement` resolves a tie by physical row order — `rank-all.ts:54`, fed by `rank-all.ts:86-89`

```js
if (!best || points > best.points) { ... }        // strict >, so the FIRST row wins a tie
```
```js
const itemRows = await tx.select({ metadata: rawItems.metadata })
  .from(rawItems).where(eq(rawItems.storyId, storyId));   // no ORDER BY
```

Two items on one story with equal `points` but different `comments` — two submissions of the same link, say — and which `comments` value is used depends on the order Postgres hands back rows.

To be precise about what this does and does not break: it does **not** violate the determinism claim at `:119`, and I want to be clear I am not reporting that. Ranking writes to `stories`, never to `raw_items`, so the physical order of the item rows is stable between two runs against the same database, and the *is deterministic for a fixed now* test is sound. What is latent is determinism **across** environments — a dump and restore, a replica, or a `VACUUM FULL` can reorder those rows and produce a different score for identical data. `ORDER BY rawItems.id`, or a tie-break on comments, settles it.

## 3. `ranked` counts stories it did not score — `rank-all.ts:139`, against its own doc at `:14`

`RankAllResult.ranked` is documented as "Stories inside the window that were scored", but the function returns `due.length` — the size of the *candidate* list. `rankOne` returns early at `:71` when the story has disappeared between the window query and its transaction, and that story is still counted. Rare, and harmless to the stored data. But the number is a report of work done, #5 will be the thing reporting it, and a count that cannot go down is not measuring what its comment says. Count the successes.

---

## Checked and cleared

- **The seed `topicKeys` family is complete.** All 7 configured keys (`openai`, `google`, `microsoft`, `meta`, `nvidia`, `hugging-face`, `mistral`) resolve to real entries in `TOPIC_SEEDS` — a key that matched nothing would tag nothing, silently, since `tagTopics` only does `byDefault.has(t.key)`. The only `PRIMARY` source without defaults is `arxiv-ai`, which is correct by the rule stated in the comment: it speaks for no single organisation. No non-primary source has defaults. There is an `anthropic` *topic* but no Anthropic *source*, so nothing was skipped. (My first parse flagged `openai` as missing; that was my own regex straddling two seed arrays, not a defect — corrected by reading `TOPIC_SEEDS` directly.)
- **`tagTopics`' new `defaultTopicKeys = []` parameter masks nothing.** It has exactly one caller in the whole repo, `run.ts:231`, which passes the real value. No second surface silently gets `[]`.
- **The postgres.js pipelining trap does not bite anything here.** Both database suites open one pool with `max: 4` and run sequentially; no test in this PR asserts or depends on two queries having used two connections, so none of them is accidentally testing one.
- **`storySources` de-duplication is value-stable.** All rows sharing a `sourceKey` join to the same `sources` row, so last-write-wins in the `Map` cannot change name or tier — only array order, which nothing downstream depends on.
- **The verification penalty is emitted exactly as `docs/api.md` now describes it** — I checked this while reviewing #33: `unverifiedPenalty`/"Unverified claim" and `emergingPenalty`/"Not yet corroborated", both negative, mutually exclusive via `if`/`else if`.

## Verified myself

- Base currency, merge state, file counts both ways, and both CI jobs' steps at `595de41` specifically — not at the SHA I started on.
- The full suite at `595de41`, run bare with a database: 155 passed, 11 files, 0 skipped, exit 0.
- Five mutations (A, B, C, D, E), each reverted, each with `git status` clean afterwards, and every one gated on a **differential** typecheck: this worktree has never been built, so `tsc` reports one pre-existing `LayoutProps` error on the clean tree. I compared each mutation's error set against that baseline and required **zero new errors** before trusting any red. All five passed that gate, so none of these reds came from code that would not compile.
- The 11 emitted keys, the 11 labels and the absence of orphans in both directions, computed rather than eyeballed.

## Taken on trust

- **That the arithmetic constants are the intended product judgement.** I verified the weights are applied as written and that the components sum to the stored score; whether `unverifiedPenalty: 6` is the right magnitude is a product call, not a review one.
- **Behaviour at real data volume.** Every database test here builds a handful of rows. One transaction per story is a sensible shape, but I did not measure a run over a realistic table, and `rankAllStories` has no production caller yet to measure.
- **`next typegen` output.** I did not generate route types in this worktree; I neutralised their absence with the differential gate above rather than by producing them.

---

Findings 1 and 3 are one line each, finding 2 is one line plus the floor. Send them and I will approve in a new comment.

zarghammir:
REVIEW VERDICT: APPROVE

Re-reviewed at **`d9b461e`**, not the `fa7a77e` I was handed — the head moved once more while I was reading, and since the newest commit touches a production file (`score.ts`) I re-based on it and re-ran everything rather than reporting on a superseded tree. Base current (`main` ahead by 0), `MERGEABLE`/`CLEAN`, both CI jobs green at this exact SHA, and my own bare run with a database: **157 passed, 11 files, 0 skipped, exit 0**.

**All four findings are fixed, and I proved each fix rather than reading it.** Every control below was gated differentially — this worktree has never been built, so `tsc` reports one pre-existing `LayoutProps` error on the clean tree, and I required **zero new errors** against that baseline before trusting any red.

| # | finding | control I ran | result |
|---|---|---|---|
| 1 | mislabelled control comment | read the replacement | fixed, and better than asked — it now names the two tests that actually control each layer, and they are exactly the two my mutations reddened |
| 2 | `c["freshness"]` defeated the label test | injected **both** spellings | `c.freshness = 5` → **1 new tsc error**; `c["freshness"] = 5` → **2 new tsc errors**. The hole I found is closed |
| 3 | engagement tie broken by physical row order | removed the `id` sort | **exactly one** test red: *breaks an engagement tie the same way whatever order the rows arrive in* |
| 4 | `ranked` counted candidates, not successes | reverted to `due.length` | **exactly one** test red: *counts stories it scored, not stories it considered* |

Both new tests discriminate on my machine, which was the open question — each reddens its own named test and nothing else. I also ran the direction the old test could never reach: adding a label with no emitter (`bogusLabel`) reddens *emits exactly the components the label map names*. The set comparison catches drift in **both** directions now, and tying the floor to the map's own length means a branch the matrix stops reaching becomes a red rather than a shrug.

Replacing a source-parsing test with a type is the right trade and I want to say so plainly: a compiler covers every spelling, a regex covers the ones its author imagined. Deriving the union from the label map so the two cannot drift is better than anything I would have suggested.

---

## One thing to fix before this is forgotten — the load-bearing absence is documented but not enforced

`score.ts:170-180` says it as loudly as prose can:

> **DO NOT ANNOTATE THIS AS Record<string, string>.** The absence of that annotation is load-bearing…

The comment is correct and it is exactly the right thing to have written. But I tested what happens when someone does it anyway, and the answer is that nothing stops them:

| I restored `: Record<string, string>` on `COMPONENT_LABELS` | result |
|---|---|
| new `tsc` errors vs the clean baseline | **0** |
| tests that went red | **0** — 157 passed, 11 files, unchanged |

So the whole guard — the thing that makes an unlabelled component impossible in any spelling — can be switched off by a tidying pass that adds a type annotation for readability, and **neither the compiler nor the suite says a word.** The protection currently rests on a future editor reading a comment, which is the state this project has repeatedly decided is not good enough: make the bad state unexpressible rather than warn about it.

**Two lines fix it, and I verified them in both directions rather than suggesting them untested:**

```ts
type AssertTrue<T extends true> = T;
export type ComponentKeyIsAFiniteUnion = AssertTrue<string extends ComponentKey ? false : true>;
```

| | new tsc errors |
|---|---|
| added at `d9b461e` as it stands | **0** — costs nothing, breaks nothing |
| added, then the annotation restored | **1** — `TS2344: Type 'false' does not satisfy the constraint 'true'`, pointing straight at the assertion |

If `ComponentKey` ever widens to `string`, `string extends ComponentKey` becomes true and that line stops compiling. It turns the comment into a check. It is type-only, so it costs nothing at runtime, and it belongs directly under `ComponentKey` where the comment already is.

**I am approving rather than blocking on this**, because the guard genuinely works today — I proved both spellings fail to compile — and the behaviour of this PR is correct. It is a durability gap in a guard, not a defect in the ranker. But it is precisely the kind of item that gets flagged, agreed with, and then dropped because it belonged to nobody, so please either take the two lines in this PR or give it an issue. I would rather say that now than have it be a surprise in six months when someone tidies a type annotation and the compiler stops protecting anything.

## Two notes, neither blocking

**`const c = {} as Record<ComponentKey, number>` (`score.ts:91`) could drop the cast.** The comment explains the cast as "because components are populated conditionally, so it starts out empty" — which is exactly what `Partial<Record<ComponentKey, number>>` expresses without asserting something untrue. The current type claims every component is always present, so a later reader who does `c.engagement.toFixed(1)` inside `rankStory` gets a clean typecheck and a runtime crash. `Partial<>` rejects both spellings of an unlabelled key just as effectively — the key set is what does that work, not the requiredness — and needs no `as`.

**`labelFor` is the right extraction.** Casting back to `Record<string, string>` inside it is contained and honest: the fallback exists for data read out of the database, where the key really is just a string, and the doc comment says so. No change wanted — noting it because I checked whether the cast could hide anything and it cannot.

---

## Verified myself

- Head, base currency, merge state and both CI jobs' steps at `d9b461e` specifically — not at either SHA I was handed.
- Full suite at this head: 157 passed, 11 files, 0 skipped, exit 0.
- Seven controls (both spellings, the annotation against `tsc` **and** the suite, the sort, the counter, the stale label), each reverted, each differentially gated, `git status` clean after every one.
- My own suggested fix, in both directions, before suggesting it.

## Taken on trust

- **That the matrix in `emitted()` will keep reaching every branch as the ranker grows.** It does today — `keys.size === labelled.length` passes, which is the check that enforces it — and that assertion is what will fail if a future branch goes unexercised. That is the right design; I am noting that its correctness is asserted rather than structural.
- **Weights as product judgement**, unchanged from my first review: I verified they are applied as written, not that the magnitudes are right.
- **Behaviour at real data volume** — every database test here builds a handful of rows, and `rankAllStories` still has no production caller to measure.

The merge click is the owner's, and I think this one is ready for it.

zarghammir:
REVIEW VERDICT: APPROVE

Re-reviewed at **`69ad800`**. The delta is the two type-only lines and nothing else — I checked that before checking anything else.

**All three things I was asked to confirm, measured on my own machine, differentially gated against the clean-tree baseline** (this worktree has never been built, so `tsc` reports one pre-existing `LayoutProps` error; every count below is *new* errors against that):

| check | result |
|---|---|
| **Nothing else moved** | `git diff d9b461e..69ad800` → 1 file, `score.ts`, +13 lines, and the whole diff is the assertion plus its comment |
| **The assertion fires** — annotation restored | **1 new error**: `score.ts(50,40): error TS2344: Type 'false' does not satisfy the constraint 'true'` — same line the lane reported |
| **It fires on the dot spelling** — `c.freshness = 5` | **1 new error**, TS2339 |
| **It fires on the bracket spelling** — `c["freshness"] = 5` | **2 new errors**, TS2339 + TS7053 |
| Clean tree at this head | **0 new errors**; full suite **157 passed, 11 files, 0 skipped, exit 0** |

The bracket spelling is the one that walked past the original source-parsing test. It no longer compiles. That hole is closed in both spellings, and by a mechanism that covers the ones nobody thought of.

## Why these two lines are not ceremony, stated here so nobody deletes them later

I confirmed the detail independently, and it is the whole argument:

**With `: Record<string, string>` restored on `COMPONENT_LABELS`, the ranking suite passes 43 of 43, exit 0.**

Not "passes with a warning" — passes completely, the same as an untouched tree. I measured the full suite too: 157 passed, unchanged. **No test can see a type-level protection**, because there is nothing at runtime to see. The union widening from eleven literal keys to `string` changes no behaviour, emits no value, and throws nothing. The guard would still *look* like protection in review — the comment would still be there saying the absence is load-bearing — while permitting every unlabelled component in every spelling.

So the two lines are not belt-and-braces over a test that already covers this. They are the *only* thing in the build that can observe it. That is why a comment alone was not enough here, and why deleting `ComponentKeyIsNarrow` as an unused type would silently remove the protection this PR exists to add. Keeping the explanatory comment beside the assertion was the right call: `TS2344: Type 'false' does not satisfy the constraint 'true'` is not self-explanatory to whoever trips it at 2am, and the comment is the difference between fixing the cause and deleting the assertion.

## The whole review, closed out

Every finding from my first pass is fixed, and I proved each fix rather than reading it:

1. **Mislabelled control comment** — replaced with one that names the two tests that genuinely control each layer, which are exactly the two my mutations A and B reddened.
2. **The spelling hole** — closed by the type, both spellings verified above.
3. **Engagement tie broken by physical row order** — removing the `id` sort reddens exactly *breaks an engagement tie the same way whatever order the rows arrive in*.
4. **`ranked` counted candidates, not successes** — reverting it reddens exactly *counts stories it scored, not stories it considered*.
5. **The load-bearing absence, unenforced** — fixed here, and this comment is the record of why.

Ten controls across the two passes, each reverted, each differentially gated, `git status` clean after every one.

## Still taken on trust

- **That the matrix in `emitted()` keeps reaching every branch as the ranker grows.** It does today, and `keys.size === labelled.length` is what will fail if it stops — the right design, but asserted rather than structural.
- **The weights as product judgement.** I verified they are applied as written, not that the magnitudes are right.
- **Behaviour at real data volume.** Every database test builds a handful of rows, and `rankAllStories` still has no production caller to measure — #5 is that caller.

## One non-blocking note, carried forward unchanged

`const c = {} as Record<ComponentKey, number>` (`score.ts:104`) could be `Partial<Record<ComponentKey, number>>` and drop the cast. The key set is what rejects an unlabelled component, not the requiredness, so the protection is identical — but the current type claims every component is always present, and a later reader who writes `c.engagement.toFixed(1)` inside `rankStory` gets a clean typecheck and a runtime crash. Worth a follow-up, not worth holding this.

Ready for the owner's click.

Issue #35, opened 2026-09-16
The LLM cost cap has no code enforcing it and no owner


zarghammir:
Found by the Platform lane while building issue #5, and confirmed by listing all 34 issues in this repo: **nothing enforces the cost cap, and no ticket owns it.**

- `LLM_MAX_STORIES_PER_DAY` exists in `.env.example`.
- `llm_usage` exists in the schema.
- **No code reads either**, and there is no issue anywhere for AI summaries or any other paid call.

So the Phase 1 requirement of hard caps is currently a rule with nothing implementing it. That is fine only while the value is zero — the day a key goes in, an unenforced cap is indistinguishable from no cap.

`docs/cost-protection.md` (landing in #5) deliberately says the guard does not exist yet rather than describing one that does. Do not "fix" that document by making it describe a guard.

**This issue must be closed before any paid call ships.** Whoever writes the first paid call brings with it: the daily counter written to `llm_usage`, the refusal path when the cap is reached, a test that the refusal actually blocks the call (not just logs), and the owner-visible statement of what a day costs at the cap.

Owner assigned by the PM so that the cap is not rediscovered at the moment a key is added. This is the requirement's owner, not a suggestion.

Issue #34, opened 2026-09-16
Docker image carry-ins from PR #29, after #29 and #5 both land


zarghammir:
Three review carry-ins from PR #29 edit files that exist only on `feat/docker-compose`, and one of them only makes sense once the real worker (issue #5) exists. They are parked here so they are not dropped between the two branches.

Owner: Platform lane. Do this on a follow-up branch cut from main once **both** PR #29 and the issue #5 worker PR have merged.

1. `COPY public/` into the image.
2. Worker restart policy in docker-compose.yml.
3. `COPY . .` after `npm ci` so the layer cache behaves.
4. The compose-smoke manifest assertion.

Also verify at that point: the smoke passes `--once` to the worker and the compose service stays argument-free, so the service runs the loop while the smoke exits. That flag is a no-op today and becomes load-bearing the moment the worker lands — the comment in the workflow explains why; do not remove either.

Filed by the PM session from the PR #29 review trail.

zarghammir:
**Correction to my own acceptance above — the assertion I specified could not have failed.**

I wrote that the smoke should assert `GET /manifest.webmanifest = 200` so that a missing `public/` could not stay green.

**That file is not in `public/`.** It is generated by `src/app/manifest.ts`, a metadata route served by Next — so it answers **200 with `public/` missing entirely.** The assertion would have been green in precisely the case it was written to catch, and the PR would have shipped with a check that proves nothing while reading as proof.

The Platform lane found it by checking where the file comes from instead of implementing the line as given. The corrected assertion hits paths that really do live in `public/` — `/icons/icon-192.png` and `/offline.html` — and keeps the manifest check with a comment saying why it cannot stand in for them. The deliberate red at #45 demonstrates the difference directly: the icon 404s while the manifest still answers 200.

**Second correction, same source.** "`COPY . .` after `npm ci`" does not work as written: reordering means `npm ci` runs before the source exists, and this package's postinstall is `next typegen`, which exits 1 with "Couldn't find any `pages` or `app` directory". Verified in an isolated directory rather than reasoned about. The reorder therefore needs `--ignore-scripts`, with `npm rebuild` once the tree is complete — six packages declare install scripts, **esbuild among them, and esbuild is what bundles the worker**, so skipping them outright would produce an image that builds and cannot run.

[...]

Issue #14, opened 2026-09-16
Radar page: the live feed with filters


zarghammir:
**Problem.** Users who want everything have no place to explore it.

**Solution.** `app/(app)/radar/page.tsx` reading `GET /api/radar`. Quick filter chips: All, News, Research, Models, Releases, Community (X arrives in Phase 3 with the Bluesky adapter). Filter sheet: topic, source, content type, verification, time range. Sort: Newest (default), Importance, Trending. Chronological list of compact rows, cursor pagination with a "Load more" button, never auto-infinite scroll. Filters in the URL so links are shareable.

**Expected output.** Playwright: choosing Research shows only papers; Importance order matches score order from the API; a URL with filters restores them on load; keyboard can reach every chip. Screenshots at both widths, both themes.

Issue #13, opened 2026-09-16
Today page: the morning brief


zarghammir:
**Problem.** The hero feature does not exist.

**Solution.** `app/(app)/today/page.tsx` (also `/`) reading `GET /api/brief`. Header: greeting by local time, full date, "N things worth knowing", estimated reading time, 5 / 10 / Everything selector (persisted in preferences). Ranked story cards per the signed-off design: category badge, title, summary or excerpt when no AI summary, why-it-matters when present, primary source + also-reported-by, detected time, verification chip, actions Open / Save / Share (Web Share API with clipboard fallback) / Hide. Empty state when no stories yet with a plain explanation and the time of the next run. Optimistic save/hide with rollback.

**Expected output.** Playwright: at 390 and 1440, light and dark, the page renders real seeded stories; the 5-minute switch shortens the list; Save toggles and survives reload; Hide removes the card and it stays hidden after reload. Screenshots in `docs/screenshots/today/` attached to the PR.

PR #37, opened 2026-09-16
Worker process: scheduler, one-shot mode, secret-guarded trigger, GitHub Actions schedule

### LINKED ISSUE #5: Worker process: scheduler, one-shot mode, secret-guarded trigger, GitHub Actions schedule

zarghammir:
Closes #5.

## In plain words

**What is happening in the records.** This adds the part of the app that actually goes and gets the news. It reads the list of sources (the 18 feeds already in the database), fetches what each one has published, stores the new items, and writes one row per source into a log table saying how many it fetched, how many were new, and what went wrong if anything did. Nothing is deleted and nothing is overwritten.

**The flow.** A self-hoster starts the app; the fetcher runs immediately and then every 30 minutes, for as long as it is running. Someone who prefers not to leave it running can instead let GitHub run it on a schedule, or press a button to run it once. There is also a private web address that triggers a run, protected by a password, for anything that needs to kick it off from outside.

**Who uses it.** Nobody clicks it. It is the machinery behind the app: whoever runs AI Radar, whether that is you on a laptop or a server somewhere, and nobody else.

**What exactly they see.** A run prints one line per source — `openai-blog: 50 fetched, 50 new` — and then one total line. That is the whole interface. There is no screen in this pull request; the screens that will display what it collects are issues #13 and #14.

**What purpose it serves.** Until now the app could be told how to read a source but had nothing that ever went and read one, so the database only filled up if a person ran a command by hand. This is the thing that keeps the app fed, and the reason your morning brief can have anything in it. **If this is not merged, the database only fills when somebody runs a command by hand, and the morning brief has nothing in it.**

**Where it stands.** Built, tested and committed. One small commit is still to come (the ranking call, explained below) and three checks are queued behind another lane's build. A click is not wanted yet.

## What it adds

- `src/worker/main.ts` — one pass with `--once`, or a pass every `INGEST_INTERVAL_MINUTES`. It is deliberately thin: every decision it makes lives in a tested module, so the entry file is wiring and signal handling.
- `src/worker/lock.ts` — one ingest at a time, anywhere, via a Postgres advisory lock.
- `src/worker/secret.ts` — refuses to start on an empty or placeholder `INTERNAL_API_SECRET`.
- `src/worker/trigger.ts` + `src/app/api/internal/ingest/route.ts` — `POST /api/internal/ingest`, 401 before any work, 409 while a pass is running, 503 when the server's own secret is unreadable.
- `.github/workflows/ingest.yml` — `*/30` plus `workflow_dispatch`, `concurrency: ingest`.
- `docs/cost-protection.md`, and the `worker` / `worker:once` scripts.

## The ranker call — landed

Per the acceptance item on the ticket (#5 comment 5705965162), the worker calls `rankAllStories` after `runIngest`. This was declared as a final commit while #36 was in flight; **#36 has merged and the commit is here**, so the branch is complete.

The reason it was stated rather than left implied: if the two pull requests had landed separately with nothing joining them, ingestion would run, nothing would be scored, the Today screen would stay empty, and **every test in both pull requests would stay green while it happened**.

It is wired inside `ingestOnce` rather than per-mode, so both worker modes and the HTTP trigger get it from the one shared path. Wiring it per-mode would have been three places to forget it in.

**The acceptance wording could not have failed as written, so the test asserts something else.** It asks for stories with a *non-null* score. `stories.score` is `NOT NULL DEFAULT 0` in the schema, so every story has a non-null score whether or not anything ever ranked it, and "every story is scored" is additionally true of no stories at all. The test instead floors the items ingested, floors the count of stories scored, checks the score *components* — which only a scored story has — and asserts the ranker's own count agrees with the table rather than with itself.

`ingestOnce` takes `RunIngestOptions` so a test can inject a fake fetch and a pinned clock and prove the wiring offline. Production passes nothing: both real call sites call it with two arguments, and the seam says so in a comment where the next person edits it. **Enforcing that structurally was considered and declined** — a grep-shaped assertion against today's file layout rots silently and then reads as protection while protecting nothing. Worth revisiting if that surface grows past two callers or the seam moves somewhere less visible.

Control: replacing the `rankAllStories` call with a zero result reddens exactly the two ranker tests and leaves the other 45 green.

**On the number in the smoke log — `483 stories` created, `scored 206`.** That gap is the ranking window and nothing else, checked rather than reasoned about:

- `rankAllStories` selects its work with `where(gte(stories.lastActivityAt, cutoff))`, a seven-day window. Feeds carry older items, so stories whose last activity predates the cutoff are never selected. Exclusion is by date alone.
- **A swallowed per-story error would produce an identical-looking number, so this is worth being explicit about: it cannot happen here.** There is no `try`/`catch` anywhere in `rank-all.ts`; a story that threw would propagate and fail the pass rather than quietly reduce the count.
- There is exactly one path that scores nothing and still moves on — `if (!story) return false` at `rank-all.ts:82`, the story having vanished between the window query and its own transaction. In this worker's pass that race cannot fire, because the pass holds the advisory lock and nothing else is writing.
- The window's behaviour is covered by the ranker's own suite, not taken on faith: *"leaves a story outside the window untouched"* and *"does not rank when there is nothing in the window"*.

Worth noting for its own sake: the advisory lock added here for story-slug collisions is what closes that third path. Nothing else is writing while the pass holds it, so the vanished-story race cannot fire inside a worker pass. That is the second guarantee on this branch doing useful work somewhere its author did not design it for — the first being the 503 on a missing secret, which is what keeps preview deployments from ingesting into the live database.

**There is deliberately no test in `src/worker/ingest.test.ts` for the window itself**, and the file says so where a reader would look for one. The window is another module's rule with two tests of its own; asserting it from here would pin someone else's behaviour from the outside and keep passing on the day they change it — coverage in appearance only. These tests own that the wiring fires.

## The ctx.log carry-in: verified, not reimplemented

The FetchContext / `ctx.log` provider owed from the PR #1 review was discharged by #30. `createFetchContext` in `src/pipeline/context.ts` builds the context, retains the lines, and `runIngest` stores them on `ingest_runs.error` when a source fails. I verified that and built on it. A second context in the worker is exactly the defect the comment in that file warns about, so there is deliberately no new one here.

## A test that nearly shipped measuring nothing

This is the part of the review I would most want a second pair of eyes on.

`withIngestLock` reserves a single connection for the whole run. An advisory lock belongs to the session that took it, so a lock taken on one pooled connection and released on another is never released at all, and every later tick is declined until the process restarts.

My first test for that guarantee **passed with the reservation removed** — three runs, all green. The name claimed a guarantee the body did not exercise. The cause: postgres.js pipelines queries onto one connection until something pins the others, so the lock and the unlock landed on the same connection anyway and the defect stayed invisible.

I reproduced the real failure in a scratch probe first — the unlock returned `false` and `pg_locks` still showed the lock held — and then rebuilt the test around the arrangement that actually exercises it: **four open transactions pinning connections, freed mid-run, with one ordinary pool query before the run ends.** Evidence: **green 2/2 with the reservation, red 3/3 without it**, reddening exactly the named test.

Worth knowing beyond this PR: any DB test in this repo that assumes two queries went to two connections is probably testing one connection.

## Kill mid-run: what happens to the lock

Asked during review, answered with a test rather than a claim. `pg_try_advisory_lock` is session-level, so the session dying *is* the release — SIGKILL never reaches the `finally`. The test starts a child that takes the lock, asserts `pg_locks` shows it held, kills the child with SIGKILL, waits for the lock to disappear, and then takes it again. Control: pointing the child at a different key reddens the test, so the assertion is live.

**Not tested:** a machine that vanishes without closing the socket. The backend would linger until the connection is reaped, and I am not going to guess at that timeout here.

## Typecheck was not clean on the first pass

Recording it because a merge reported as clean is a weaker record than one that needed a fix and got it.

After merging current main, `npm run typecheck` failed. Next's `global.d.ts` redeclares `NodeJS.ProcessEnv` as just a required `NODE_ENV`, without the index signature `@types/node` normally provides, so a plain `{ INTERNAL_API_SECRET }` literal is not assignable to it and the guard could not be called from a test. Fixed by typing the parameter as what the function actually reads. The merge was a clean git merge; this still came out of it.

## Two things this change broke, both fixed here

Both would otherwise have landed as somebody else's mystery red.

**The compose smoke.** Its first step is `cp .env.example .env`, which leaves `INTERNAL_API_SECRET=change-me`. The worker now refuses that value, so the moment this branch landed the smoke's worker step would have exited 1 — and the red would have looked like a broken worker rather than a stale fixture. The step now generates a real secret, which is also what the quick start tells a self-hoster to do, so the smoke still tests the documented path rather than a CI-only one.

**The README quick start.** It was `cp .env.example .env && docker compose up`. With the guard, that start now produces a web app that works and a worker that exits, with the reason only visible in a container log. The quick start generates a secret and says why.

## Since review: the smoke's gate is now known to be able to fail

The review found that `compose-smoke`'s success check proved the worker *launched*, not that it did anything: it grepped for a line `docker/worker-entrypoint.sh` prints in the shell before `exec`. It now greps `^total: [1-9][0-9]* sources`, which `formatTotalLine` cannot print unless a pass completed, with the `[1-9]` flooring the source count.

That gate had only ever been seen to pass, so it was proven able to fail, deliberately, on a throwaway branch:

**Run `35163983665`**, from a branch whose worker returned "declined" without doing anything.

`compose-smoke` (job 105020830766) went RED at *Run one worker pass* with `worker exited 0 without completing a pass`. That red is attributable to the assertion: the image built (231 MB), postgres came up healthy, the container started, the entrypoint printed its line and the worker exited 0 — the exact state that used to be reported as a successful pass.

**Both jobs on that run failed, not just the smoke.** `lint, typecheck, test, build` also went red, at *Unit tests*, on `trigger.test.ts > runs the ingest when the secret is right` — the positive assertion added under finding 4 of this review. That is the suite doing its job: a worker deliberately broken to never ingest fails the test that asserts a correct secret causes an ingest. It does not weaken the control, because the control's claim is about which step in `compose-smoke` rejected which output, but the honest version of this paragraph names it rather than implying the smoke was the only thing red. An earlier draft of this section said lint and typecheck passing meant nothing else could produce a red; that was true of the local runs it was written from and not true of the CI run it was later attached to.

The branch was deleted and its pull request closed as soon as the run id was recorded.

**What the green on this branch does and does not show**, stated because a green read as more than it carries is how a check becomes decoration: `compose-smoke` passing here shows the assertion accepts a real pass in a real container. The deliberate red above is what shows it rejects one that did nothing. Neither run alone would have been enough.

The other six findings — an expired comment, a negative assertion with no positive beside it, no floor on the ingest interval, two states sharing one operator-facing message, an unlock masking the failure that caused it, and a `concurrency` comment that described skipping where GitHub queues — are in the commits and answered in full at the review reply.

## The merge with #31, and which instrument verified it

#31 and this branch both appended scripts after `"sources:check"`, so both edited that line to add a trailing comma and the second to merge went dirty. Resolved by keeping **all five** — `sources:check`, `a11y`, `verify:shell`, `worker`, `worker:once` — rather than taking a side, because taking one side here drops the other's scripts and the result still looks reasonable.

Checked after resolving rather than assumed: all four added scripts present, all eight of #31's new packages still declared (six dependencies and two devDependencies), `package-lock.json` identical to `origin/main`'s since this branch never touched it, and every dependency in `package.json` has a lockfile entry — the condition `npm ci` actually enforces.

**The light checks were not run locally for this merge**, and the PR body should say so rather than let a reader assume otherwise. The merge brings in seven packages this worktree does not have installed, and #31's components import them, so a local typecheck would have failed on unresolved imports rather than on anything real. CI installs from the lockfile on a clean checkout, which is the stronger instrument here; the conclusion lines at the resolved head are the check.

**One argument for this PR landing, from another lane's failure.** Intelligence's #42 failed in `compose-smoke` with *"Failed to collect configuration for /api/topics"* — the same defect this branch fixed for `/api/internal/ingest`: `src/db/client.ts` throws at module scope when `DATABASE_URL` is unset, and `next build` imports every route module to collect page data. The `Build` step added here, which runs with `DATABASE_URL: ""`, catches that class of defect in the fast job instead of two minutes into an image build.

A caveat worth recording for anyone reproducing it: `env -u DATABASE_URL npx next build` can still **succeed**, because Next loads `.env` from disk regardless of the environment. My own reproduction was valid only because this worktree has no `.env` file at all. In a worktree that has one, the `.env` must be moved aside to see the failure.

## Verification

- `npm run lint`, `npm run format:check`, `npm run typecheck` — all exit 0.
- `npx vitest run src/worker` — 40 tests, 4 files, exit 0.
- Deletion controls, each run in the same window as the test it checks: removing the placeholder refusal reddens only the placeholder test; removing the fixed-length digest reddens only the length test; removing the decline reddens the second-worker test; moving the unlock out of the `finally` reddens the throw test; removing the connection reservation reddens the pooling test; running the ingest before the secret check reddens all three 401 tests *and* the positive.
- The 401 tests assert `ingest_runs` is untouched, not just the status code, and the positive test beside them asserts a correct secret does write a row — a route broken in both directions would pass the rejections alone.

Acceptance, against a throwaway database rather than the dev one, run through the bundled `ops/worker.cjs` using the Dockerfile's own esbuild command:

```
  openai-blog: 50 fetched, 50 new
  ... 18 sources ...
total: 18 sources, 516 fetched, 516 new, 502 stories in 18.9s     exit 0

second run:
total: 18 sources, 516 fetched, 0 new, 0 stories in 5.9s          exit 0
raw_items = 516 (unchanged)
```

With no secret set it refuses and exits 1, naming `openssl rand -hex 32`.

## Still to run, queued behind another lane's build slot

`npm test` in full (I have only run `src/worker`, and this branch adds an export to `src/db/client.ts`), `npm run build`, and the ticket's `curl` check through a real HTTP server. The 401/200/409 behaviour is proven at the handler level against a real database, but not yet through a server.

## Not in this PR

- The three Docker image carry-ins — #34, mine, on a branch cut after this one merges.
- The LLM cost cap — #35. `LLM_MAX_STORIES_PER_DAY` and `llm_usage` exist and nothing reads either. `docs/cost-protection.md` says so deliberately: it records that the guard does not exist rather than describing one that does, and it should not be "fixed" by making it describe a guard.


zarghammir:

Reviewed `feat/worker` at **`d71f49d`**, not the `210f6e6` I was handed — the head moved twice while I was reading, and the newest commit ("take the database client at request time, not at module load") touches the route I was reviewing, so I re-based on it rather than reporting on a superseded tree. Base is current (`main` ahead by 0), `MERGEABLE`/`CLEAN`, 18 files, counts agree both ways. Both CI jobs ran every step at this exact SHA. My own bare run with a database: **169 passed, 14 files, 0 skipped, exit 0** — and no `DROP DATABASE` hang, which the request-time import is what prevents.

This is careful work and most of it survived everything I pointed at it. The secret guard is fail-closed, constant-time, and its placeholder list is pinned against `.env.example` **with a floor on the capture** so an empty match cannot satisfy the assertion for the wrong reason — that is the vacuity guard I would otherwise have asked for, already written. The advisory lock reserves a connection because the lock is session-scoped, and its tests read `pg_locks` directly to prove the lock is held during the work and released after it, including after a throw.

Two findings are substantive. The rest are small.

---

## 1. The smoke's success check cannot tell a real pass from a worker that did nothing — `.github/workflows/ci.yml`

The step's own comment is explicit about what it refuses:

> *Accepts exactly two outcomes and nothing else: a real pass once issue #5 lands, or today's explicit refusal. **Silence, a crash or an empty exit 0 all fail.***

The exit-0 branch tests for this string:

```sh
echo "$OUT" | grep -q "\[worker\] running one ingestion pass"
```

That line is not printed by the worker. It is printed by `docker/worker-entrypoint.sh:11`, in the shell, **immediately before `exec node /app/ops/worker.cjs`** — before any work happens at all. It proves the bundle exists and was launched. It cannot prove a pass ran.

So an empty exit 0 does **not** fail, and neither does a pass that was skipped: `src/worker/main.ts:50-53` logs `another ingest is already running; skipping this pass` and returns **0**, which lands in the exit-0 branch, finds the entrypoint's line in the output, and reports *"RESULT: the worker ran a pass."* — when none did.

The right string is already there and your own CI run prints it:

```
[worker] running one ingestion pass          <- the entrypoint, before anything
total: 18 sources, 496 fetched, 496 new, 482 stories, 1 failed in 9.0s   <- the pass
```

`total:` comes from `formatTotalLine` at `main.ts:57`, which is reached only when a pass completed. Grepping that instead — ideally with a floor on the source count, the way the seeding step already floors `select count(*) from sources` — makes the assertion match the sentence above it.

This matters more than a usual CI nit because `compose-smoke` is the only place anything proves the worker works end to end, and this PR is what turns it from a stub into the real thing.

## 2. The scheduled workflow's runner cost is unaccounted — `.github/workflows/ingest.yml`, and `docs/cost-protection.md`

`ingest.yml` runs `*/30 * * * *`: **48 runs a day, ~1,440 a month.** I confirmed with `gh` that `zarghammir/ai-radar` is **PRIVATE**, so those minutes bill against the Free plan's 2,000-minute monthly allowance rather than being free as they would be on a public repo.

Measured rather than estimated, from your own runs:

| | measured |
|---|---|
| steps `ingest.yml` shares with CI (checkout, node, npm pin, `npm ci`) | **22s** |
| a real ingestion pass, from the compose-smoke log | **9.0s** |
| existing `lint, typecheck, test, build` job | 86s |
| existing `compose-smoke` job | 93s |

GitHub bills whole minutes, rounded up, per job. So each scheduled run is **≥1 minute** — call it ~1,440 minutes a month for the ingest schedule alone, **roughly 72% of the entire free allowance**, leaving ~560 minutes for all CI, while a single push already costs ~4 minutes across the two jobs.

`docs/cost-protection.md` is a genuinely good document and I verified every factual claim in it (below). But all four of its rules are about **paid API calls**, and the word "minutes" does not appear. This PR is the one that introduces a recurring compute cost, in the same change that adds the cost document. Issue #35 does not cover it either — I read it, and it owns `LLM_MAX_STORIES_PER_DAY`, a different axis.

I am not asking you to change the schedule; that is the owner's call. I am asking that the document that exists to make costs visible say what this one is, so nobody discovers it from a billing email. A fifth "what is true today" bullet, with the arithmetic and the allowance, would do it. If 30 minutes is more often than the product needs, hourly halves it.

**A related number worth a sentence in the same place:** 1,440 runs × 18 sources ≈ **26,000 requests a month to other people's feeds**, plus one round of 18 on every pull request from `compose-smoke`. Nothing here is paid, so it is not a billing risk, but it is the kind of thing a publisher rate-limits, and this document is where a self-hoster would look for it.

---

## Smaller

**3. A comment this PR falsifies — `ci.yml`, the worker-pass step.** *"It is a no-op today: the not-implemented path ignores arguments."* `--once` is no longer a no-op: your own run took the real path and exited on one pass. The rest of that comment is still right and still worth keeping; only the last sentence has expired, and this is the PR that expired it.

**4. A negative assertion with no positive beside it — `ci.yml`, the secret-generation step.**
```sh
if grep -q '^INTERNAL_API_SECRET=change-me$' .env; then ... exit 1; fi
```
This proves the value is *not the placeholder*. If the `sed` ever matched nothing — a renamed key, a reordered `.env.example` — the line would be absent entirely and this guard would pass on a file with no secret at all. `grep -qE '^INTERNAL_API_SECRET=[0-9a-f]{64}$'` asserts what you actually want and costs the same.

**5. `readIntervalMinutes` has no lower bound — `src/worker/report.ts:69-77`.** Any positive finite number is accepted, so `INGEST_INTERVAL_MINUTES=0.01` is a pass every 0.6 seconds against 18 third-party feeds. The function already refuses a value it cannot parse *because* "a typo that silently becomes 30 minutes is a worker running on a schedule nobody chose" — a typo that becomes 0.01 is the same argument pointing at a worse outcome. In the PR that adds the cost document, a floor belongs here.

**6. Two states share one message — `main.ts:50` and `trigger.ts:42`.** `!outcome.ran || !outcome.result` reports "another ingest is already running". Today `result` cannot be absent when `ran` is true, so the second half is type-narrowing rather than a reachable branch — I checked. But if it ever became reachable, the operator would be told the lock was busy when it was not, which is the one thing that message must never say wrongly. Narrow on `outcome.ran` and let the type carry the rest.

**7. A failing unlock masks the failure that caused it — `lock.ts:45-49`.** The `finally` is right and the comment explaining it is right. But if `work()` throws and `pg_advisory_unlock` then also throws, the unlock's error replaces the original in the propagated exception, and the real cause is lost exactly when someone is reading logs to find it.

---

## Checked and cleared

- **The worker really is wired into the image — I chased this one hard and it came back clean.** `docker/worker-entrypoint.sh` still carries text saying `src/worker/main.ts` does not exist yet, which reads stale now that this PR adds it, so I went looking for a gap. There is none: the `Dockerfile` bundles the entry conditionally (`if [ -f src/worker/main.ts ]`), the file now exists, and the compose-smoke log at this SHA shows a genuine pass — `18 sources, 496 fetched, 496 new, 482 stories, 1 failed in 9.0s`, exit 0. "The service is wired and will run it the moment that file lands" turned out to be literally true.
- **Every factual claim in `docs/cost-protection.md` verifies.** I ran the document's own commands: only `.env.example` is tracked; no `NEXT_PUBLIC_` in `src/`; exactly one `route.ts` under `src/app`; nothing imports `@anthropic-ai/sdk` (the `anthropic` hits are a topic key and a HN keyword, not an import); `llm_usage` exists in the schema and no code reads it or the cap — which is exactly what the document says, including that the cap is unenforced. A document that states its own gaps plainly is worth more than one that does not, and this one is accurate.
- **The request-time database import is a real fix, and the CI guard around it is sound in principle.** `src/db/client.ts:6` is `if (!connectionString)`, so the empty string `DATABASE_URL: ""` in the Build step is falsy and does throw — the guard is not vacuous by construction.
- **The secret guard.** Refuses empty and refuses the published placeholder; digests both sides before `timingSafeEqual` so neither length nor a buffer-length throw leaks anything; the trigger reads and compares the secret before touching the database, so an unauthenticated caller leaves nothing in `ingest_runs`; the 503 says nothing useful to a caller while logging the real reason server-side; and the only value ever interpolated into a thrown message is the placeholder, which is public by definition.
- **The lock.** `sql.reserve()` for the whole run is the correct answer to a session-scoped advisory lock, and the reasoning in the comment is right. Tests assert held-count 1 while work runs and 0 after, a declined second run whose work never executed, and release after a throw.
- **The tests clean up after themselves.** No `DROP DATABASE` hang, and no test database left behind after my run.

## Verified myself

- Base currency, merge state, file counts both ways, and every step of both CI jobs at `d71f49d`.
- The full suite, run bare with `DATABASE_URL` set: 169 passed, 14 files, 0 skipped, no skip warnings.
- Repository visibility (`PRIVATE`) via `gh`, which is what makes finding 2 a cost rather than a non-issue.
- All timings in finding 2, read from the jobs API on this SHA's own runs, not estimated.
- Which branch the smoke actually took, read from the job log rather than inferred from the script.
- Every claim in `docs/cost-protection.md`, by running its own commands.

## Taken on trust

- **That the `DATABASE_URL: ""` Build step actually reddens** if the route's import moves back to module scope. The logic is sound and I verified the empty string is falsy, but proving the guard bites needs a `next build`, which needs the machine — it is held elsewhere. I will run that control and report it separately rather than leave it implied.
- **The declared final commit.** The `rankAllStories` call is not in this tree; I confirmed nothing in `src/worker/` references it yet. I will review that commit when it lands rather than assume its shape.
- **Behaviour of the loop over days.** I read the signal handling closely and traced `sleep`/`wake` through a signal arriving during a pass, during a sleep and between the two, and it is correct in all three; but I ran no long-lived worker.
- **GitHub's `concurrency` semantics under a live schedule.** `cancel-in-progress: false` queues a pending run rather than dropping it, which is not quite the "skipped, not queued" the comment at `ingest.yml:12-15` describes. I did not execute a scheduled workflow to prove it, so I am flagging it here rather than filing it — worth a look from someone who can.

---

Findings 1 and 2 are the ones I would not merge without: one is the only end-to-end proof the worker works, the other is a recurring bill nobody has written down. The other five are one line each.

zarghammir:
All seven addressed at **`fff4bc3`**. Nothing pushed back on — every finding reproduced against the code before I touched it. Two places where what I found differs slightly from the verdict are called out below rather than buried.

## 1 — the smoke could not tell a real pass from one that did nothing

Confirmed, including the exact mechanism: `docker/worker-entrypoint.sh:11` prints that line in the shell before `exec`, and `main.ts` logs a different line and returns `0` when the lock declines the pass.

**Reproduced before fixing, with real worker output and no Docker.** I ran the worker against a scratch database twice — once normally, once with the advisory lock held from another connection — captured both outputs, prepended the entrypoint's literal line exactly as the container produces it, and fed each to the step's own shell assertion:

```
── the assertion on main ──
  completed pass : ACCEPTED as a real pass
  SKIPPED pass   : ACCEPTED as a real pass      <- the defect
── the assertion now ──
  completed pass : ACCEPTED as a real pass
  SKIPPED pass   : REJECTED
```

The declined run's entire output was:

```
[worker] running one ingestion pass
[worker] another ingest is already running; skipping this pass.
```

It now greps `^total: [1-9][0-9]* sources`. `[1-9]` floors the source count, so a pass over an empty catalogue fails here too, matching how the seeding step already floors its count.

**What that control proves and what it does not:** it proves the shell assertion discriminates, against genuine worker output, which is the part I could run without the machine. It does not exercise the container. The `compose-smoke` run on this push is that check, and it is the one to look at rather than my simulation.

## 2 — the schedule's runner cost

Added as a fifth rule and a section of its own: 48 runs a day, ~1,440 a month, against the Free plan's 2,000 minutes, which apply because the repository is private. The schedule is unchanged — that is the owner's call and the PM is putting the numbers in front of him.

**One refinement, because I could not confirm the rounding.** GitHub's billing page states the 2,000-minute allowance and free minutes for public repositories, but says nothing about per-job rounding on that page. Rather than assert it, the document gives both arithmetics — ~744 minutes a month billed by actual time, ~1,440 billed per job rounded up — and says to assume the larger, because that is the one that can run out. Your conclusion is unchanged either way: it is the largest line in the budget.

The ~26,000 feed requests a month is in the same section, tied to the interval floor added under finding 5, and there is a formula to recompute after a schedule change rather than a number that silently expires.

## 3, 4 — the two `ci.yml` one-liners

The expired sentence is gone. The secret check now asserts `^INTERNAL_API_SECRET=[0-9a-f]{64}$` and prints the offending line. Controlled in three states: a generated secret accepted, the untouched placeholder rejected, and — the case the old check was blind to — the key absent entirely, also rejected.

## 5 — no floor on the interval

One minute, with the reasoning that the interval is a request rate against other people's servers rather than a local loop. Two controls: removing the floor reddens the two refusal tests; changing `<` to `<=` reddens the boundary test that accepts the floor itself.

## 6 — two states sharing one message

`LockOutcome` is now `{ ran: true; result: T } | { ran: false }`. Both call sites narrow on `ran` alone, and a declined outcome carries no result at all rather than an undefined one. The test that asserted `result` was undefined is gone, because that state is no longer representable; it asserts the whole shape instead.

## 7 — a failing unlock masking its cause

Releasing never throws now: it reports its own failure and leaves the original to propagate. A completed run whose unlock failed is still reported as completed, because it completed, and the connection closing releases the lock regardless.

Two tests, driven by a fake reserved connection whose unlock rejects — no database, since what is under test is which error survives. Before the fix the first one failed with `expected 'the ingest itself failed' but got 'connection reset while unlocking'`, which is the finding exactly.

## Your open item on `concurrency`

You flagged this under "taken on trust" and it was worth flagging: **you were right and my comment was wrong.** Checked against GitHub's concurrency documentation — with `cancel-in-progress: false` a run arriving mid-flight is *queued*, not skipped, and any earlier *pending* run is cancelled so the newest takes its place. At most one ever waits, which is the property I wanted, but that is not skipping and a queued run still costs a minute. The comment now says what actually happens, and names the advisory lock as the second mechanism for the case where a queued run starts while the previous is still working.

`docker/worker-entrypoint.sh` also no longer says `src/worker/main.ts` does not exist. The fallback stays — the `Dockerfile` bundles conditionally — and the string your 78-branch greps for is untouched.

## Verification at this head

`npm run lint`, `npm run format:check`, `npm run typecheck` all exit 0. Worker suite: **45 passed, 4 files** (40 before; +3 for the floor, +2 for the unlock). I did not run the full suite or a build locally — the machine is held elsewhere and CI on a clean checkout is the better instrument for both.

Still outstanding and unchanged: the `rankAllStories` commit, which waits on #36.

zarghammir:
REVIEW VERDICT: APPROVE

Re-reviewed at **`781091e`**, not the `fff4bc3` I was handed. The head moved twice while I was starting, and the newest commits are the merge of `main` plus the declared ranker commit — so rather than review an incomplete tree and re-check, I re-baselined on the complete one. Reviewing the merge and the wiring together is the right way round: they are the part that can only go wrong in combination.

Base carries current `main` (0 behind), `MERGEABLE`/`CLEAN`, 20 files, counts agree both ways, both CI jobs green with every real step executed. My own bare run with a database: **204 passed, 16 files, 0 skipped, exit 0.**

**All seven findings are fixed. Two of the fixes are better than what I asked for.** Everything below I verified myself; where I ran a control, the control is named.

---

## Finding 1 — the smoke's assertion. Closed, and proven in both directions.

I promised to check the join between the shell assertion and what the worker actually emits, byte for byte, because that join is what the whole proof rests on. I ran the real `formatTotalLine` against the real regex:

| input | `^total: [1-9][0-9]* sources` |
|---|---|
| `total: 18 sources, 180 fetched, 36 new, 18 stories in 9.0s` | **match** |
| `total: 1 sources, 10 fetched, 2 new, 1 stories in 1.0s` | **match** |
| `total: 0 sources, 0 fetched, 0 new, 0 stories in 0.5s` — empty catalogue | **reject** |
| `[worker] another ingest is already running; skipping this pass.` — the declined pass | **reject** |
| `[worker] running one ingestion pass` — the entrypoint's launch line | **reject** |

The last two rows are the defect. The old assertion accepted both; this one takes neither. The third row is the `[1-9]` floor doing work the finding did not ask for.

**And it is armed in the container, which is the half a simulation cannot reach.** I read both runs rather than the ticks:

- **Negative — run `35163983665`**, branch `control/smoke-can-fail`: `compose-smoke` red **at "Run one worker pass"**, with every prior step green (image built, postgres healthy, web 200, catalogue seeded), the entrypoint's line present in the output, and `worker exit code: 0`. That is exactly the state that used to be reported as success.
- **Positive — this PR's own run**: `total: 18 sources, 497 fetched, 497 new, 483 stories, 1 failed in 9.9s`, exit 0, `RESULT: the worker ran a pass.`

A gate known only to pass is not known to work. This one is now known to fail for the right reason and pass for the right reason.

**On the correction that came with it:** the claim that lint and typecheck stayed green on the control branch so nothing else could produce the red is **not true**, and I found that independently before I was told. Both jobs failed — `compose-smoke` at the intended step, and `lint, typecheck, test, build` at **Unit tests**, because a worker deliberately broken to never ingest fails the test asserting that a correct secret causes an ingest. The control is untouched by this: its claim is about which step rejected which output, and that step rejected that output. The PR body already carries the honest version and names the superseded draft, which is the right way to correct a claim. It is also, incidentally, a small piece of evidence for the unit tests: the same break was caught twice, by two independent instruments.

## Finding 2 — the runner cost. Written down better than I asked.

The new section carries the private-repo fact, the 2,000-minute allowance, the measured timings, **and both arithmetics** — ~744 minutes by actual time, ~1,440 billed per job rounded up — with "assume the second until someone confirms otherwise, because it is the one that can run out", plus a formula to recompute after a schedule change.

That is more careful than my finding was. **I asserted per-job rounding as fact; it was my own inference and I was not entitled to it.** I tried to settle it afterwards through the Actions timing API, which reports billable minutes per job, and it is not available for these runs — so I cannot close it either. Carrying both numbers and naming which to assume is the correct treatment of a mechanism nobody has confirmed. The conclusion is unchanged on either arithmetic: the schedule is the largest line in the budget by a wide margin.

The `≈26,000 requests a month to other people's servers` line is the other half of the same clock and I am glad it is there.

## Findings 3–7

| # | fix | how I checked it |
|---|---|---|
| 3 | the `--once` comment no longer says "no-op today" | read; the run proves the flag is load-bearing |
| 4 | `grep -qE '^INTERNAL_API_SECRET=[0-9a-f]{64}$'` replaces the negative-only check, with a diagnostic that prints the line or says there is none | read |
| 5 | `MINIMUM_INTERVAL_MINUTES`, with a message naming the reason | **control:** removing the floor reddens exactly the two named floor tests |
| 6 | `LockOutcome<T>` is now `{ ran: true; result: T } \| { ran: false }` | both call sites narrow on `ran` alone; there is no second condition left to get wrong |
| 7 | `releaseLock` catches and logs, so a failing unlock cannot replace the failure that caused it | read |

**Finding 6 deserves singling out.** I asked for the narrowing to be tidied. You made the bad state unexpressible instead — a discriminated union cannot report "another ingest is already running" for any other reason, because there is no other shape to be in. That is the better fix and the one this project keeps reaching for correctly.

## The item I had left under "taken on trust" is closed against the branch

I flagged that `cancel-in-progress: false` **queues** rather than skips, and that the comment said otherwise, without being able to prove it. The comment now says at most one run ever waits, that this is **not the same as skipping**, that a queued run still costs a minute, and names the advisory lock as the second mechanism for a queued run that starts while the previous is still working. Corrected in the direction I flagged.

## The ranker commit and the merge

Both claims I was asked to test hold:

- **`rank-all.ts` and `score.ts` are byte-identical to `main`.** Imported and called, nothing more.
- **`run.ts` is identical to `main` too** — the `tagTopics` parameter landed via #36's branch, so it is no longer this PR's change at all.

The wiring is inside `ingestOnce`, so both worker modes and the HTTP trigger get it from one path. **Control: replacing the `rankAllStories` call with a zero result reddens exactly two tests — the two ranker tests — and leaves the other 202 green**, with zero new type errors. That matches the lane's report.

**The test meets the corrected acceptance, and says why it had to.** `stories.score` is `real().notNull().default(0)` — I read it in the schema — so "every story has a non-null score" is true before anything runs and true of no stories at all. The test floors `itemsInserted`, floors `ranked.ranked`, floors the row count, asserts the score **components** which only a ranked story has, and separately checks the ranker's own count against the table. A count that agrees only with itself is worth nothing, and this one does not.

I also note the merge produced a real semantic problem that a clean `git merge` could not show — Next's `global.d.ts` redeclaring `NodeJS.ProcessEnv` without an index signature, breaking the injectable `SecretEnv` — and that it was found by running the checks after merging rather than by trusting the absence of conflict markers. That is the correct habit and the reason to run them.

## Verified myself

- Base currency, merge state, both file counts, and every step of both CI jobs at `781091e`.
- The full suite, bare, with a database: 204 passed, 16 files, 0 skipped.
- Four controls, each reverted, each gated differentially against the clean tree's one known error (this worktree has no build, so `tsc` reports `LayoutProps`): the ranker call, the interval floor, **the secret comparison** — which reddens all three rejection tests including *rejects a secret that is merely a prefix of the real one*, so the constant-time comparison is genuinely pinned — and the regex join above.
- Both control runs for the smoke, read from their job logs and step lists rather than from their conclusions.
- That `rank-all.ts`, `score.ts` and `run.ts` are identical to `main`.
- `stories.score`'s schema definition, which is what makes the corrected acceptance necessary.

## Taken on trust

- **The `DATABASE_URL: ""` Build step.** I still have not proven it reddens when the route's import moves back to module scope. It needs a `next build` and the machine is held elsewhere. It is queued, I will run it, and **I will report it as its own note rather than folding it silently into this approval** — if it turns out not to bite, that is a finding after the fact and I will say so.
- **The HTTP surface over the wire.** `POST /api/internal/ingest` is verified at the handler level — four tests, including the three rejections above — and the route file is a three-line wrapper I read. I did not drive it over HTTP against a running server.
- **The loop over time.** I traced `sleep`/`wake` through a signal arriving during a pass, during a sleep, and between them, and it is correct in all three; I ran no long-lived worker.
- **Real billing.** Every timing in the cost section is measured, but what GitHub actually charges is not something I can observe from here.

---

Approved. The one control I owe is on a CI guard rather than on the product, and it is queued rather than forgotten.

zarghammir:
REVIEW VERDICT: APPROVE

Re-posted at **`d1e4fd4`**. My previous approval named `781091e`, and the branch has moved since. **Click this one.** Nothing in the earlier verdict changes; this comment exists so the SHA the owner clicks is one a reviewer has named.

## I did not take "it is only a comment" on trust

It is fourteen added lines of JSDoc in `src/worker/ingest.test.ts`, and I proved it is inert rather than reading it and agreeing:

| check | result |
|---|---|
| executable text, comments stripped, `781091e` vs `d1e4fd4` | **identical** — 3846 characters both sides |
| control: does that comparison detect a change? | **yes** — renaming one test string is caught |
| `it()` blocks | 2 before, 2 after |
| `it.skip` / `it.only` / `describe.only` introduced | none |
| files touched | one, and it is a test file |

The one `describe.skip` the scan finds is `const withDb = configured ? describe : describe.skip`, present at both heads and not from this delta. I checked what protects it rather than assuming: `ingest.test.ts:20` throws `"CI must set DATABASE_URL; refusing to skip the ingest pass tests silently"` when `CI` is set and the URL is not, and `ci.yml:43` sets a job-level `DATABASE_URL` against a real postgres service. **So a green CI is itself proof these tests ran** — the gate cannot pass by running nothing. CI executed **16 test files** at this head, matching my local run exactly.

## The comment itself is worth having

It records that the seven-day ranking window belongs to `rank-all.ts` and is asserted by that module's own suite, and that pinning it from the worker's file would keep passing on the day the rule changed. That is the same reasoning that made the mislabelled control in #36 worth reporting: a test that asserts another module's rule from outside reads as coverage while covering nothing. Recording why a test is *absent* is rarer than recording why one exists, and more useful.

It also explains a number a future reader would otherwise trip on — that a real pass reports fewer stories scored than created, because the window excludes the old ones.

## Everything from `781091e` carries

All seven findings fixed and verified there, unchanged here: the smoke assertion proven in five directions against the real `formatTotalLine` and armed in the container in both directions; the runner cost written with both arithmetics; the interval floor, the positive secret assertion, the discriminated `LockOutcome`, the unlock that no longer masks its cause; and the ranker commit wired on the one shared path, with `rank-all.ts`, `score.ts` and `run.ts` byte-identical to `main`.

Four controls, each reverted, each differentially gated, all still valid at this head because the executable text is unchanged: the ranker call, the interval floor, the secret comparison, and the regex join.

## Still owed by me

The `DATABASE_URL: ""` Build step control — proving it reddens when the route's import moves back to module scope — needs a `next build` and a machine that is held elsewhere. It is queued, and **it gets its own note whichever way it goes.** If it turns out not to bite, that is a finding after this approval and I will post it as one.

Approved at `d1e4fd4`.

zarghammir:
**The build control I owed on this PR, reported on its own as promised — and it bites.**

In my approval I listed one thing under "taken on trust": that `ci.yml`'s `DATABASE_URL: ""` on the Build step actually reddens if the route's database import moves back to module scope. I said I would run it and report it either way rather than fold it into an approval. The machine came free; here it is.

| | `DATABASE_URL="" npm run build` |
|---|---|
| **as it stands** — `const { db, sql } = await import("@/db/client")` inside `POST` | **exit 0** |
| **control** — the same import hoisted to the top of the file | **exit 1** |

The control's failure is the exact mechanism the step's comment describes, not an incidental error:

```
Error: Failed to collect configuration for /api/internal/ingest
  [cause]: Error: DATABASE_URL is not set. Copy .env.example to .env and edit it.
  >  7 |   throw new Error("DATABASE_URL is not set. …");
```

`next build` collects page data for every route module, `src/db/client.ts:6` throws on a falsy connection string, and an empty string is falsy — so the guard catches the regression it was written for, seconds into CI rather than two minutes into an image build. Reverted immediately; the tree is clean.

That closes the last open item from my review. Nothing in the approval changes — it only stops resting on my word.

---

One note from running it, which is mine rather than the branch's. My first attempt failed before any mutation, with `Symlink [project]/node_modules is invalid, it points out of the filesystem root`: I had symlinked this worktree's `node_modules` to another checkout's, and Turbopack refuses that. **The branch was never involved** — the baseline failing is what told me the instrument was wrong rather than the code, which is the only reason it did not become a false finding. Replaced with a real `npm ci`, after which the baseline passed and the control behaved as above.

zarghammir:
**Approval carries to `2a3469e`.** Click this one.

I judged the merge rather than the hunk, since it brings #31's and #33's whole trees onto the branch, and the check that settles it is stronger than the lockfile comparison:

**This branch differs from current `main` in exactly 20 files, and every one of them is `feat/worker`'s own** — the nine `src/worker/*` files, the route, `client.ts`, both workflows, the entrypoint, `.env.example`, `README.md`, `docs/cost-protection.md` and `package.json`. Nothing else differs, which means everything `#31` and `#33` landed arrived intact *by construction* rather than by inspection. `package-lock.json`, `next.config.ts`, `docs/api.md`, `src/app/layout.tsx` and `public/sw.js` are byte-identical to `main`.

- `package.json` is the only one worth a second look, and it adds **two scripts** (`worker`, `worker:once`) and nothing else: dependencies 14 → 14, devDependencies 16 → 16, no version changes, no removals. So the lockfile being identical is not a coincidence to check but a consequence — there is no dependency change for it to record, and the `npm ci` condition holds trivially.
- The resolved hunk in `src/db/client.ts` is additive: `export { sql }` and its comment, nothing else, and the file is otherwise identical to `main`.
- Carries current `main` (0 behind), `merge-tree --write-tree` against `main` exits **0**, GitHub reads `MERGEABLE`/`CLEAN`.
- CI at this exact SHA: both jobs green with every real step executed — Lint, Typecheck, Unit tests, Build, Format check, and in the smoke the image build, the 200 wait, the seed confirmation and a real worker pass.

The PR body is right that local light checks would have failed here for an environmental reason rather than a real one, and that CI on the resolved tree is the instrument. That is the same point I had to correct in my own first verdict on #31: **a claim about what CI covers is a property of the tree at a SHA**, and this tree is new. It is the tree CI ran on.

Nothing in the approval at `d1e4fd4` changes. Everything it rested on — the smoke assertion proven in five directions and armed in the container both ways, the four deletion controls including the prefix-secret one, the discriminated `LockOutcome`, the ranker wired on the one shared path — is untouched by a merge that added no work of its own. The `DATABASE_URL: ""` build control I owed is closed in its own comment above.

One thing for whoever merges next, not for this PR: `src/db/client.ts` is about to be rewritten by #42, whose lazy shape has no place for the raw `sql` client this PR's advisory lock needs. That is settled — #37 merges first and #42 rebases onto it and owns the reconciliation — but it means the next change to this file needs a test proving the lock still takes and releases on **one** reserved connection, not merely that it compiles.

zarghammir:
**Approval carries to `d4eef2b`.** Click this one — it supersedes my line at `2a3469e`.

The delta is **one file, `docs/api.md`**, and nothing else: 24 insertions, 16 deletions across the two sites that made the claim. No source, no workflow, no test.

**The correction is right, and it is the right PR to make it in.** `docs/api.md` said twice and unconditionally that no route triggers ingestion and no route makes an external call — once as a ground rule, once as a deliberately-absent row. Neither was wrong when written. This PR's `POST /api/internal/ingest` is what made them false, so this is where they get fixed. Nothing in a diff could have shown it: the file did not change, the world did. That is the same class as the eight stale sentences in #33, and finding it needed someone to re-read a merged document against a branch rather than review a change.

**I checked the load-bearing sentence rather than the wording.** The new text claims the internal route *"refuses before it reads or writes anything — an unauthenticated caller leaves no row in `ingest_runs` and costs nothing to turn away."* That is a property, not a description, and it holds:

- **Order, in `src/worker/trigger.ts` at this SHA**: `readInternalSecret` → 503 on failure → `secretMatches` → 401 on failure → **and only then** `ingestOnce`. Nothing reads or writes the database before the refusal.
- **It is asserted, not just documented.** `trigger.test.ts` has a `runCount()` helper over `ingest_runs`, and both rejection tests end in `expect(await runCount()).toBe(0)` — literally the sentence in the document.
- **The zero is not vacuous**, which is what I would otherwise have asked: *"runs the ingest when the secret is right"* asserts `runCount()` is **1**, and its comment says exactly why it exists — *"without it, a route broken in both directions would satisfy every rejection test above."*
- I ran the deletion control on this in my approval: removing the `secretMatches` comparison reddens all three rejection tests, including *"rejects a secret that is merely a prefix of the real one"*, so the guard those zeros depend on is genuinely pinned.

The rest of `docs/api.md` is untouched, so everything I verified when approving the contract in #33 — the 22 properties, the key and label claims against merged code — is unaffected.

Nothing in the approval at `d1e4fd4`, or in my confirmation of the merge at `2a3469e`, changes. CI is green at this SHA.

---

**One thing for #42 rather than for this PR, found while checking the sentence above.** `docs/cost-protection.md:42-44` — which this branch owns — says:

> **The app has exactly one API route**, `POST /api/internal/ingest` … There is no public route to protect yet, because there is no public route.

That is **true here**: this branch has exactly one `route.ts`. #42 adds twelve, so it becomes false the moment #42 lands, and #42 does not currently touch that file. By the rule this very commit establishes — the PR that makes a statement false is the PR that corrects it — that correction belongs in #42's rebase, and I will look for it there.

It is worth more than a wording fix. Rule 4 of that document is *"No public route reaches a paid service"*, and today it is satisfied **vacuously**, by there being no public route at all. Once twelve exist, the rule stops being trivially true and starts being a claim someone has to check. I have already checked it for #42's routes — 21 files, with a floor and a working control: nothing in the route path fetches, ingests, ranks or touches `@anthropic` — so the rule still holds. But the document's stated *reason* for holding will be gone, and a rule that is true for a reason the document no longer gives is one nobody can re-verify from the document.

The scoping is also just more honest than what it replaces. "No route makes an external call" was a promise the codebase had outgrown; "no route a client can reach" is a promise it can keep, with the exception named, guarded, and pointed at the worker's documentation rather than hidden.

PR #39, opened 2026-09-16
[CONTROL — DO NOT MERGE] prove compose-smoke can fail

zarghammir:
Throwaway control for #37, finding 1.

The worker is broken to return "declined" without doing anything: it starts cleanly, prints the skip line, and exits 0 without ever printing a `total:` line. That is exactly the state the compose smoke's assertion must reject, and until now that assertion had only ever been seen to pass.

Expected: **compose-smoke RED** at the "Run one worker pass" step with `worker exited 0 without completing a pass`.

Lint and typecheck pass on this branch on purpose, so the red can only come from the assertion. Will be closed and the branch deleted as soon as the red is recorded.


zarghammir:
Control complete and recorded on #37.

Run id **35163983665**, job 105020830766: compose-smoke RED at "Run one worker pass" with `worker exited 0 without completing a pass`.

The red is from the assertion and nothing else — the image built (231 MB), postgres came up healthy, the container started, the entrypoint printed its line and the worker exited 0. Exactly the state that used to be reported as a successful pass.

Closing; branch deleted.

zarghammir:
Correction to this pull request's description, for anyone who reads it as evidence.

The description said lint and typecheck pass on this branch "so the red can only come from the assertion". That was true of the local runs it was written from and **not** true of the CI run: **both** jobs failed here.

- `compose-smoke` — the intended red, at *Run one worker pass*: `worker exited 0 without completing a pass`.
- `lint, typecheck, test, build` — also red, at *Unit tests*: `trigger.test.ts > runs the ingest when the secret is right`.

The second is the suite working correctly — a worker deliberately broken to never ingest fails the test asserting that a correct secret causes an ingest. The control still stands, because its claim is that compose-smoke's assertion step rejected a worker that exited 0 without doing a pass, and it did. But "nothing else could go red" was the strong form and the strong form was wrong.

Recorded on #37 as well.

PR #42, opened 2026-09-17
API implementation, and the client reconciliation with the worker

### LINKED ISSUE #10: API contract for the web app and future native clients

zarghammir:
Closes #10.

The seam between the data and anything that shows it: twelve routes, built against the contract in #33 rather than invented alongside it. Nothing here fetches a feed, runs the worker, or calls anything outside this machine — opening a page cannot cost money, and that is now checkable rather than asserted.

It also carries the reconciliation with #37. Both branches rewrote `src/db/client.ts`, each was clean against main and neither against the other, and nothing compares two open PRs. #42 owns the resolution because #42 introduced the lazy shape.

## The routes

`GET /api/brief` · `GET /api/radar` · `GET /api/radar/histogram` · `GET /api/stories/:slug` · `GET /api/saved` + `POST`/`DELETE /api/saved/:storyId` · `POST /api/read/:storyId` · `POST /api/hide/:storyId` · `GET`/`PUT /api/preferences` · `GET /api/topics` · `GET /api/sources` · `PUT /api/sources/:key`

All fourteen documented method-and-path pairs exist, and no route exists the contract does not document. Checked mechanically.

## Explicit accessors, deliberately not a Proxy

`getDb()` and `getSql()`. The proxy that was here before was measured before it was dropped: `'select' in db` was **false**, `Object.keys(db)` was **empty**, and `db.select !== db.select`.

Completing those traps is not the fix. `has` and `ownKeys` must answer from the real handle, which means **connecting in order to answer them** — so enumerating this module's exports would connect to a database, which is the build failure this PR exists to fix wearing a different hat. A function cannot be enumerated into a connection, has stable identity, and has no trap semantics to get subtly wrong.

**The memoisation is load-bearing and invisible to types.** A `getSql()` handing out a fresh client per call typechecks perfectly and would give the worker one pool to run on and another to close. It is asserted directly, and removing it reddens exactly that test.

### A control named for a property it does not test

Reported rather than glossed, because removing exactly this shape is most of what this branch has done today.

The fresh-client-per-call control was specified as guarding the **lock**. It does not. Run, it reddens only the identity assertion — advisory locks are session-scoped, so two pools still produce exactly one winner. I then tried the other obvious variant, taking the lock on the pool rather than a reserved connection, and **that reddened nothing either**, because sequential calls reuse one pooled connection anyway.

So the control guards the memoisation, which is real and worth having, and does not guard the lock. That is a measurement, not an opinion.

**Which is why a test was removed.** #37's `lock.test.ts` already covers the lock's semantics thoroughly, including "holds and releases on one connection even when the pool frees others mid-run". My release test duplicated it and no control could make it fail — padding that reads as proof. What remains asserts only the narrower thing this change could break: that the accessor hands back a client on which `reserve()` and the lock still work. Borrowing someone else's coverage and calling it mine would have been the worse option.

## #37's call sites, adapted and nothing else

Five lines each way across `src/worker/main.ts` and `src/app/api/internal/ingest/route.ts`, so the diff reads as "the shape changed underneath this". `lock.ts` needed nothing — `withIngestLock` already takes `sql` as a parameter. That route's request-time import is Platform's deliberate choice and is untouched.

## A cost rule that can be checked

`docs/cost-protection.md` said the rule held because there was no public route at all. True — and true for a reason that has now gone.

Finding it was not a wording exercise. I wrote the check and **my own routes failed it**: nine of twelve transitively imported three feed adapters and the HTTP client, through one import line, because `scoreComponentList` lived in the ingest module. Nothing called them, so the rule held — but it could only be confirmed by tracing call sites, which is true until the next refactor and re-checkable by nobody.

`scoreComponentList` moves to `score.ts`, beside the label map it uses and where it always belonged. All twelve routes are now clean, so **the graph is the evidence**.

`npm run routes:check` walks the import graph out of every public route and fails if one reaches an adapter, the HTTP client or an LLM SDK.

- **No exceptions list.** A carve-out is a hole with a comment on it. The public/internal boundary is the `internal/` path segment — a rule about where a route lives, so a new public route cannot inherit an exemption.
- **Type-only edges excluded**, because `import type` is erased and cannot reach anything. That is modelling what an import is, not excusing one; a value import of the same module is still a violation and still named.
- **It prints how many routes it examined**, and a walk that finds too few fails. A checker that walks nothing reports no violations.

Controls, measured unpiped: an adapter edge on one route **fails naming that route**; a walk finding one route **fails on the floor**.

**The script is run by hand.** Issue #17 owns wiring it into CI. An unwired check that claims to be a gate would be the third instance of that shape today.

## Four things Next 16 settled

`params` is a Promise and must be awaited. Route handlers are not cached by default, and `force-static` must never be added here because these responses carry a reader's own saved and read state. The handlers take a plain `Request`, since none needs a Next-only API. And a `Date` interpolated into a raw drizzle `sql` template reaches the driver unserialised.

## Tests

**257 passing across 20 files.** Differential mutation controls throughout: each compares the tsc error set against a clean baseline and requires zero new errors, so no red can come from code that does not compile.

Three things the controls found that the tests did not: the brief never asserted it excludes hidden stories, nothing asserted a card de-duplicates its sources, and `user_preferences` was missing from a truncate, so one test's brief length leaked into another's default.

| command | exit |
|---|---|
| `npx tsc --noEmit` | 0 |
| `npx vitest run` | 0 |
| `npx eslint .` | 0 |
| `npm run routes:check` | 0 |

Both database suites create, migrate and drop their own database, each with its own name because vitest runs files in parallel, and both refuse to skip under `CI`.


zarghammir:

Reviewed at **`e1a61f4`**, the rebased tree, re-verifying rather than carrying anything over — the base moved onto merged #37, so everything from my static pass was re-run here. Carries current `main` (0 behind), `MERGEABLE`/`CLEAN`, 36 files both ways, 13 routes, both CI jobs green. My own bare run: **257 passed, 20 files, 0 skipped.**

**The implementation is correct and I could not break it.** Two findings are one line each, in the same file, and both are about the *same* load-bearing property being described and tested wrongly. The third is a small hole in a new checker. I would approve on sight.

---

## The control we all specified does not test what we said, and I confirmed it independently

You reported it rather than letting it stand, and you were right. I re-ran it: replacing `current()` with a fresh `connect()` per call reddens **only** *"hands back the same client and the same database every time"* — 1 failed, 1 passed, zero new type errors. The lock test stays green.

Then I checked *why*, because "the control didn't fire" is a claim about the world and not just about the test. Two genuinely distinct `postgres()` clients, concurrent, through the real `withIngestLock`:

```
clientsAreDistinct: true
firstRan: true    firstWorkExecuted: true
secondRan: false  secondWorkExecuted: false
→ exactly one winner across two separate clients
```

`pg_try_advisory_lock` is **database-global**, not pool-local. Two pools are two sessions, and the second session is refused exactly as a second connection in one pool would be. So the memoisation cannot be what keeps the lock single-writer, and no control on the lock could have shown otherwise.

**So `src/db/client.ts:45-50` states a rationale that is false**, and this is finding 1:

> *"…would give two callers two pools, both would reserve successfully, and **both would believe they hold a lock** that is meant to admit one."*

The first half is true — both reserve. The second is not: the second caller's `pg_try_advisory_lock` returns false and its work never runs. I measured it.

The memoisation **is** load-bearing, and the real reason is one line up in the same file. `connect()` populates `globalThis.__aiRadarSql` only when `NODE_ENV !== "production"` (`client.ts:32`), so in production a non-memoised accessor builds a **new pool on every call** — connection exhaustion, and a worker holding one pool while the process closes another. That is worth protecting and worth saying. The lock is not the reason, and leaving it as the reason is how someone later concludes the lock is safe because the client is memoised and removes the `reserve()`.

## Finding 2 — the assertion protecting the `sql` half cannot fail where it runs

`client.db.test.ts:53-54`:

```ts
expect(getSql()).toBe(getSql());   // line 53
expect(getDb()).toBe(getDb());     // line 54
```

Under the fresh-client control, the failure is at **`client.db.test.ts:54`**. Line 53 **passes** with the memoisation removed. I proved the mechanism rather than inferring it, by running `connect()` twice with no memoisation at each `NODE_ENV`:

| `NODE_ENV` | `getSql() === getSql()` with no memoisation | |
|---|---|---|
| `test` — what vitest runs as | **true** | the assertion passes anyway |
| `production` | **false** | the assertion would fail, as intended |

`connect()` reuses `globalThis.__aiRadarSql` in non-production, so both calls hand back the same underlying client whatever `current()` does; only `drizzle()` builds a fresh wrapper, which is why line 54 still catches it. **The entire control is carried by the `getDb()` line, and the `sql` line contributes nothing in the environment it runs in.**

Two things make that worth a line of work rather than a shrug. The masked half is the one the comment says matters — *"would put the advisory lock on a pool nothing else uses"* — and the environment where the masking stops is **production**, which is exactly the case that would bite. Stubbing `NODE_ENV` to `production` for that assertion, or deleting `globalThis.__aiRadarSql` between the two calls, makes line 53 able to fail.

## Finding 3, non-blocking — `routes:check` misses a bare side-effect import

The checker is good and I want to say so before the hole: given a normal value import it fails **naming the route and the whole chain**, which is more than I would have asked:

```
FAIL saved/route.ts can reach:
       src/sources/arxiv/adapter.ts via src/api/radar.ts
       src/sources/http.ts via src/sources/arxiv/adapter.ts
```

And the floor bites: with the routes moved out of the tree it exits 1 with *"examined only 1 public routes, expected at least 8 — the walk found too little to be believed."*

But a bare `import "@/sources/arxiv/adapter";` in `src/api/radar.ts` passes — **exit 0, "OK 12 public routes"**. The script's own reasoning is that `import type` is skipped because it is erased and "cannot reach anything"; by that same model a bare side-effect import is a value import that executes module code, so this is an accidental exception in a checker that deliberately refuses to have an exceptions list. The practical risk is small — a bare import cannot call an adapter unless the module has top-level side effects — which is why it is not blocking. One regex.

---

## Two of my own controls were wrong before they were right, which is the useful part

My first `routes:check` controls **both came back exit 0**, and the honest reading of that was not "the checker is broken":

- I renamed route directories to `_radar` and the checker still reported 12. `walk()` collects any `route.ts` under `src/app/api` regardless of directory name, so renaming removes nothing. I had to move the files out of the tree entirely.
- I used a bare side-effect import as my "adapter edge", which happens to be the one form the checker misses — so my control was testing a hole rather than the property, and reported a pass for the wrong reason.

Concluding "the gate does not work" from either would have been a false finding against a gate that works. **The tell was control 2: it reported 12 routes after I had moved nine directories, and a number that cannot have been right is a result rather than a pass.** The gap in finding 3 is real, but I only know that because I re-ran it deliberately, both ways, after fixing the instrument.

## Re-confirmed on this tree, not carried over

- **The derived truncate is still genuinely derived**: 10 tables, and adding a `probe_two` table to `schema.ts` with no edit to `tables.ts` takes it to 11. Covers `saved_items`, `read_state` and `user_preferences`.
- **The sort control still flips the real column** — `update stories set score = 90 where slug = 'older'` — so it proves the sort reads what the ranker writes.
- **saved/read/hide still assert through the database**, `select read_at, hidden from read_state where story_id = …`, not through the API's own echo.
- **`docs/cost-protection.md` is corrected**, and better than I flagged. I raised that "there is no public route" was about to become false; the rewrite says twelve exist and none can reach a paid service, and makes it *checkable* — *"used to hold because there was no public route at all — true, but true for a reason that has now gone."* Replacing a vacuous justification with an executable one is the right move, and *"a carve-out is a hole with a comment on it"* is the right reason to have no exceptions list.
- **Cutting the duplicated lock-release test was right.** It duplicated `lock.test.ts` including the pool-churn case, and no control could redden it. The remaining test says so itself, and *"calling it proof of the lock itself would be borrowing someone else's coverage"* is exactly the standard.

## Verified myself

Base, merge state, file counts both ways, CI steps; the full suite; the fresh-client control and *which* assertion carries it; the two-client lock experiment; the `NODE_ENV` masking mechanism; both `routes:check` controls, twice, after fixing my own; the derived-table control; and `routes:check` run unpiped with its exit status reported rather than its output.

## Taken on trust

- **Production behaviour of the pool.** I reasoned the non-memoised production case from `client.ts:32` and proved the identity consequence at `NODE_ENV=production` in a harness; I did not run this application in production.
- **The twelve routes against a browser.** They are covered by 257 tests including route-level ones, and I read the handlers, but I did not drive them over HTTP from a client.
- **The lane's timings**, given the disclosed unslotted install and build. Nothing in my own numbers looks contended, and every result above is from my own run.

Three one-line fixes and I will approve in a new comment. The behaviour is right; it is the two sentences describing *why* it is right, and one assertion that cannot fail, that I am asking for.

zarghammir:
REVIEW VERDICT: APPROVE

Verified at **`fa28c7f`**. All three findings fixed, each confirmed by a control I ran rather than by reading the diff. Carries current `main`, `MERGEABLE`/`CLEAN`, both CI jobs green with every real step executed. My own run after a clean install: **258 passed, 20 files, 0 skipped**, `npm run typecheck` **0**, `eslint` **0**, `prettier` **0**, `routes:check` **0**.

| finding | control | result |
|---|---|---|
| 1 — the memoisation rationale | read against my own measurement | replacement is **true** |
| 2 — `getSql()` identity unfalsifiable in test | drop the memoisation | **both** assertions now red, at `:50` and `:76` |
| 3 — bare side-effect import missed | bare import, then `import type` | bare **fails naming the route**; type still **ignored** |

## Finding 1 — the replacement is true, which was the thing to check

It is the second rationale written into that file today, so I checked it against what I measured rather than against whether it reads well. Every clause holds:

- *"pg_try_advisory_lock is database-global rather than pool-local, so two separate clients are two sessions and the second is refused"* — that is exactly my two-client experiment: first acquires and runs, second declined, its work never executed.
- *"globalThis is only populated when NODE_ENV !== 'production', so in production a non-memoised accessor builds a NEW POOL on every call"* — `client.ts:32` guards the write, and I measured the consequence directly: two `connect()` calls with no memoisation give the same client under `test` and **different clients under `production`**.

Keeping the false claim on the record with *why* it was dangerous — "would have led someone to conclude the lock is safe because the client is memoised and then remove the reserve()" — is worth more than deleting it. The next person to read that file learns the trap instead of rediscovering it.

## Finding 2 — fixed, and it exposed something I did not have

Dropping the memoisation now reddens **both** assertions — `client.db.test.ts:50` (`getDb`) and `:76` (the production `getSql`) — 2 failed, 1 passed, zero new type errors. Previously only `:50` fired.

**The hazard it surfaced is the more valuable half, and I verified the mechanism rather than taking it.** `connect()` **reads** `globalThis.__aiRadarSql` unconditionally and only skips **writing** it in production (`client.ts:29-32`). So a test that stubs `NODE_ENV=production` and calls the accessor gets the *shared* client back — and ending it kills the pool for every test after it. That does not present as a failed assertion; it presents as **the next test hanging**, which reads as a stuck machine rather than a defect.

The fix saves the cached client, deletes it, stubs, and restores in a `finally`, so it only ever ends a pool it created itself. That is the right shape: a test that damages a process-wide singleton lands its damage on its neighbour, and the neighbour has no way to know.

## Finding 3 — fixed in the direction it was broken, without overcorrecting

A bare `import "@/sources/arxiv/adapter";` now exits 1 and names the route and the chain. I paired it, because fixing "catch bare imports" is exactly the change that could start flagging erased type edges: `import type { ArxivItem } from "@/sources/arxiv/adapter"` still exits **0**. The deliberate skip is intact and the accidental exception is gone.

## The instrument note I would keep from this review

**My first two controls on `routes:check` both came back exit 0, and neither was evidence the checker worked.** One renamed route directories, which removes nothing because `walk()` collects any `route.ts` under the root; the other used the single import form the checker missed, so it exercised the hole rather than the property. What caught it was not the exit code but the **content** of a passing result — 12 routes reported after nine directories had been moved. A number that cannot be right is a result, not a pass.

That is the same shape as the three false signals my own environment produced today, the last one on this PR: `npm run typecheck` exited 2 here with `Cannot find module 'lucide-react'`, because my `node_modules` predated the rebase that brought #31's dependencies in. CI's Typecheck step was green at the same SHA. I reinstalled and it exits 0. **None of that belonged in a verdict**, and the rule that kept it out is the same one: when your instrument and CI disagree at the same SHA, suspect the instrument first.

## Everything else, re-confirmed at this head

The derived truncate is still genuinely derived — 10 tables, 11 after adding a probe table with no edit to `tables.ts`, covering `saved_items`, `read_state` and `user_preferences`. The sort control still flips `stories.score` directly. saved/read/hide still assert through the database rather than the API's echo. `routes:check` reports **12 public routes examined**, and its floor fails at *"examined only 1 public routes, expected at least 8"*.

## Taken on trust

- **Production pool behaviour.** I proved the identity consequence at `NODE_ENV=production` in a harness and read the guard at `client.ts:32`; I did not run this application in production.
- **The twelve routes over the wire.** 258 tests including route-level ones, and I read the handlers, but I did not drive them from a browser.
- **The lane's timings**, given the disclosed unslotted install and build — every number above is from my own run.

PR #44, opened 2026-09-17
Docker image carry-ins: public/, worker restart policy, cached install layer

### LINKED ISSUE #34: Docker image carry-ins from PR #29, after #29 and #5 both land

zarghammir:
Closes #34. The three carry-ins from the PR #29 review that were waiting for both #29 and the worker (#5) to land, plus the assertion that keeps the first one honest.

## In plain words

**What is happening in the records.** Nothing. No data is touched. This changes how the app is packaged into its container and how the container behaves when something goes wrong.

**The flow.** A self-hoster runs `docker compose up`. Three things are different afterwards: the app's icons and offline page actually load instead of coming back missing; the part that fetches news restarts by itself if it crashes, instead of quietly staying dead; and rebuilding the image after a code change is much faster, because it no longer re-downloads every dependency each time.

**Who.** Anybody self-hosting AI Radar, and anybody waiting on a rebuild.

**What they see.** The installable app looks right on a phone instead of showing a blank icon, and the offline page appears when there is no connection. Otherwise nothing visible — this is packaging.

**What it is for.** The container was shipping without its `public/` folder, which holds the icons, the offline page and the service worker. Every existing check passed anyway: pages rendered, the worker ran, the smoke went green. That is the shape of defect worth closing at the source rather than discovering from a phone.

**Where it stands.** Waiting on CI. Not a click yet.

## The four items

1. **`COPY public/` into the runtime image.** Next's standalone output omits it deliberately — its documentation assumes a CDN serves it — and nothing serves it here.
2. **Worker restart policy.** `restart: unless-stopped`. It loops now rather than running one pass, so a crash should not quietly end ingestion, and a deliberate stop is still respected.
3. **`COPY . .` after `npm ci`.** Only `package.json`, `package-lock.json` and `.npmrc` are copied before the install, so editing a component no longer invalidates it.
4. **The static-asset assertion in `compose-smoke`**, so a missing `public/` cannot stay green again.

## Two things worth the reviewer's attention

**The assertion the ticket asked for would not have worked.** #34 specifies `GET /manifest.webmanifest = 200`. That file is not in `public/` — it is generated by `src/app/manifest.ts`, a metadata route served by Next, so it answers 200 with `public/` missing entirely. The step asserts `/icons/icon-192.png` and `/offline.html`, which do live in `public/`, and checks the manifest as well, with a comment saying why it cannot stand in for the others.

**The layer reordering forces `--ignore-scripts`, and that is not a preference.** This package's own `postinstall` is `next typegen`, which exits 1 with *"Couldn't find any `pages` or `app` directory"* when the source has not been copied yet — I ran it rather than reasoning about it. Nothing is lost, because `next build` generates the same route types.

**`npm rebuild` is not redundant and should not be tidied away.** `--ignore-scripts` does not mean "these scripts were unnecessary", it means "not yet": six packages in this lockfile declare install scripts, and `npm rebuild` runs them once the tree is complete. esbuild is one of the six, and esbuild is what bundles the worker and the database scripts. If its platform binary were ever wrong, the bundling steps immediately below would fail the build rather than ship an image that cannot run — which is the outcome worth keeping, and it depends on those scripts having been run.

## Untouched on purpose

The `--once` flag in the smoke and the comment explaining it, and the compose service staying argument-free. That distinction is load-bearing now that the worker exists: the service runs the loop, the smoke runs one pass and exits. The compose file now says so in a comment, so nobody "tidies" `--once` into it and turns a self-hoster's continuous ingestion into a single pass at boot.

## Verification

Docker cannot run on this machine, so `compose-smoke` is the instrument, as it has been for every image change on this project. Conclusion lines to follow.

**The gate is known to be able to fail, not assumed to be.** Two deliberate reds were run, each on an image built without `public/` and without the build-time check, so the smoke step was the only thing that could catch it.

- **Run 35166339703** reddened the first version of the step. It exited on the first 404, so `/manifest.webmanifest` was never requested and the run could not show the comparison below. The step was rewritten because of that, to check every path and decide afterwards.
- **Run 35166618377** reddened the rewritten gate — the one that will actually guard this image. A changed gate is an untested gate until it has been seen to fail, and this is that run. Its only failing step was *Confirm static assets and the manifest are served*; the other job passed, the image built and the stack came up.

```
GET /offline.html -> 404
GET /manifest.webmanifest -> 200
expected 200 for: /icons/icon-192.png /offline.html
```

The 200 beside the two 404s is now measured rather than reasoned: it distinguishes an image that lost `public/` from a server that is not serving, and it is the direct demonstration that `GET /manifest.webmanifest = 200` — the assertion #34 specified — answers either way and could never have caught this.

Both control branches are closed and deleted, local and remote.


zarghammir:
REVIEW VERDICT: APPROVE

Reviewed at **`104c627`**. Three files, carries current `main`, `MERGEABLE`/`CLEAN`, counts agree both ways, both CI jobs green with every real step executed.

Small, and every claim in it is checkable — so I checked them rather than reading them. All of them hold.

## The acceptance criterion could not have failed, and the control proves it rather than argues it

`#34` specified `GET /manifest.webmanifest = 200`. That file is generated by `src/app/manifest.ts`, a metadata route, so it answers **200 with `public/` missing entirely**. Asserting it would have been an assertion that cannot fail — the family this project has spent the day clearing.

I read the control run's **printed lines**, not its exit code:

```
GET /icons/icon-192.png  -> 404
GET /offline.html        -> 404
GET /manifest.webmanifest -> 200
expected 200 for: /icons/icon-192.png /offline.html
```

That is the whole argument in four lines: the specified assertion is green in exactly the state the gate is meant to catch. And the red is attributable — I checked what the control branch actually changed, and it is **one file, one line**: the `COPY --from=builder /app/public ./public` removed, with a THROWAWAY marker. `compose-smoke` failed at **only** *"Confirm static assets and the manifest are served"*, with the image built, the stack up, web answering 200 and the catalogue seeded before it.

**Rewriting the step after the first control was the right call and the PR says so itself** — the first version exited on the first 404, so the manifest was never requested and the contrast could not be shown. A control that produces the right colour for an unexamined reason is not evidence, and reporting that rather than banking the red is the harder thing to do.

## The number I nearly filed a finding against

The Dockerfile says *"six packages declare install scripts"*. I checked it three ways:

| instrument | answer |
|---|---|
| my own `node_modules` walker | **2** |
| `npm query ":attr(scripts,[install])…"` | **2** |
| `package-lock.json`, `hasInstallScript` | **6** |

The lockfile is right, and the PR body is more precise than my first two instruments: it says *"six packages **in this lockfile**"*. The six are the root package, three separate `esbuild` entries (`esbuild`, `tsx/node_modules/esbuild`, `@esbuild-kit/core-utils/node_modules/esbuild`), `fsevents` and `unrs-resolver`. `npm query` collapses duplicate names and reflects one platform's installed tree; `npm ci` reads the lockfile, which is the unit that matters for a Docker build on a different platform than mine.

**That is the third time today my instrument was weaker than the claim it was testing**, and the only reason it did not become a finding is that "2 versus 6" is a disagreement to resolve rather than a result to report.

The two supporting claims hold too: this package's `postinstall` really is `next typegen`, and the body says *"I ran it rather than reasoning about it"* — the honest form.

## The rebuild is evidenced in the container, not by a tick

`--ignore-scripts` is forced by the layer reordering, and `npm rebuild` puts the skipped scripts back once the tree is complete. What makes that more than a story is what the green run shows **inside** the image:

```
runtime image has every path its entrypoints reference
GET /icons/icon-192.png -> 200   GET /offline.html -> 200   GET /manifest.webmanifest -> 200
total: 18 sources, 496 fetched, 496 new, 484 stories, 1 failed in 9.7s
RESULT: the worker ran a pass.
```

The worker bundle is produced by esbuild. A real pass inside the container is therefore direct evidence that esbuild's install scripts ran — the thing `npm rebuild` exists to guarantee — rather than an inference from the build having succeeded.

The new build-time assertion (`test -f /app/public/icons/icon-192.png`) is a second, earlier line of defence on the same defect, and it fires two minutes before the smoke would.

## Checked and cleared

- **`public/` coverage.** Seven files arrive by a single `COPY`, and the gate spans both levels of it — `offline.html` at the root and `icons/icon-192.png` nested. `sw.js` is unchecked, which I considered and am not filing: there is no partial-copy failure mode for one `COPY`, and the two checked paths already prove the tree arrived.
- **`.npmrc` exists**, so the narrowed `COPY` cannot fail — and the green image build is the proof rather than my `ls`.
- **`restart: unless-stopped`** is the right change now the worker loops, and the comment owns its one awkward consequence: a placeholder `INTERNAL_API_SECRET` produces a restart loop rather than a clean stop. I agree with the trade — a refusal repeated in the log is harder to miss than one printed once — and it is documented where someone will read it.
- **The compose service staying argument-free** while only the smoke passes `--once` now has a comment explaining why, and that distinction became load-bearing the moment the worker gained a loop. Adding `--once` there would silently turn a self-hoster's continuous ingestion into one pass at boot.
- **Control branches.** #44's are gone. Two `control/*` refs remain on the remote and I checked both: `control/no-browser-is-red` and `control/no-server-is-red` — the live, authorised #47 controls, not leftovers from this work.

## Verified myself

Base, merge state, file counts both ways, every CI step at this SHA; the control run's failing step, its printed lines and the single-line mutation that produced it; the six-package claim against the lockfile after two weaker instruments disagreed; the `postinstall` value; the container's own worker pass; `public/` contents against what the gate asserts.

## Taken on trust

- **Docker itself.** I cannot run it here, so `compose-smoke` is the instrument for every image claim above — which is what the PR body says, and it is the same instrument that caught the original defect when nothing else did.
- **Layer-cache behaviour over time.** The reordering is correct by construction, but I did not measure a second build to see the `npm ci` layer reused.
- **Alpine's install-script set.** The lockfile's six is platform-independent; I did not enumerate what actually runs on `node:22-alpine`.

PR #45, opened 2026-09-17
[CONTROL — DO NOT MERGE] prove the static-asset assertion can fail

zarghammir:
**This branch exists to be RED and must never be merged.**

Throwaway control for #44. The image is built without `public/`, and the build-time `test -f /app/public/...` assertion is removed too, so the only thing left that can catch it is the new `compose-smoke` step.

Expected: **compose-smoke RED** at *Confirm static assets and the manifest are served*, on `/icons/icon-192.png`.

Expected alongside it, and the reason the ticket's suggested assertion was not enough on its own: **`/manifest.webmanifest` would still answer 200 here**, because it is generated by `src/app/manifest.ts` rather than served from `public/`.

Closed and deleted as soon as the run id is recorded.


zarghammir:
Control complete. Run **35166339703**, job 105028164779.

`compose-smoke` RED at *Confirm static assets and the manifest are served*:
```
expected 200 for /icons/icon-192.png
```
The other job passed, the image built (233 MB), the stack came up — so the red is the assertion catching a missing `public/`, which is what it exists to do.

**One thing this run does NOT show**, stated rather than implied: the manifest was never requested. The loop exits on the first failure and `/manifest.webmanifest` is checked last, so this is not evidence that the manifest still answers 200 with `public/` missing. That claim rests on `src/app/manifest.ts` being a metadata route, not on this run. #44 is being changed to check all three and report them together, which would have made the contrast visible here.

Closing; branch deleted.

PR #46, opened 2026-09-17
[CONTROL — DO NOT MERGE] prove the rewritten asset gate can fail

zarghammir:

Throwaway control for #44, and distinct from the one already run at #45. That one proved the *previous* version of the assertion could fail. The step was then rewritten (104c627) to check every path and decide afterwards instead of exiting on the first 404 — so the gate that will actually guard the image has not yet been seen to go red. A gate that has been changed is a gate that has not been tested.

The image is built without `public/`, and the build-time `test -f /app/public/...` assertion is removed with it, so the `compose-smoke` step is the only thing that can catch it.

Expected: **compose-smoke RED** at *Confirm static assets and the manifest are served*, printing all three results rather than stopping at the first:

```
GET /offline.html -> 404
GET /manifest.webmanifest -> 200
expected 200 for: /icons/icon-192.png /offline.html
```

The 200 beside the two 404s is the secondary point: `/manifest.webmanifest` is generated by `src/app/manifest.ts`, a metadata route, so it answers whether or not `public/` is in the image. That is why the assertion #34 originally specified could not have caught this, and why an image missing `public/` is distinguishable from a server that is not serving.


zarghammir:
Control complete. Run **35166618377**.

`compose-smoke` RED, and the **only** failing step is *Confirm static assets and the manifest are served*. The other job passed, the image built and the stack came up, so nothing incidental produced this.

```
GET /offline.html -> 404
GET /manifest.webmanifest -> 200
expected 200 for: /icons/icon-192.png /offline.html
```

Both things this run existed to establish:

1. **The rewritten gate can fail.** #45 proved the previous version could; the step was then changed to check every path and decide afterwards, and a changed gate is an untested one until it has been seen to go red.
2. **The contrast is measured, not reasoned.** Two 404s beside a 200 is what distinguishes an image missing `public/` from a server that is not serving — and it is why `GET /manifest.webmanifest = 200`, the assertion issue #34 originally specified, could never have caught this: that path is generated by `src/app/manifest.ts` and answers either way.

PR #47, opened 2026-09-17
CI: run the browser checks, and treat "could not run" as a failure

### LINKED ISSUE #17: End-to-end harness: Playwright, axe, screenshots, CI job

zarghammir:
Part of #17 — see **Scope** below, because this does not close it.

## In plain words

**What is happening in the records.** Nothing. This adds an automatic check that opens the app in a real browser on every proposed change.

**The flow.** Until now, two scripts in this project could open the app in a browser and check things no other test can see — whether anything overflows the screen, whether the bottom tab bar is right on a phone and absent on a laptop, whether the theme switcher works by keyboard, whether the offline page is stored correctly. They only ran when somebody remembered to run them. From now on they run by themselves, every time.

**Who.** Nobody clicks it. It runs on every proposed change before a human looks.

**What they see.** A green or red tick, and if it is red, a log saying which of the three things went wrong.

**What it is for.** A check that runs only when someone remembers is absent exactly when nobody is looking — which is when problems arrive.

**Where it stands.** Waiting on CI, plus one deliberate failure to prove the new check can actually fail. Not a click yet.

## The part that matters most

These scripts use **three** exit codes, and the middle one is the point:

| exit | meaning | this job |
|---|---|---|
| 0 | the checks passed | green |
| 1 | the checks **ran and failed** | red, saying so |
| 2 | the instrument **could not run** — no browser, or nothing serving | **red**, saying so |

A job that treats 2 as a skip produces a green tick for a run that measured nothing. Both non-zero cases are red here, separated only so the log says which happened, because the fixes are unrelated. All four branches of that logic were driven by hand before it was committed.

## Cost, stated because we are counting

**118 seconds**, measured on this branch's own run rather than estimated — containers 21s, install 17s, build 21s, the sweep 34s, the shell checks 8s, everything else under 3s. My original estimate here was 3–4 minutes; the real figure is a little under two, and the next person weighing every-PR against a path filter should weigh it against that.

It runs on **every** pull request rather than behind a path filter: a filter would have to guess which files can change a rendered page, and the honest answer is most of them — a utility, a token, a layout, a dependency bump. A browser check that runs only sometimes is the problem this job exists to fix. `needs: ci` keeps it from spending anything on a branch that already fails lint, types, tests or the build.

No browser is downloaded. `playwright-core` ships none by design, and the runner image carries Google Chrome, which is what `scripts/lib/browser.mjs` resolves through Playwright's `chrome` channel. The version is printed rather than assumed — and if it ever is not there, the scripts exit 2 and this job goes red saying the instrument could not run, which is the correct outcome rather than a silent skip.

## Scope — this does not close #17

Delivered here: the CI job, the three-state handling, and the control below.

**Not delivered, and someone should decide where it goes:**
- The Playwright config with named projects, the `npm run screenshots` script and the `docs/screenshots/<area>/` naming scheme the ticket asks for. Much of the ticket's *intent* is already met by the two existing scripts, which sweep six routes at two widths in two themes — but the screenshot half is genuinely absent.
- The story-count floor ("at least N stories rendered"). This job seeds the catalogue but does not ingest, so pages render their chrome without live stories. Seeding is deterministic; ingesting would put a network fetch of eighteen third-party feeds inside a browser job, and the checks wired here measure the shell rather than the content. A story-count floor would change that trade and should be decided deliberately.
- `npm run routes:check` from #42, the third unwired check. It does not exist on main yet, and wiring a script that is not there would fail the job. It is one step to add once #42 merges.

## Verification

**Green, and the sweep reports what it measured.** Run `35166985867`: all three jobs success, no failing steps, and the script's own output reads `"statesAudited": 24` — six routes, two widths, two themes. That number comes from `report.states` in `scripts/a11y-audit.mjs`, incremented as each state is visited; this job prints no count of its own, so there is nothing here that could restate an expected figure. The script also floors itself (a11y-audit.mjs:191-214): a sweep that visited the wrong number of states fails regardless of what axe found.

**Both exit-2 causes are proved to go red, separately.** The job's message names two causes — no browser, or nothing serving — and a message naming two causes is a claim that both are reachable. Each was demonstrated on its own throwaway branch, both now closed and deleted:

| control | run | arrangement | the script's own line |
|---|---|---|---|
| no browser | `35167284095` | Chrome deleted, **server up** (`server answered 200 after 2s`) | `No browser to drive.` |
| nothing serving | `35167411488` | Chrome present, **no server started** | `Nothing is serving http://127.0.0.1:3210.` |

Both failed at *Accessibility, overflow and bottom-bar sweep* and nowhere else; the other two jobs passed in both runs. So exit 2 fails the job rather than passing or skipping, **and the two causes are distinguishable in the log** — the script names the specific one, with remediation, directly above the job's line that names both because an exit code alone cannot tell them apart.

The second control exists for the future rather than the present: someone editing the server-start step could turn "nothing serving" into a hang or a different error, and without this the only thing asserting otherwise would be a sentence in this body.

The scripts themselves are untouched — they belong to the UI lane.


zarghammir:

Reviewed at **`b0e2464`**. One file, carries current `main`, `MERGEABLE`/`CLEAN`, counts agree both ways, all three jobs green.

**This job is the right thing and the three-state contract holds.** One finding, one line, and I would approve on sight. Everything else I attacked came back clean — including two things I nearly filed and one number that turns out better than the PR claims.

## The finding — `VERIFY_URL` binds one script, and the comment says two

```yaml
# Both scripts default to this and take it as argv[2]; naming it once here
# keeps the server and the scripts from drifting apart.
VERIFY_URL: http://127.0.0.1:3210
```

Only one script reads it:

| script | how it resolves its base URL |
|---|---|
| `scripts/verify-shell.mjs:20` | `process.argv[2] \|\| process.env.VERIFY_URL \|\| "http://127.0.0.1:3210"` |
| `scripts/a11y-audit.mjs:21` | `process.argv[2] \|\| "http://127.0.0.1:3210"` |

Neither step passes an argument, so both currently fall through to their own literal default — which is why everything is green and why this is latent rather than broken.

**The trap is the comment's own stated purpose.** It says naming the URL once *"keeps the server and the scripts from drifting apart"*. Change the port here — a conflict on 3210, a second server, a matrix — and `verify:shell` follows it while the sweep keeps driving `127.0.0.1:3210`. The two scripts would then be measuring different servers, or the sweep would fail to connect and exit 2 while the variable that was supposed to prevent exactly that sat right above it. A knob that binds half of what it claims is worse than no knob, because the comment tells the next person they do not need to check.

One line either way: add `|| process.env.VERIFY_URL` to `a11y-audit.mjs:21`, or narrow the comment to the one script it governs. I would take the first — the server port is set in this job, and having the scripts read it is the property the comment is reaching for.

## The three-state contract holds, and I tested the mechanism rather than the shape

The contract depends entirely on `npm run` handing the script's exit code back untouched. If npm normalised non-zero to 1, exit 2 would arrive as "RAN and FAILED", the job would still be red, and **the diagnosis would be silently wrong while the logic still looked correct**. So I measured it in this repo:

| | result |
|---|---|
| script exits 2, via `npm run` | **2** |
| script exits 7, via `npm run` | **7** |
| control: bare `node -e "process.exit(2)"` | **2** |

And every path through the `case` is terminal: `0` falls through and the step succeeds; `2` prints the could-not-run error and `exit 1`; **`*)` catches everything else** and re-raises the code. There is no branch that reaches the end of the step without either succeeding deliberately or exiting non-zero — so a 2 cannot become a skip or a green.

`browser.mjs` really does have two `exit(2)` sites — *"No browser to drive."* and *"Nothing is serving …"* with remediation — which is what makes the job's both-causes wording honest rather than vague.

## Both controls, read as logs rather than as exit codes — and the isolation holds

I asked for the second control and it earned its run. The two causes **are** distinguishable, and the distinguishing line comes from the script, directly above the job's summary:

**`control/no-browser-is-red`** — Chrome present (`Google Chrome 152.0.7977.82`), then deliberately removed, and critically **`server answered 200 after 2s`**. So its exit 2 could only be the browser. Script printed `No browser to drive.`

**`control/no-server-is-red`** — Chrome present and *not* removed, no server started. So its exit 2 could only be the server. Script printed `Nothing is serving http://127.0.0.1:3210.`

Each control eliminates the other's cause from its own log. That is what makes them two controls rather than two ways of breaking one thing, and it is the distinction that would have made a single control worthless.

**One thing neither control reaches, non-blocking:** both failed at *"Accessibility, overflow and bottom-bar sweep"*, so the job stopped and *"The four checks axe cannot see"* never ran. Its `case` block is — I checked, after my first comparison said otherwise and was wrong — structurally identical to the first. So the contract is proven once and copied once. Not a defect; but two copies of a guard where one control reaches one of them is how divergence starts. A shell function or a composite action would leave one copy and let one control cover both.

## `statesAudited: 24` is measured, and I checked the sharper question

Not whether the counter exists, but where it sits relative to the work. `report.states++` is at `a11y-audit.mjs:75`, **after** `await page.goto(base + route, { waitUntil: "networkidle" })` and `await page.evaluate(() => document.fonts.ready)`. A counter incremented at the top of a loop that then visits nothing is still a counter; this one only advances once a navigation has completed and the page is ready. Initialised `0` at `:48`, printed from `report.states` at `:219` — there is no constant for the job to restate, and the job prints no count of its own.

The green run's own JSON: `statesAudited: 24`, `barsMeasured: 12`, `floorPassed: true`, `seriousOrCritical: 0`, `horizontalOverflow: 0`. And the script's floors (`:191-214`) are independent of that counter, so the two would disagree if either drifted — I proved those floors bite on this same script during #31, by renaming the bar's landmark so the selector missed: it reported zero clipped labels and zero overflow, which without the floor reads as a clean pass, and exited 1 naming the real cause.

## The cost is better than the PR claims, measured rather than reasoned

The body says roughly 3-4 runner minutes. Measured from this run's own job timings, **the browser-checks job takes 118 seconds — 2.0 minutes**:

```
Initialize containers 21s · Install dependencies 17s · Build 21s
Accessibility sweep 34s · The four checks axe cannot see 8s · everything else ≤3s
```

You flagged the cost case as a decision signed off on reasoning rather than measurement. It is now measured, and it lands in the PR's favour — roughly half the estimate. I would put the real number in the body rather than the estimate, because the next person to weigh this trade should weigh it against 2 minutes.

Running on every pull request rather than a path filter is the right call for the reason the comment gives: a path filter has to guess which files can change a rendered page, and a browser check that runs only sometimes is the problem this job exists to fix. `needs: ci` is the right economy — a branch already failing lint or the build does not pay for a browser.

## Verified myself

Base, merge state, file counts, all three jobs and their steps; `npm run` exit-code propagation with a control; every branch of both `case` blocks; the two `exit(2)` sites; both control runs' failing step, printed lines and isolation; the counter's position relative to the work; the green run's JSON; and the job's real duration from the timings API.

## Taken on trust

- **The runner image.** Chrome 152 is printed by the job itself, so its presence is evidenced — but I did not verify the image manifest, and "no browser download" rests on `playwright-core` shipping none, which I confirmed from `package.json` rather than by watching an install.
- **npm's behaviour at the pinned version.** I measured propagation with npm 11.11.0 locally; CI pins 11.19.1 via `packageManager`. Exit-code propagation is not version-sensitive behaviour, but I measured one and not the other.
- **The scoped-out items** — the screenshot harness (#48), the story-count floor, and the `routes:check` step waiting on #42 — I read in the body and did not independently confirm each is genuinely blocked rather than deferred.

One line and I will approve in a new comment.

zarghammir:
REVIEW VERDICT: APPROVE

Verified at **`f21d7c2`**. The finding is fixed by a third option better than either I offered, and the non-blocking note was taken while the file was open.

## The fix I did not think of

I offered two: wire `VERIFY_URL` into `a11y-audit.mjs`, or narrow the comment. The lane read both scripts and found the one input **both already honour** — `argv[2]` — and passed the URL as an argument:

```yaml
run: bash .github/scripts/run-browser-check.sh "The accessibility sweep" a11y "$VERIFY_URL"
run: bash .github/scripts/run-browser-check.sh "The shell verification" verify:shell "$VERIFY_URL"
```

One value genuinely reaches both scripts, `VERIFY_URL` stays the single place it is named, and no half-bound knob needs a comment explaining which half it binds. The comment that remains is now *accurate* about the asymmetry rather than papering over it — *"verify-shell.mjs reads argv[2] || VERIFY_URL, a11y-audit.mjs reads argv[2] only"* — which is exactly the right thing for the next person to know.

It is also the option that respects the constraint I did not know about: it required no edit to UI's script, which my first suggestion would have.

## The extraction — I drove all four states myself, because the controls are gone

A refactor of the thing that reports the three states is where a state can quietly stop being reachable, and both controls that would have caught it are deleted branches. So I ran the extracted `.github/scripts/run-browser-check.sh` against probe scripts exiting 0, 1, 2 and 7:

| script exits | wrapper exits | message |
|---|---|---|
| 0 | **0** | `RESULT: … ran against http://example.test:9999 and passed.` |
| 1 | **1** | `… RAN and FAILED (exit 1).` |
| **2** | **1** | `… could NOT RUN — no browser could be resolved, or nothing was serving http://example.test:9999. This is not a pass and not a skip; see the script's own message above for which.` |
| 7 | **7** | `… RAN and FAILED (exit 7).` |

All four preserved. The `2` case still maps to a red that says *could not run* rather than *ran and failed*, the unexpected code is still re-raised rather than flattened, and the URL now appears in the message — which it could not before, because the wrapper did not know it.

I also re-measured the layer underneath, in this repo rather than taking it from my own earlier verdict: `npm run <script> -- <url>` returns **2** for a script exiting 2, and `argv[2]` receives the URL. The `--` passthrough does not disturb either.

Extracting it was the right call for the reason the file now states: *"Two copies of a contract that a control can only ever reach one of is how the copies drift apart."* Both controls failed at the first step, so the second copy had never been executed by anything.

## Checked and cleared

- **The file mode.** I had to `chmod +x` it to run my own probes, which made me check: the workflow invokes it as `bash .github/scripts/run-browser-check.sh`, so the committed `100644` is correct and no exec bit is needed. Reverted my change; the tree is clean.
- **`statesAudited`**, the floors, the two `exit(2)` sites and both control runs' isolation are unchanged from my verdict and I am not re-litigating them.
- **The measured 118 seconds is in the body with the breakdown**, and keeping the original 3-4 minute estimate in the record beside it is better than replacing it — the gap between what was reasoned and what was measured is the useful part.

## The instrument note, because it happened three times inside one verification

My first attempt at driving the four states reported that exit 2 produced *"RAN and FAILED"* — which would have been a serious finding against the extraction. It was not true. My probe scripts were inline `node -e` with nested quotes that the shell mangled, so the exit-2 probe died on a syntax error and exited 1. The echoed command in the log showed the broken quoting; that is what gave it away.

Then my rewritten loop reported `MISMATCH` on all four rows, because `set -- $probe` does not word-split under zsh, so every row compared against an empty expectation.

Only the third attempt — explicit calls, probes in real files — measured anything. **Three broken instruments, one after another, against code that was correct the whole time.** The verdict it would have produced was available at every step and wrong at every step. What made the difference was not care in reading the result but reading the *content* of it: a message that did not match the exit code, and an expectation column that was blank.

That is the fifth time today my instrument was weaker than the claim it was testing, and it is the thing I would most want the next reviewer of this repository to take from the day.

## Verified myself

The four exit states through the extracted script; `npm run -- <url>` propagation and `argv[2]` delivery; both scripts' URL resolution at this head; both call sites passing it; the invocation form against the committed file mode.

## Taken on trust

- **CI at this SHA**, which I waited for rather than assumed — all three jobs green, and the browser-checks job is the one that exercises this wrapper for real against a live server and a real browser, which my probes cannot.
- **npm's behaviour at the pinned version** — I measured with npm 11.11.0 locally; CI pins 11.19.1.
- The runner image's Chrome, unchanged from my verdict.

PR #49, opened 2026-09-17
[CONTROL — DO NOT MERGE] prove exit 2 fails the browser job

zarghammir:

Throwaway control for #47. The runner's Google Chrome is deleted before the browser checks run, so `scripts/lib/browser.mjs` cannot resolve a browser and exits **2** — *the instrument could not run*.

Expected: **browser checks RED** at *Accessibility, overflow and bottom-bar sweep*, with

```
::error::The accessibility sweep could NOT RUN — no browser could be resolved, or nothing was serving. This is not a pass and not a skip.
```

and the job failing rather than passing or skipping.

The server is still started, so the exit 2 comes from the no-browser path rather than the nothing-serving path. Both produce the same code and the job's message names both causes; this control exercises the first.

The point: exit 2 is the state a CI job is most likely to get wrong, because it is non-zero for a reason that looks like an environment problem rather than a defect. A job that treats it as a skip reports a green tick for a run that measured nothing. This proves it does not.


zarghammir:
Control complete. Run **35167284095**.

`browser checks` RED, only failing step *Accessibility, overflow and bottom-bar sweep*. The other two jobs passed.

```
chrome is gone, as intended
server answered 200 after 2s
No browser to drive.
::error::The accessibility sweep could NOT RUN — no browser could be resolved, or nothing was serving.
```

The server answered 200, so this isolates the **no browser** cause rather than the nothing-serving one, which is what it was built to do. Exit 2 fails the job rather than passing or skipping.

PR #50, opened 2026-09-17
[CONTROL — DO NOT MERGE] prove the nothing-serving exit 2 goes red

zarghammir:

The second of the two exit-2 controls for #47, asked for by the reviewer. #49 proved the **no browser** cause. This proves the other one the job's message names: **nothing serving**.

The server is deliberately not started; Chrome is left in place. So `scripts/lib/browser.mjs` exits 2 from `requireServer` rather than from `launchBrowser`.

Expected: **browser checks RED** at *Accessibility, overflow and bottom-bar sweep*, with the script naming the specific cause —

```
Nothing is serving http://127.0.0.1:3210.
```

— followed by the job's own line, which names both possible causes because it cannot tell them apart from an exit code alone:

```
::error::The accessibility sweep could NOT RUN — no browser could be resolved, or nothing was serving.
```

**Why this control and not a sentence.** A message that names two causes is a claim that both are reachable and distinguishable. Half of that was proved. The other half mattered more for the future than the present: someone editing the server-start step could turn "nothing serving" into a hang, a different error, or no exit 2 at all, and the only thing asserting otherwise would be prose.

If the two causes turn out to be indistinguishable in the log, that is a finding on #47 rather than a failed control.


zarghammir:
Control complete. Run **35167411488**.

`browser checks` RED, only failing step *Accessibility, overflow and bottom-bar sweep*. The other two jobs passed.

```
no server on 127.0.0.1:3210, as intended
Nothing is serving http://127.0.0.1:3210.
  This script measures a PRODUCTION build; start one first:
    npx next build && npx next start -p 3210
  then re-run with the same URL.
  fetch failed
::error::The accessibility sweep could NOT RUN — no browser could be resolved, or nothing was serving.
```

**The two causes are distinguishable, which is what this control existed to establish.** #49 printed `No browser to drive.` with the server up; this printed `Nothing is serving …` with Chrome present. The job's own line names both because an exit code alone cannot tell them apart — the script supplies the specific cause directly above it, including how to fix this one.

Issue #51, opened 2026-09-17
The empty Today state cannot say when the next run is — and the honest answer is the last one


zarghammir:
**Raised by the UI lane while building #13, which wrote the empty state WITHOUT inventing a time rather than hard-coding "every 30 minutes".** That was the right call and this ticket exists so the gap has an owner instead of a guess.

Issue #13 asks the empty state to show "the time of the next run". Nothing in the contract exposes it: `/api/brief` returns the window, not the worker's schedule, and `INGEST_INTERVAL_MINUTES` is server-side config the browser never sees.

**The complication that makes "next run" the wrong field to add:** there are two different schedulers, and which one is authoritative depends on how the app is deployed.

- Self-hosted via compose, the worker loops on `INGEST_INTERVAL_MINUTES`.
- Hosted, the GitHub Actions cron drives it (every 30 minutes today, and that cadence is under review because it consumes ~72% of the free runner allowance — see the cost discussion on #37).

So a "next run at HH:MM" rendered from server config would be **confidently wrong on the hosted deployment**, which is the one the owner will actually look at.

**Proposed instead: show when the last run finished.** `ingest_runs` already records it, it is true in both deployments, it needs no knowledge of any scheduler, and it answers the reader's actual question — *is this thing working, or is it stuck?* "Nothing yet · last checked 11 minutes ago" tells them the system is alive and the news is quiet. "Next run at 14:30" tells them a fact that may be false.

If the owner specifically wants a next-run time, it needs the scheduler to publish its own cadence rather than the web app inferring it — a bigger change, and worth its own decision.

**Scope:** add a last-run timestamp to `/api/brief` (or a small status field), contract updated in the same PR per the rule that the PR making a statement false is the PR that corrects it. UI then renders it.

Not blocking #13 — the empty state ships without a time and gains one when this lands.

PR #52, opened 2026-09-17
Today page: the morning brief

### LINKED ISSUE #13: Today page: the morning brief

zarghammir:
Closes #13.

## What it is in plain words

The hero feature now exists. Open the app and you get your morning brief: a ranked list of what happened, each story with what kind of thing it is, how well-sourced it is, who said it first, when we saw it, and how long it takes to read. You can choose how much time you have, save a story for later, share it, or hide it.

This reads from fixtures rather than live data. **That sentence was written when the routes were unmerged; they landed on `main` while this PR was in review**, so fixture mode is now a choice rather than a necessity — kept for this PR because switching the data source mid-review would invalidate everything the reviewer verified, and flipped in a small follow-up. It is a deliberate, explicit mode rather than a fallback, for a reason given below.

## Read this first: my fixtures lied about reading time, and it changed what the screen does

Every fixture story carried **3 to 6 minutes** of reading time. The product does not compute that number. It computes it from the story's short summary or excerpt at 220 words a minute, and these excerpts run about 25 words — so the true value is **1 minute for every one of them**. I had written *article* time.

That is not a cosmetic error. It changed the behaviour on screen: a **ten-minute brief held two stories instead of all six**, which made the reading-length switch look far more aggressive than it will ever be. The first set of screenshots showed exactly that. Anyone reviewing them — including the owner — would have been judging a behaviour this product does not have.

This is the second time on this branch that a fixture claimed something the pipeline cannot produce. The first was a story marked *Unverified* while listing four newsrooms, which the product grades *Corroborated*; that one was the fixture whose whole job was to demonstrate the owner's ruling that unverified items belong in the brief when labelled. Both are now guarded the same way: **a test derives the value from the story's own data** rather than trusting what I typed.

**One of my own tests was passing only because of the bad data**, and this is the part I most want to survive review. It asserted that the ten-minute brief is *strictly* shorter than the whole brief. With honest one-minute stories, all six fit inside ten minutes, so the ten-minute brief **is** the whole brief. The assertion was true only of numbers the pipeline cannot generate. It now asserts the real property — five is shorter than ten, ten is no longer than all — and the test records **why the weaker-looking assertion is the correct one**, because `ten < all` reads like the more thorough check and is exactly what someone would restore.

## What a person can now do

- **Read the brief.** Ranked stories, largest first. Importance is card size, not a badge — the design the owner signed off.
- **Choose how long they have.** 5 minutes, 10 minutes, or everything. The choice lives in the URL, so it survives a reload, can be shared, and the switch works before JavaScript runs. The last choice is remembered.
- **Tell at a glance how well-sourced something is.** A four-bar meter **and** the word, never one without the other. What kind of thing it is sits on the left as a plain word, in a different shape, because they are two different things.
- **Save, share or hide a story.** Save and Hide apply instantly and roll back if the write fails — and the failure is *announced*, not left as a button that silently snapped back. Share uses the system share sheet where it exists and copies the link otherwise.

## Two decisions I would not want quietly simplified

**Fixture mode is an explicit flag, not a fallback.** A client that tries the network and quietly uses fixtures when a route returns 404 cannot tell *"this route does not exist yet"* from *"this save failed"* — it would report a failed write as a success, and the reader would believe their save happened. `NEXT_PUBLIC_USE_FIXTURES=0` switches to the real routes. This looks like something to "simplify into a fallback because it is more robust". It is the opposite.

**The over-budget header currently cannot fire, and that is fine.** The brief always returns at least one story, so a single long story can exceed a five-minute setting; when that happens the header says *"longer than your 5-minute setting, the top story alone runs 6 minutes"* rather than presenting a contradiction as though nothing happened. With honest one-minute stories, no single story can exceed a five-minute budget, so this is **unreachable rather than wrong**. It becomes reachable the day a summariser writes longer bodies. Please do not delete it as dead code, and please do not "fix" it into firing.

## Screenshots

> The repository is private, so GitHub will not render these inline. They are under `docs/screenshots/today/`.

**Where it lives** — the brief inside the app, with the sidebar showing which of the six surfaces you are on.
→ [Today on a laptop, dark](https://github.com/zarghammir/ai-radar/blob/981b2d29a49c1e1bd5328236b1ac5f9e1aa4ff58/docs/screenshots/today/today-laptop-dark.png)

**The main screen** — what you actually open in the morning: newest and most important first, each story saying what kind of thing it is and how well-sourced it is.
→ [Today on a phone, dark](https://github.com/zarghammir/ai-radar/blob/981b2d29a49c1e1bd5328236b1ac5f9e1aa4ff58/docs/screenshots/today/today-phone-dark.png)

**The same thing in daylight** — the palette holds; paper stays light in both themes because you read on the print, not on the bench.
→ [Today on a phone, light](https://github.com/zarghammir/ai-radar/blob/981b2d29a49c1e1bd5328236b1ac5f9e1aa4ff58/docs/screenshots/today/today-phone-light.png)

**Choosing how much time you have** — five minutes gives you five of the six stories, and the count and minutes at the top change with it.
→ [Five-minute mode on a phone](https://github.com/zarghammir/ai-radar/blob/981b2d29a49c1e1bd5328236b1ac5f9e1aa4ff58/docs/screenshots/today/today-phone-5min.png)

## How to verify

```bash
npm ci
npx next build && npx next start -p 3210
npm run a11y            # 24-state sweep: axe, overflow, the bottom bar
npm run verify:today    # the four behaviours this ticket requires
```

**Measured on a production build at `1c9477f1034582b7345542cd0904fe84a22f0de4`:**

| | result |
|---|---|
| 5-minute switch shortens the list | 5 of 6 stories, all 6 at ten minutes |
| Save toggles and survives a reload | yes |
| Hide removes the card and it stays hidden | yes, and the hidden title is genuinely absent from the list, not merely uncounted |
| Stories rendered at 390/1440 × light/dark | 6, 6, 6, 6 |
| Horizontal overflow | 0 in every state |
| axe across 24 states | **0 serious, 0 of any impact** |

**The story-count floor** is why those are worth reading. A Today page rendering zero cards makes every assertion trivially true: the switch "shortened" the list (0 is not more than 0), hiding "removed the card" (there was no card), nothing "overflowed" (nothing was drawn). The check requires at least two stories in every state and fails naming the state that came up short.

**Shared surface:** `PageShell` gained an optional `controls` slot and its `summary` widened from a string to a node. Five other pages use it, so the sweep covered all 24 states rather than just this page — clean. It also turned up a real hazard in my own code: the summary slot wrapped its child behind a truthiness check, and **an element is always truthy**, so a summary rendering null would still draw the wrapper and its margin. Fixed on the container with `empty:hidden`, because no call site can see it.

## Not done, and not claimed

- **The empty state does not show the time of the next ingestion run**, which the ticket asks for. Nothing in the contract exposes it, and there are two schedulers — a self-hosted loop and a cron — so a time rendered from server config would be confidently wrong on the deployment the owner actually uses. Filed as #51; the empty state gains it there.
- **No live data.** Fixture mode until the routes merge.


zarghammir:

Reviewed at **`1c9477f`**. 23 files, carries current `main`, `MERGEABLE`/`CLEAN`, counts agree both ways, both CI jobs green. My own run: suite green, and I re-ran the browser evidence myself rather than reading it.

One finding. It is one line, and I would approve on sight — but it is the sharpest shape I have seen on this project, so I want to be exact about why.

---

## The over-budget test asserts nothing, and the PR body already says why

`src/lib/api/brief-summary.test.ts:62-67`

```ts
it("is only ever reachable with a single story, which is why the copy can say so", () => {
  for (const length of ["5", "10"] as const) {
    const b = fixtureBrief(length);
    if (b.readingMinutes > Number(length)) expect(b.count).toBe(1);
  }
});
```

**Proved rather than reasoned:** changing `toBe(1)` to `toBe(999)` leaves the file **fully green — 10 passed**. The assertion never executes, because with honest one-minute fixtures `b.readingMinutes > Number(length)` is never true.

What makes this different from an ordinary vacuous test is that **its vacuity is already documented in prose, one document away.** The PR body says:

> *"With honest one-minute stories, no single story can exceed a five-minute budget, so this is **unreachable rather than wrong**."*

That sentence and this test cannot both be load-bearing. If the state is unreachable with these fixtures, then a test gated on reaching it asserts nothing — and the test's name is a coverage claim about exactly the thing the body is reasoning about: *"which is why the copy can say so."* The copy's truthfulness is left unpinned by the test that says it pins it.

**The property is true, and that is what makes this worth fixing rather than deleting.** `docs/api.md:373-375` says the brief returns stories *"until the cumulative reading time would exceed the target, always returning at least one story"* — so only the first story can push past the budget, an over-budget brief always holds exactly one, and "the top story alone" is accurate. The comment above the test derives it correctly. Only the assertion cannot see it.

Two ways out, and the file already contains the pattern for the first: the neighbouring `overBudget` tests construct the state directly (`briefOf([{ readingMinutes: 6 }], "5")`) instead of hoping a fixture produces it. Doing the same here — assert that a brief whose first story exceeds the budget holds exactly one story — pins the property against the rule rather than against the data. Alternatively, floor the loop so it fails when it finds no over-budget case, which would make the test honest by going red today.

**One instance, not a pattern.** I swept before filing: it is the only conditionally-guarded assertion in any of the new test files.

---

## Not a finding, but the derivation test deserves a floor

Both negative assertions in `fixtures.test.ts` — `:140` and `:174` — are individually vacuous on an empty fixture set, and I checked whether anything catches that rather than assuming:

| control | result |
|---|---|
| corrupt one fixture's `readingMinutes` | derivation test **red**, naming the slug: `openai-opens-codex-agent-harness-api: fixture 6, computed 1` |
| export `FIXTURE_STORIES` as `[]` | **7 other tests in the file go red** — but both `toEqual([])` tests stay **green** |

So the protection is real and would fire; the two tests are vacuous but not unprotected. That is the difference between a defect and a note, and it is why this is a note. A floor inside them — `expect(FIXTURE_STORIES.length).toBeGreaterThan(0)` — would make each honest on its own, which matters most for the derivation test, since it is the one guarding the defect that actually bit.

## The fixture correction itself is right, and the sibling hunt came back clean

The original defect is well described: `readingMinutes()` computes from the excerpt at 220 wpm, the fixtures carried article time, and a ten-minute brief held two stories instead of six. The fix derives each fixture's minutes from its own body text and **names the offending slug on failure** rather than reporting a count.

What I went looking for was what else had been calibrated against the wrong numbers while they were wrong. The one assertion the lane found and corrected is recorded in the code itself — *"An earlier version asserted ten < all and passed only because the fixtures overstated reading time"* — and I found no others: every remaining numeric assertion in the new tests is a structural floor (`> 0`, `toBe(1)` on a deliberately-quiet brief, label-table sizes) rather than a duration or a count tuned to the old values.

## Verified myself, on the machine

#52's CI has **two** jobs, not three — the browser checks job does not exist on this branch because #47 has not merged. So the 24-state sweep in the body is the author's own local run, which is the claim least worth taking on trust on a surface six pages share. I ran it:

```
statesAudited 24 · serious 0 · violations 0 · overflow 0
floorPassed true · bars 12/12 · tabs 6 · narrowest 65px · clipped 0
verify:shell exit 0, floor passed · verify:today exit 0, floor passed, minStories 2
```

**And I re-ran the floors rather than carrying the proof over from #31.** Renaming the bottom bar's landmark so the selector misses: exit 1, *"bar found in 0 of 12 phone states — that is the selector or the render, not the layout"*, while `anyLabelClipped` and `anyNavOverflow` both read **0** — which without the floor reads as a clean pass. `statesAudited` stayed **24**, and that is the part that makes it a diagnosis: the sweep visited everything and found no bar, so the failure is the selector rather than a run that stopped early. Two numbers disagreeing is what tells them apart.

## The shared surface

`PageShell`'s `summary` widening to `ReactNode` and gaining a `controls` slot is safe for the five other pages — they pass strings, which are valid nodes — and the `<p>` → `<div>` change is required rather than incidental, since a `<p>` cannot legally contain the structure a node summary allows.

`empty:hidden` is on the **container**, with the reasoning written beside it: *"An element is always truthy, so a `summary` that renders null still draws this div and its margin; the ternary cannot see that."* That is the correct fix and the correct place — a check at each call site cannot see what its own child renders.

## Also checked and cleared

- **"The top story alone" is accurate**, and accurate because of a rule the code cites rather than an assumption — see the contract quote above.
- **`overBudget` itself is genuinely covered**, by tests that construct the state directly, including the boundary: 5 minutes under a 5-minute setting is **not** over. So the logic is exercised even though fixtures cannot reach it, and *"unreachable rather than wrong"* is correctly scoped in the body, with the condition under which it becomes reachable named.
- **Keeping it rather than deleting it is right.** A correct-but-currently-unreachable branch with both failure modes named — do not delete as dead code, do not "fix" into firing — is better than either alternative.

## Taken on trust

- **The screenshots.** Regenerated after the fixture correction; I measured the live pages instead of diffing images.
- **The browser job's verdict on this branch**, which does not exist yet — my sweep is a local run on my own machine, with the same standing as the lane's, differing only in being a second independent one.
- **Real API behaviour.** Everything here runs against fixtures; the contract's packing rule is quoted and reasoned about, not exercised against a server.

One line and I will approve in a new comment.

zarghammir:
Fixed at `981b2d29a49c1e1bd5328236b1ac5f9e1aa4ff58`. You were right, and the part that stings is the part you pointed at: **the information was already written down.** My own PR body says the over-budget state is unreachable with these fixtures, one document away from a test that depended on it being reachable. Those two statements could not both be load-bearing, and the silent one was the test.

**What changed.** The property is real and worth testing — only the first story can push past the budget, because every later one that would exceed it is skipped, so an over-budget selection always holds exactly one story. That is what lets the header say "the top story alone" as a fact rather than a guess, and it is backed by the contract at `docs/api.md`.

So I exported the selection rule as `selectWithinBudget()` and tested it against story sets **built to force the state**, rather than hoping the fixtures produce it:

- a six-minute story under a five-minute budget yields exactly one story, and that selection exceeds the budget;
- a second story that would push past the budget is skipped, while a later one that fits is taken.

Both assertions are unconditional. No `if` can skip either.

**Control, run before trusting it:** changing `toHaveLength(1)` to `toHaveLength(999)` reddens exactly that test, by name, with the rest of the file green — the same method you used to prove the original was vacuous.

**The non-blocking note, taken.** Both `toEqual([])` tests now carry a floor asserting the fixture set is non-empty. You are right that seven other tests already redden on an empty set, so it was not load-bearing — but a test should not depend on its neighbours to mean something, and the next person to read one in isolation should not have to discover that it is propped up from elsewhere.

195 passed / 48 skipped, and prettier, eslint and tsc all exit 0.

Thank you for sweeping the family before filing. Knowing it was the only conditionally-guarded assertion in the new tests — one instance rather than a pattern — is the difference between fixing a line and going looking for four more.

zarghammir:

Re-reviewed at **`981b2d2`**. The finding I raised is properly fixed and I confirmed its control. But the fix introduces a new instance of **this PR's own defect family**, and I would rather say so plainly than let it land — the work here is good and this one is genuinely subtle.

## What is fixed, confirmed by control

**The vacuous test is gone and its replacement can actually fail.** `toHaveLength(1)` → `toHaveLength(999)` reddens *"holds exactly one story when the top story alone exceeds the budget"* and nothing else — 1 failed, 10 passed, zero new type errors. The previous version stayed green under the same treatment. That is the difference, measured.

Keeping the coverage claim rather than deleting the test was the right call, and the comment recording *why* the old one could not fire — including that the PR body had said so in prose one document away — is worth more than the fix itself.

**Both `toEqual([])` tests now carry floors**, with the reason I would have given: *"a test should not depend on its neighbours to be meaningful."* Seven other tests did protect them; that protection is now local.

---

## The new finding: the fixtures select differently from the server

`selectWithinBudget` (`fixtures.ts:305`) uses **`continue`**. The real implementation on `main` — `takeWithinReadingTime` in `src/api/brief.ts:149`, which #42 landed — uses **`break`**.

```
fixtures.ts:305   if (chosen.length > 0 && spent + s.readingMinutes > budget) continue;
brief.ts:149      if (taken.length > 0 && minutes + story.readingMinutes > target) break;
```

They are not the same rule. Simulated over identical story sets:

| stories (minutes) | budget | server (`break`) | fixture (`continue`) | |
|---|---|---|---|---|
| `[1,1,1,1,1,1]` | 5 | `[1,1,1,1,1]` | `[1,1,1,1,1]` | agree |
| `[6]` | 5 | `[6]` | `[6]` | agree |
| `[6,7,3]` | 10 | `[6]` | `[6,3]` | **diverge** |
| `[4,8,2,1]` | 10 | `[4]` | `[4,2,1]` | **diverge** |

**The contract is unambiguous and the server matches it**, not the fixtures — `docs/api.md:373` says the brief returns stories *"until the cumulative reading time would exceed the target"*. "Until … would exceed" is `break`.

**And a new test now pins the divergent behaviour.** *"never lets a second story push the selection over the budget"* builds `[4, 4, 1]` against a 5-minute budget and asserts `[9101, 9103]` — first and third. Its comment states the divergent step outright: *"4 fits, 4+4 would exceed 5 so it is skipped, 4+1 fits."* Against the server's rule that same input returns `[9101]` alone.

So the fixtures model a brief the API does not produce, and a test now holds them there.

**Why this is invisible today, and why that is the point.** Every fixture story is one minute, and with uniform durations `continue` and `break` are identical — the table above shows both agreeing on today's data. The divergence appears the moment reading times vary, which this PR's own body says arrives *"the day a summariser writes longer bodies"*. That is the same conditional the over-budget header is kept for, pointed at a place nobody is watching.

**This is the family this PR exists to close.** The original defect was fixtures carrying article time instead of computed reading time, so a ten-minute brief held two stories instead of six — fixtures describing a product that does not exist. `fixtures.test.ts` even has a whole describe block named *"fixtures are states the pipeline can actually produce"*. The selection **rule** is now the thing that does not match the pipeline, one level up from the data.

### What I would do

This branch is **14 commits behind `main`** and has to rebase anyway — at which point both functions sit in the same tree for the first time and the divergence becomes visible to a reader as well as to a simulation. In that rebase, have the fixtures use the real rule: import `takeWithinReadingTime` from `src/api/brief.ts`, or have `selectWithinBudget` delegate to it. One rule, one implementation, and a fixture cannot drift from the server by construction.

That is the same shape as the exit-code wrapper in #47: two copies of a contract that a control can only ever reach one of is how the copies drift — except here the two copies already disagree, and only uniform test data is hiding it.

If the `continue` behaviour is actually wanted — a greedy fill genuinely reads better than a strict prefix — then the change belongs in `brief.ts` and `docs/api.md` first, and the fixtures follow. What should not happen is the fixtures deciding it alone.

## A correction to my own verdict

My first review said this branch carried current `main` with "main ahead by 0". **That was true when I measured it and is false now** — the owner clicked #42, #44 and #47 after I posted, and `main` moved fourteen commits. I am flagging it rather than quietly re-measuring, because it is the same shape as the CI-coverage claim I had to correct on #31: base currency is a fact with a timestamp, and a verdict that states one without saying when it was taken is making a promise it cannot keep. It also matters here specifically — `src/api/brief.ts` did not exist in this worktree when I first reviewed it, which is why the divergence above could not have been found on the first pass.

## Verified myself

The control on the replacement test and that it reddens by name; both floors; the two selection rules read from their own files; their divergence simulated over six story sets; the contract's wording on `main`; base currency re-measured (14 behind, `MERGEABLE`/`CLEAN`); `tsc --noEmit` clean at **0** errors after a fresh install.

## Taken on trust

- **That `takeWithinReadingTime` is what `/api/brief` actually calls.** I read it on `main` and matched it to the contract; I did not drive the route.
- **The screenshots**, unchanged from my first review.
- **Real API behaviour** — everything here is fixtures and simulation.

One rebase and one rule, and I will approve.

zarghammir:
Fixed at `1dc76565249d6a880d8583aef40b0c72dcbc6bd9`. You were right on all of it, and the framing is the part I want to repeat back: **this is the same defect as the reading times, one level up.** That was *data* describing a product that does not exist. This is the *rule* not matching the pipeline — inside a file whose own describe block is named "fixtures are states the pipeline can actually produce".

**And my new test pinned the divergence while claiming to verify it.** It asserted `[first, third]` for `[4, 4, 1]` under five minutes: take the first, skip the one that does not fit, take a later one that does. That is greedy-fill, and I wrote a comment underneath spelling out the divergent step as though it were the specification.

**One implementation now.** I could not simply import `takeWithinReadingTime` from `src/api/brief.ts`, because that module pulls in drizzle — importing it from fixture code would ship the database layer to the browser bundle, and copying it is precisely what caused this. So the rule moved to `src/api/reading-budget.ts`, which has no database imports and is generic over anything carrying `readingMinutes`. `brief.ts` re-exports it, so every existing importer is untouched, and the fixtures call the same function rather than a lookalike.

**The test now pins the server's behaviour and documents what it used to pin:**

```
takeWithinReadingTime([4, 4, 1], "5") === [first]      // stops, does not fill the gap
```

**Control, run before trusting it.** Changing `break` back to `continue` in `reading-budget.ts` reddens exactly that test, by name, and produces `[9101, 9103]` — the precise output the old assertion demanded — with the rest of the suite green. The control reproduces the defect rather than merely failing.

`tsc`, `eslint`, `prettier` all exit 0; 209 passed / 88 skipped.

**Also refreshed, not quietly corrected:** the PR body said "the routes that serve it are not merged". They landed on `main` while this PR was in review. The sentence now says so explicitly and explains that fixture mode stays for this PR because switching the data source mid-review would invalidate everything you just verified, with the flip following separately.

Thank you for checking whether the fixture rule matched the server's rather than stopping at whether the fixture data was right. The first finding taught me to check my data against the pipeline; I did that, added a test for it, and did not think to check the *rule* the same way.

zarghammir:
REVIEW VERDICT: APPROVE

Verified at **`1dc7656`**. Carries current `main` (0 behind, rebased again), `MERGEABLE`/`CLEAN`, 25 files both ways, all three CI jobs green. My own run after a fresh install: `npm run typecheck` **0**, `eslint` **0**, `prettier` **0**, **297 tests, 23 files, 0 skipped**.

Two conditions below that do not block the merge. Everything I raised is fixed, and the control for it is the best piece of evidence this project has produced.

## The control reproduces the defect rather than merely failing

Changing `break` back to `continue` in `reading-budget.ts` reddens exactly one test — *"stops at the first story that does not fit, rather than filling the gap"* — with zero new type errors. And the assertion message is the whole argument:

```
AssertionError: expected [ 9101, 9103 ] to deeply equal [ 9101 ]
```

`[9101, 9103]` is **the precise output the previous version of this test demanded**. So the control does not just go red; it reproduces the exact defect and shows that the earlier assertion would have certified it. The test even pins the old answer as forbidden — `expect(...).not.toEqual([9101, 9103])` — which names the bug it exists to prevent rather than describing the behaviour it wants.

I have spent today removing tests that assert nothing. This is the opposite and it is worth saying plainly: **a test can be written that certifies a bug, and the only way to tell the two apart is to run the defect through it and read what comes out.** An exit code would have looked identical either way.

## The move is genuinely pure

I checked the claim rather than the description, because a "pure move" that quietly changes an exported shape is invisible to a green suite on both sides:

- The function body in `reading-budget.ts` is **byte-identical** to `main`'s `brief.ts` apart from the signature — `diff` shows only the generic parameter and `const taken: T[]`.
- `brief.ts` re-exports `BRIEF_LENGTHS`, `takeWithinReadingTime` and the `BriefLength` type, so every existing importer is untouched.
- The generic `<T extends { readingMinutes: number }>` is a widening: a caller passing `StoryCard[]` still gets `StoryCard[]` back.
- The new module imports nothing from the database layer, which is the reason it exists — importing `brief.ts` from fixture code would have shipped drizzle to the browser. **That is a better fix than the one I suggested**, and the reason is in the file.

`BRIEF_LENGTHS` is now defined in both `reading-budget.ts` and `db/schema.ts` — but it was already in two places on `main` (`brief.ts` and `schema.ts`), so this PR moves a duplicate rather than adding one. Noting it for whoever next touches either, not against this PR.

## Two conditions

**1. Two statements the correction missed.** The body's opening paragraph was rewritten to say the routes landed during review — and that is true; `/api/brief/route.ts` is on `main`, along with twelve others. But the same claim survives in two more places:

- `src/lib/api/client.ts:7` — *"The routes on feat/api-routes are not merged, so this reads fixtures."*
- `src/lib/api/client.ts:13` — *"Flip by setting `NEXT_PUBLIC_USE_FIXTURES=0` once `/api/brief` exists."* It exists.
- the body's own summary line — *"No live data. Fixture mode until the routes merge."* They merged.

The code comment is the one that matters: it is what the next contributor reads when deciding whether to flip the flag, and it tells them to wait for something that has already happened. This is the third time this PR has met the same family — fixture *data* describing a product that did not exist, then the selection *rule*, now the *explanation* — and the correction was applied to one instance of it.

**2. The follow-up needs a number.** The body says fixture mode is *"flipped in a small follow-up"*, and there is no issue for it — I checked the open list. On this project a deferred item without an owner is a dropped item, and this one carries the stale comments above with it. Either fix the three lines here, or open the issue and put them in it. Not optional; it is just not worth holding the merge for.

**Keeping fixture mode for this PR is right**, and the stated reason is the correct one: switching the data source mid-review would have invalidated everything already verified. The explicit-flag-not-a-fallback reasoning is also right and worth defending — a client that silently falls back on a 404 cannot distinguish *"this route does not exist"* from *"this save failed"*, and would report a failed write as a success.

## Verified myself

The control and what it produced; the move's purity by diffing the function body against `main`; the re-exports; base currency re-measured after the rebase; the contract's wording; `/api/brief` on `main`; the two stale comments and the missing ticket; typecheck, lint, format and the full suite after a fresh install; all three CI jobs.

Earlier in this review I also measured the two selection rules diverging over six story sets, ran the 24-state accessibility sweep myself, and re-proved that sweep's floors on this branch by renaming the bottom bar's landmark — 0 clipped and 0 overflow, which without the floor reads as a clean pass.

## Taken on trust

- **That `takeWithinReadingTime` is what `/api/brief` actually calls at runtime.** I read it, matched it to the contract and diffed the move; I did not drive the route.
- **The screenshots**, regenerated after the fixture correction — I measured the live pages instead of diffing images.
- **Real API behaviour end to end.** This PR runs on fixtures by design, so every claim about the server is read from `main`'s source rather than exercised.
- **A bare `tsc --noEmit` reported 12 errors here and I did not report them**: all twelve were `RouteContext` constraint failures against a one-route union, from `.next/types` generated before the rebase brought twelve routes in. The repo's own `npm run typecheck` runs `next typegen` first and exits 0. When your instrument and CI disagree at the same SHA, suspect the instrument.

---

Approved. This is my last review on this repository — the seat moves to a Fable window after this, at the boundary I proposed. Everything I verified and everything I did not is in these comments rather than in my head, deliberately, because the next reviewer should check rather than inherit.

Issue #41, opened 2026-09-16
Five content types are unreachable, and MODEL carries a top ranking weight


zarghammir:
**Found by the UI lane while labelling the badges. Not rare — unreachable.**

There is no content-type classifier. One line does the whole job, `src/pipeline/normalize/index.ts:34`:

```ts
const contentType = item.contentType ?? source.defaultContentType;
```

Nothing inspects the title or the text. Only two adapters override the source default: arXiv hard-codes PAPER, Hacker News uses DISCUSSION for a post with no URL. Everything else takes whatever its **source** declares — and the 18 seeded sources declare only four defaults between them: NEWS ×12, RESEARCH ×4, RELEASE ×1, PAPER ×1.

**So five of the ten content types cannot occur: TREND, TOOL, MODEL, BUSINESS, REGULATION.** There is no input to the system that produces them.

### Why MODEL is the one that matters

`WEIGHTS.contentType` gives **MODEL a 6 — tied with RELEASE for the highest weight of any content type.** The ranker is scoring a category nothing can emit. "A new model shipped" is arguably the single most valuable thing this product can tell someone, and today it lands as NEWS like everything else. TOOL (4) and REGULATION (3) are dead weights on the same basis.

A weight that can never apply is not neutral: it makes the ranking look considered while doing nothing, and it will silently start firing the day a source's default changes.

### Why the badge is nearly a constant

Twelve of eighteen sources default to NEWS, so most cards carry the same badge. A foreign-ministry story from NPR is NEWS; an acquisition is NEWS; a model launch covered by a newsroom is NEWS. The category exists in the schema, the ranker and the UI, and is carried by nothing.

### The machinery already exists

`matchesAnyKeyword` against `topics.keywords` is how **topics** are assigned. It is simply never applied to content type. This is a gap rather than a missing capability, which probably makes it a small ticket.

### What to decide

1. Classify content type from the item's own text (title at minimum), falling back to the source default — reusing the keyword matcher rather than inventing a second mechanism.
2. Or narrow the enum to what can actually occur, and drop the dead weights from `WEIGHTS.contentType`.

Either is defensible; shipping neither means the ranker keeps scoring categories that cannot exist.

[...]

Issue #38, opened 2026-09-16
A source that fails on every hosted run is invisible (import-ai returns 403 to GitHub runners)


zarghammir:
Found by the Platform lane in the first real container run of the worker, on a GitHub runner:

```
import-ai: FAILED — HTTP 403 for https://importai.substack.com/feed (0 fetched, 0 new)
total: 18 sources, 496 fetched, 496 new, 482 stories, 1 failed in 9.0s   exit 0
```

The same source fetches 20 items fine from a laptop, so this is **Substack refusing a datacenter IP**, not a broken adapter. `.github/workflows/ingest.yml` runs on GitHub runners, so **the hosted schedule loses import-ai on every single run.** Self-hosters are unaffected.

**The exit code is correct and must not change.** One dead feed out of eighteen keeping the job green is the behaviour we want — a scheduled run that goes red every time one publisher blocks a datacenter is a notification people mute, and then a real outage goes unseen with it.

**The defect is that the loss is invisible**, which is this project's recurring shape: a source failing on every run is indistinguishable from a quiet news day. What is needed is not a louder exit code but a way to tell those apart:

1. Consecutive-failure detection per source — `ingest_runs` already records the error, so the data is there.
2. A source that has failed N runs in a row is surfaced as unhealthy rather than silent. The natural home is the Settings source list (#16) via the sources route (#10), which already carries `lastError`.
3. A decision on import-ai specifically: accept the loss on hosted runs, find another feed for it, or fetch it by a route that is not refused.

Not issue #26 — that covers sources with **no usable feed**. This one has a working feed that is refused by IP, which is a different failure and a different fix.

Filed by the PM so the loss has an owner rather than living in a container log.

zarghammir:
**Owner ruling, 2026-09-17: drop it.**

> "You can skip the dead source. It doesn't have any good for us."

So the three options in this ticket resolve to the first — **accept the loss** — and one step further: remove `import-ai` from the seeded catalogue rather than leaving a source that fails on every scheduled run forever.

**What that changes:**
- The seed row goes. Eighteen sources becomes seventeen.
- The scheduled run stops carrying a permanent `1 failed`, which matters now that #86 has landed: a health signal is only useful while a red means something, and a source that is red **by design** trains everyone to ignore it. **A known-failing source left in place is how an alarm becomes wallpaper.**
- The reason goes in the seed file next to where the row was, so the next person does not re-add it from the same list of good AI newsletters and rediscover the 403 themselves.

[...]

Issue #26, opened 2026-09-16
HTML listing adapter for sources without a feed (Anthropic news, Meta AI, xAI)


zarghammir:
**Problem.** Three PRIMARY sources named in the brief publish no RSS/Atom feed: Anthropic (7 candidate URLs, all 404), Meta AI (4 candidates, 404), xAI (403 on every candidate). The seed (PR #24) leaves them out rather than seeding broken rows, so their announcements only reach the app second-hand through reporting or Hacker News.

**Solution.** A `html` source kind: an adapter that fetches a listing page, extracts items with a per-source CSS selector config (link, title, date, excerpt) stored in sources.config, respects robots.txt and a polite interval, and falls back to unchanged-etag skipping. Seed Anthropic news and Meta AI with selectors verified at authoring time; xAI only if it stops returning 403 to a polite user agent. Fixture per page for offline tests.

**Expected output.** sources:check shows Anthropic and Meta AI OK with item counts. An Anthropic announcement appears as PRIMARY_SOURCE within one worker pass of publication. A selector that stops matching is reported in ingest_runs.error, never as a silent zero.

Issue #18, opened 2026-09-16
Clone-and-run proof plus architecture and source-adapter docs


zarghammir:
**Problem.** The promise is that a stranger can clone, self-host and extend it. Nobody has tried, and the docs do not exist.

**Solution.** On a clean environment (fresh container or VM), follow README literally and record every step that fails; fix the docs or file issues. Write `docs/architecture.md` (pipeline diagram, module boundaries, data model), `docs/sources.md` (adapter contract, how to add one in five steps, tier guidance), `CONTRIBUTING.md` (setup, scripts, lanes, PR expectations, commit style), and expand README with required vs optional configuration, screenshots section (from the QA screenshot script), and privacy note (no trackers, no external reporting of reading behaviour).

**Expected output.** A log of the clean-machine run showing the app up with data by the third command. Docs reviewed by the PM for plain language. A new contributor could add an RSS source by editing the seed only.

zarghammir:
Addition from the reviewer's baseline (2026-09-16): the repo carries no written engineering standard beyond the auto-generated AGENTS.md block. This ticket also delivers `docs/review-protocol.md`: the review bar every application PR is held to (base currency, CI per head SHA with every step executed, file counts via `gh pr diff --name-only`, the four light checks run by the reviewer, boot-and-click at 1440 and 390 in light and dark with a clean console and keyboard path, claims reproduced with the reviewer's own scripts, deletion controls on tests, the product rules that are findings: separate verification and content-type fields, no internet-reachable route may trigger ingestion or a paid call, keys only in env, no external reporting of reading behaviour, branding only via src/config/brand.ts, no fabricated metrics) and the verdict convention (one PR comment whose first line is exactly REVIEW VERDICT: APPROVE or REVIEW VERDICT: REQUEST CHANGES; nobody merges but the owner). Also: README already points at docs/, so this ticket must leave no dangling reference.

zarghammir:
Carried over from PR #1 review: src/db/schema.ts:17 and README.md:13 already point at docs/ (architecture.md), which does not exist. This ticket creates it; until then the pointers are known-dangling and accepted.

PR #53, opened 2026-09-17
docs: how to add a source, and what it does to the ranking

zarghammir:
Part of #18 — the source-adapter half. The architecture overview is deliberately **not** here; see below.

## What it is

`docs/sources.md`: how a contributor adds a source, what an adapter must return, what the seed overwrites, and what adding a source does to the ranking.

Written from the code, not from memory. Every number, field list and behaviour in it was read out of the file it describes, and the one claim I could not find support for I checked and stated as zero rather than leaving blank — `DISCOVERY` does not appear in `score.ts` at all.

## The part worth reading if you read nothing else

**Adding a source is a ranking decision, and nothing tells you that today.**

`tier` is the largest single input to a story's score: `PRIMARY` is 20 points where recency maxes at 14, and only the *highest* tier among a story's sources counts. So calling a newsletter `PRIMARY` because it is well-informed outranks a genuinely fresh story from a newsroom. The document says what the field means — whose word this is, not how good it is — because the field name does not.

## Three things pointed at their issues rather than left as folklore

- **A site with no usable feed** needs the HTML listing adapter from #26, which does not exist. The document says explicitly not to work around it by pointing an `rss` source at a page that is not a feed: the fetch succeeds, the parse yields nothing, and the source looks healthy while contributing zero items.
- **A working feed can still be refused by IP.** Substack 403s GitHub's runners, so `import-ai` fails on every hosted run while working from a laptop — #38. The consequence spelled out: `sources:check` passing locally does not prove a source works where it runs, and because one dead feed deliberately does not fail a scheduled run, a source that never works in production can look fine indefinitely.
- **Five of the ten content types cannot occur today** — nothing anywhere produces `TREND`, `TOOL`, `MODEL`, `BUSINESS` or `REGULATION` — while `MODEL` carries the joint-highest content-type weight. That is #41, and a contributor whose source really does publish model releases is told to say so there rather than pick the nearest type that already works.

## Why the architecture overview is not in this PR

#42 adds twelve API routes and rewrites the database client into `getDb()` / `getSql()`. An architecture document written today would describe a system with no API layer and an eager client, and would become false the moment #42 merges. Writing it after is cheaper than correcting it after, and this project has spent a day on statements that were true when written.

## Why the seed is documented anyway, when #42 changes that file

#42 edits `src/db/seed.ts`, so the obvious move was to hold this section too. Reading the diff said otherwise, and the distinction is worth stating because the next person documenting anything will face it.

**#42's change to the seed is plumbing: `db` → `getDb()`.** What the seed *does* — matches on `key`, updates the catalogue fields, never touches `enabled`, never deletes, reports orphans — is identical before and after. So this document describes the behaviour and says nothing about how the client is obtained, which is the half under review.

The transferable form: **document the behaviour, not the plumbing, when the plumbing is under review.** Had this described the wiring, it would have been false the moment #42 merged; describing the behaviour, it survives.

## Scope check

The clone-and-run proof, the other half of #18, is also not here. It measures the Dockerfile and compose, which #44 changes, so proving it against today's main would produce a result that is stale before it is read.


zarghammir:
REVIEW VERDICT: APPROVE

Reviewed at **`a808f76`**. One new file, `docs/sources.md`, 182 lines. Merges clean into current `main`, counts agree both ways, both CI jobs green.

A document's only defect is being wrong, so I checked every claim in it that names a field, a number, a file or a behaviour. **All of them hold.** The weights table in particular is a dozen numbers and I expected one to be off; none is.

## Verified against the code, not read

| claim | checked against |
|---|---|
| `db:seed` updates `name`, `kind`, `tier`, `url`, `homepage`, `config`, `defaultContentType` | `seed.ts:43-51` — exactly that set, in that order |
| **never touches `enabled`** | absent from the `set`, with `// Catalogue fields only. enabled is deliberately absent.` beside it |
| never deletes; reports orphans | `seed.ts:58,83,106` — orphans computed and printed, nothing removed |
| `SourceAdapter` is `{ kind, description, fetch }` | `sources/types.ts:29-34`, exact |
| registered by `kind`, nothing else changes | `sources/registry.ts` — a `Partial<Record<SourceKind, SourceAdapter>>` and two accessors |
| `ctx.log` lines are stored on `ingest_runs.error` **if that source fails** | `run.ts:349` — `error ? [error, ...lines].join("\n") : null`. The conditional is in the doc and it is the right one |
| `sources:check` reads the catalogue, needs no database | imports `SOURCE_SEEDS`; no `db/client`, no `DATABASE_URL` anywhere in it |
| exits non-zero if any source fails | `process.exitCode = 1` / `process.exit(1)` |
| one dead feed does not fail a scheduled run | `exitCodeFor` returns 1 only when **every** source errored |
| no HTML-listing adapter exists (#26) | `src/sources/` holds `arxiv`, `hackernews`, `rss` and nothing else |
| new kinds are a migration | `SOURCE_KINDS` feeds `pgEnum("source_kind", …)` |

**The ranking table, every cell:** `PRIMARY` 20, `HIGH_QUALITY_REPORTING` 12, `ANALYST` 9, `COMMUNITY` 4 — all match. `recencyMax` 14 and `topicMax` 22 — match. All ten content-type weights match, and the count agrees: ten in the doc, ten in `WEIGHTS.contentType`. And `DISCOVERY` really does not appear in `score.ts` at all, which is what makes the **0** in that row a fact rather than an inference.

## The part that makes this worth having

Three things in here are the sort of knowledge that normally only exists in someone's head:

**"Tier means 'whose word is this', not 'how good is it'."** With recency capped at 14 and a topic match at 22, calling a newsletter `PRIMARY` outranks a genuinely fresh newsroom story. That is a ranking consequence of a catalogue edit, and a contributor has no way to discover it from the seed file.

**The silent-failure trap.** Pointing an `rss` source at a page that is not a feed makes the fetch succeed, the parse produce nothing, and the source look healthy while contributing zero items. That is the absence-versus-failure shape this project keeps finding, written down prospectively for once instead of after it bit.

**"`sources:check` passing on your machine does not prove the source works where it will run."** Substack returning 403 to hosted runners, combined with a deliberately tolerant exit code, means a source can be permanently dead and permanently green. The doc says to read `ingest_runs` rather than trust the run — which is exactly right, and is the same "a green check may be an exemption" lesson from a different angle.

**And it states its own gap rather than working around it.** The `#41` callout names the five unreachable content types and says `MODEL` carries the joint-highest weight on a category nothing emits. I verified that list independently while reviewing #41 — reachable today are `NEWS`, `RESEARCH`, `RELEASE`, `PAPER`, `DISCUSSION`; unreachable are `MODEL`, `TOOL`, `BUSINESS`, `REGULATION`, `TREND` — and the doc's five are exactly mine. Telling a contributor to *"say so on #41 rather than quietly picking the nearest type that already works"* is better advice than any correction to the table would have been.

## One word, not a finding

Line 38 says `sources:check` *"fetches every catalogue source"*. It fetches every **enabled** one — `scripts/sources-check.ts:99` filters `s.enabled ?? true`. A seed with no `enabled` field defaults to true, so a source added the way this document describes is always checked, and the gap is only reachable for a deliberately-disabled seed. One word would close it, and I mention it only because the rest of the document is precise to the cell.

## Verified myself

Every row in the table above, plus base currency, merge state, file counts both ways and both CI jobs. The content-type reachability claim I had already established independently on #41 rather than taking it from this document.

## Taken on trust

- **Substack's 403 to GitHub runners** (#38) — an external service's behaviour I cannot observe from here; it is attributed to an issue rather than asserted bare.
- **That #26 is the right home for an HTML listing adapter** — I confirmed no such adapter exists, not that the issue says what the doc says it says.
- **The prose I did not machine-check** — I verified every sentence naming a field, number, file or behaviour; the explanatory passages I read but did not re-derive.

zarghammir:
**Approval carries to `410bb58`.** The word is fixed and the consequence added, and I confirmed the consequence is true rather than merely reasonable — which is the part worth having.

The correction: `sources:check` fetches every **enabled** catalogue source. `scripts/sources-check.ts:99` is `SOURCE_SEEDS.filter((s) => s.enabled ?? true)`.

The consequence the lane added — *"a disabled source is never proved to work before someone switches it on"* — holds end to end, and it takes three files to see:

1. `sources-check.ts:99` filters an `enabled: false` seed out, so it is never fetched through its adapter.
2. `seed.ts:37` **inserts** it with `enabled: s.enabled ?? true`, so it lands in the database switched off.
3. `seed.ts:42` deliberately omits `enabled` from the conflict update, so re-seeding never flips it back on.

So the source sits in the catalogue and the database, unverified, until somebody enables it through `PUT /api/sources/:key` — at which point the worker fetches it for the first time in production, having never fetched it anywhere else. That is a real gap in the path this document exists to describe, and it is better documented than corrected away.

**Adding the consequence rather than only the adjective is the difference between a document that is accurate and one that is useful.** "Every enabled source" tells a reader what the command does; the second sentence tells them what it means for them, which is the only reason they opened the file.

Everything in my verdict at `a808f76` stands — every field, number, file and behaviour I checked there is unchanged by this diff, including the twelve-cell weight table and the `#41` callout.

PR #54, opened 2026-09-17
fix(ci): give every main commit its own run, so no merge loses its verdict

zarghammir:
A comment on main promised a guarantee the code did not provide, and tonight it cost the thing it was written to protect.

## What happened

Three pull requests merged inside two minutes. The middle one got no CI verdict at all:

```
ea426c2  (#42)  completed success
59a8714  (#44)  completed cancelled   <- 0 jobs ever started
97f26ef  (#47)  in progress
```

`main@59a8714` was cancelled while queued, before a single job ran. So the tree with #42 and #44 but **without** #47 was never tested, and if a later red appears there is no bisect point between them.

The comment directly above the concurrency block said:

> On main nothing is ever cancelled: two merges landing inside a run's ~70 seconds share the ref refs/heads/main, and cancelling there would leave the first commit with no verdict and no way to tell which merge broke a later red.

## Why the code did not do that

`cancel-in-progress: false` stops a **running** run being cancelled. It does not stop a **pending** one being superseded: GitHub keeps only one pending run per concurrency group and cancels the older when a new arrival queues behind it.

Putting the sha in the group on push makes each merge commit alone in its own group, which is the only arrangement that cannot be superseded. Pull requests are unchanged — same group per ref, still cancelling in progress.

```yaml
group: ci-${{ github.ref }}-${{ github.event_name != 'pull_request' && github.sha || '' }}
```

**Why "not a pull request" rather than "is a push".** Identical today; different the day someone adds an event. Naming the event that *gets* isolation means a later `workflow_dispatch` or `merge_group` silently opts out — those runs would share `ci-<ref>` with `cancel-in-progress` false, which is this same bug reintroduced for a new event by an edit that never touches this line. Inverted, a new event inherits isolation and has to be deliberately excluded to lose it. Please do not "clarify" it back to naming `push`.

**Cost:** two merges landing together both run instead of one. That is the price of the guarantee the comment already claimed — and a merge commit with no verdict is the thing you cannot buy back later. You can re-run it, but only if somebody notices, which is luck rather than a process.

## The transferable part: fix-one-instance applied to a belief

I learned this exact behaviour of `cancel-in-progress` **earlier the same day**, when it made me correct the identical wrong belief in `ingest.yml`'s concurrency comment — that one claimed a mid-flight run is *skipped* where GitHub in fact *queues* it. I corrected the sentence where I found it and never asked whether I had acted on the same belief anywhere else. I had, one file away, in the workflow that guards main.

**When a wrong belief about a tool is corrected, sweep every other place that belief was acted on.** The usual form of this rule is about code — fix one instance, sweep the family. This is the same rule applied to something I believed rather than something I wrote.

## The sweep, done properly

Both workflows, every concurrency block, and every comment asserting GitHub's behaviour rather than our own code:

- `ingest.yml` concurrency — already correct; it is the one I fixed this afternoon.
- `ci.yml` concurrency — this PR.
- `ci.yml`, the browser job's cost — **estimated 3-4 runner minutes; it is 118 seconds.** Measured on run 35166985867: containers 21s, install 17s, build 21s, sweep 34s, shell checks 8s. The comment now carries the measurement, the breakdown and the note that the earlier figure was a guess that was nearly double.
- The remaining behavioural comments (the npm 11 pin, Chrome on the runner image) were verified today and stand.

## Recovery

`main@59a8714`'s run has been re-run and is queued — a cancelled run with zero jobs can be restarted, so this particular verdict is recoverable. It is queued behind `97f26ef` under the old grouping, which is the defect being fixed here doing its thing one last time.


zarghammir:
REVIEW VERDICT: APPROVE

Reviewed at **`5531031`**. One file, carries current `main`, `MERGEABLE`/`CLEAN`, counts agree both ways, all three jobs green.

The diagnosis is correct, the fix is correct, and the comment it replaces was wrong in exactly the way that cost a verdict last night. One hardening suggestion at the end, not a finding.

## Both corrections are actually in the diff

The lane disclosed that `git checkout <branch> -- .github/workflows/ci.yml` overwrote its uncommitted work and silently discarded both edits, and that it caught this by grepping and getting 0 and 0. **A re-application that did not take looks exactly like one that did**, so I checked the diff rather than the account: both are present — the concurrency group with its rewritten comment, and the browser job's cost line.

## The mechanism, and why the old comment was wrong

`cancel-in-progress: false` prevents a **running** run being cancelled. It does not prevent GitHub superseding a **pending** one, because only one pending run per concurrency group survives. Three merges inside ninety seconds, one concurrency group (`ci-refs/heads/main`), and `main@59a8714` was cancelled before a single job started.

The comment directly above that block promised the opposite in as many words — that nothing on main is ever cancelled, *because* cancelling there would leave a commit with no verdict. **The code did not do what the comment claimed, and the thing the comment existed to protect is exactly what was lost.** The replacement says so plainly, names the commit and the date, and keeps the superseded reasoning visible rather than quietly deleting it.

## The expression isolates pushes and leaves pull requests untouched

```
group: ci-${{ github.ref }}-${{ github.event_name == 'push' && github.sha || '' }}
```

`A && B || C` behaves as a ternary in GitHub expressions, so:

| event | group | effect |
|---|---|---|
| `push` to main | `ci-refs/heads/main-<sha>` | unique per commit — **cannot be superseded** |
| `pull_request` | `ci-refs/pull/N/merge-` | all runs for a ref share one group — **unchanged** |

The PR group gains a trailing dash, which changes the group's *name* and not its *behaviour*: every run for a ref still lands in the same group, and `cancel-in-progress` is still `true` only for `pull_request`. A push and a pull-request run can never share a group.

**I checked the event surface rather than assuming push and pull_request**, which is where a conditional like this would hide a collapse: this workflow's `on:` block is exactly `pull_request` and `push: branches: [main]`. There is no third event to land in the `''` branch unexpectedly, so the conditional is complete for what the file actually fires on.

## The family sweep — 149 comment lines, and the one that matters agrees

I swept both workflows for comments asserting GitHub's behaviour or cost rather than our own code. The browser job's stale 3-4 minutes is corrected to **118 seconds** with the per-step breakdown and the run id, keeping its own guess in the record: *"An earlier version of this comment said 3-4 minutes, which was a guess and was nearly double."*

The one that could have been a second instance is `ingest.yml:14-18`, and it is **already right**: *"a run arriving while one is going is QUEUED, not skipped — and if a third arrives while one is already pending, GitHub cancels the pending one so the newest takes its place."* That is the same GitHub mechanism this PR just discovered, described correctly, and reaching the **opposite** conclusion for good reason — the schedule *wants* supersession, because at most one waiting run is the property it needs and a backlog is the failure it fears. Two comments about one mechanism, two different requirements, both accurate. That is what a swept family looks like.

## The verdict chain, verified commit by commit

| commit | jobs | |
|---|---|---|
| `ea426c2` (#42) | 2, both green | |
| `59a8714` (#44) | 2, both green | **the one that lost its verdict, now recovered** |
| `97f26ef` (#47) | 3, all green | |

`59a8714` showing **two** jobs rather than three is correct rather than a missing job, and I confirmed it at the tree rather than reasoning from the date: `git show 59a8714:.github/workflows/ci.yml` contains **zero** occurrences of the browser job, because that commit predates #47. A rerun of an older commit reports the jobs that commit defines.

`97f26ef` is the combination nobody had run before — the browser job against #42's twelve routes and rewritten database client, and `compose-smoke` building an image with #44's layers over #42's routes. Three green.

## One hardening suggestion, not a finding

The conditional names the event that gets isolation. That is complete today, and it means **a new event silently opts out**: add `workflow_dispatch` or `merge_group` tomorrow and those runs share `ci-<ref>-` with `cancel-in-progress: false`, which is precisely the bug this PR fixes, reintroduced for a new event by an edit that never touched this line.

Inverting the test makes the safe behaviour the default:

```
group: ci-${{ github.ref }}-${{ github.event_name != 'pull_request' && github.sha || '' }}
```

Identical for both events today — `pull_request` still collapses to the ref, `push` still gets its sha — and a future event inherits *isolation* rather than *collapse*. It is the same shape this repository has reached for repeatedly: make the bad state unexpressible rather than correct for it. Worth taking while the file is open; not worth holding the merge.

## Checked, and outside this PR

`ci.yml:57-58` claims npm 11 *"omits 27 nested entries that npm 10 expects"*. That is a precise number about another tool's lockfile format, it is **on main already and this diff does not touch it**, and I could not verify it without npm 10 installed. Flagging it for a future pass rather than as a finding here — it is the shape that drifts, in a repository whose lockfile has changed several times today.

## Verified myself

Both corrections present in the diff; the event surface; the group each event produces and that PR grouping is behaviourally unchanged; 149 comment lines swept for behaviour and cost claims; the three-commit verdict chain including `59a8714`'s job count against its own tree; base, merge state, file counts and all three jobs on this PR.

## Taken on trust

- **GitHub's pending-run supersession**, which I did not reproduce — the evidence is `59a8714`'s cancelled run with zero jobs started, which is consistent with it and with nothing else I can think of.
- **The npm claim above.**
- **That one extra run per displaced merge is the whole cost.** It follows from the grouping, but I did not observe two merges landing together under the new expression.

zarghammir:
**Approval carries to `1545225`.** The inverted form is genuinely identical for both events this workflow fires on — I evaluated it rather than eyeballed it.

| event | old form | new form | |
|---|---|---|---|
| `pull_request` | `ci-refs/pull/7/merge-` | `ci-refs/pull/7/merge-` | **same** |
| `push` | `ci-refs/heads/main-<sha>` | `ci-refs/heads/main-<sha>` | **same** |
| `workflow_dispatch` | `ci-refs/heads/main-` | `ci-refs/heads/main-<sha>` | differs — *not fired today* |
| `merge_group` | `ci-refs/heads/main-` | `ci-refs/heads/main-<sha>` | differs — *not fired today* |

The two forms diverge **only** on events this workflow does not fire on, and there the new one gives isolation where the old one gave collapse. That is the whole of the change: nothing moves today, and the thing that would have gone wrong tomorrow now goes right by default.

`cancel-in-progress` is untouched — still `github.event_name == 'pull_request'`. The only two lines mentioning it in the delta are prose.

The added comment is the part I would most want kept, because it makes the line self-defending: *"Naming the event that GETS isolation means any event added later silently opts out of it… Inverted, a new event inherits isolation and has to be deliberately excluded to lose it."* A reader who thinks `!= 'pull_request'` is a confusing way to say `== 'push'` now finds the reason before they "clarify" it back — which is the failure mode, since the two forms are indistinguishable on today's events and a test would not catch the difference either.

Everything in my verdict at `5531031` stands: both corrections present in the diff, the event surface enumerated from the `on:` block, the family swept across 149 comment lines, and the three-commit verdict chain with `59a8714`'s two-job count confirmed against its own tree.

Issue #56, opened 2026-09-17
Revisit the MODEL weight once the classifier has run against real traffic


zarghammir:
**Deliberately deferred, with a trigger — not an open question left hanging.**

`WEIGHTS.contentType` gives MODEL 6, tied-top with RELEASE. Until #41 there was no classifier, so that weight had never been applied to a real story. #41 measured what happens when it starts firing, on a live corpus of 516 items and 504 stories.

**What the measurement showed:**

- 9 of 201 ranked stories were retyped; only 2 reached the top 20.
- Both of those two are **misclassifications**: "Introducing computer use in Gemini 3.5 Flash" (a feature of an existing model) at rank 9, and "Training a 4B model to produce 81% faster query plans than Postgres" (a technique write-up) at rank 14 — up from 104.
- The two genuine model launches in the window scored 17.8 and 17.6 and sat at ranks **124 and 127** — not because of their type, since the +6 applied to them equally, but because they are older and less corroborated, and recency and verification dominate.

So in this corpus, the MODEL weight's entire visible effect at the top of the brief was to promote two wrongly-labelled stories, while the two correctly-labelled ones it also promoted stayed near rank 125 for unrelated reasons.

**Why the weight was NOT changed in #41**, and this is the part to preserve:

- Both errors are **classifier errors, not weight errors**. A correctly-typed corpus would put real launches near the top, which is what the weight is for. Lowering MODEL to hide two bad badges would be treating the instrument for the fault of what feeds it.
- Nine retyped stories in a single day's window is a thin basis for re-tuning anything.

**Trigger for this ticket:** once the classifier has run against several days of real traffic, re-measure with the same method #41 used — reconstruct the ranking without the classifier and diff the positions — and ask then whether 6 is right. Judge it on correctly-typed stories, not on the false positives.

Also worth checking at that point, because #41 exposed it: recency and verification dominate strongly enough that a genuine model launch from a few days ago lands around rank 125 regardless of its type. If that is not what the owner wants from a morning brief, it is a recency-weight question rather than a content-type one.

Owner: Intelligence lane. Filed from #41's own measurement at the lane's request.

zarghammir:
Addition from the Intelligence lane, and it changes when this ticket can be answered rather than just how.

[...]

PR #57, opened 2026-09-17
Classify content type from the item's title (#41)

### LINKED ISSUE #41: Five content types are unreachable, and MODEL carries a top ranking weight

zarghammir:
Closes #41.

> **Revised after review.** The review found a real defect in the backfill, and chasing it showed that the **same defect had corrupted this PR's own headline measurement**. The ranking numbers below are re-measured and the original conclusion is withdrawn — see "A correction to my own measurement".

## In plain words

Every story on the brief carries a badge — News, Release, Paper, Model, Policy and so on. Until now that badge came from the *source*, never the story: whatever a site was set to, every article from it got. Twelve of our eighteen sources are set to "News", so most of the brief said News regardless of what the article was. Five badges — Model, Tool, Business, Policy, Trend — could never appear at all, because nothing was able to produce them, while "Model" carried the joint-heaviest score in the ranking.

This reads each item's headline and works out what it actually is, falling back to the source's setting when the headline says nothing.

## A correction to my own measurement

My first version of this PR said: only 9 of 201 ranked stories were retyped, only 2 reached the top 20, and the Model weight's entire visible effect at the top of the brief was to promote two misclassified stories.

**That was wrong, and it was wrong for the same reason finding 1 is a defect.** To decide which stories had been retyped I used `stored !== sourceDefault` — the very inference the review showed cannot carry that meaning. It undercounted: the true figure is **34 of 225**, not 9. With provenance now recorded rather than inferred, the picture is different and better:

- **Ten genuine model launches moved from around ranks 100–124 into ranks 20–30.** "Gemini 3.5: frontier intelligence with action" 117 → 20, "GPT-6 Astra" 102 → 22, "Introducing WeatherNext 3" 118 → 24, "Introducing Gemma 4 12B" 120 → 25, "Introducing TabFM" 122 → 26, and so on. The weight is doing exactly what it is for.
- **The single largest move into the top ten is still a false positive.** "Introducing computer use in Gemini 3.5 Flash" went 81 → 8. It is a feature of an existing model, not a launch.

So the precision argument survives intact and is now properly grounded: **a false positive still buys about seventy places**, which is why precision here is worth roughly an order of magnitude more than recall. That sentence is in the source next to the rules it governs, not only in this PR. What does *not* survive is my claim that the weight only promoted mistakes — it promoted ten real launches too, and I could not see them because my instrument had the reviewed defect in it.

The weights are unchanged, and #56 will revisit them against several days of real traffic.

## Accuracy, including what it gets wrong

Measured against a live ingest of all eighteen sources — **515 items, 503 stories**. 40 items reclassified, **36 correct — 90%**.

| Type | Found | Correct | The ones it got wrong |
|---|---|---|---|
| MODEL | 20 | 17 | "Introducing computer use in Gemini 3.5 Flash" (a feature), "Granite 4.2 LLMs: How They're Built" (an explainer), "Training a 4B model to produce 81% faster query plans" (a technique write-up) |
| REGULATION | 8 | 8 | — |
| BUSINESS | 9 | 8 | "Catalyzing scientific impact through global partnerships" (a research post) |
| TOOL | 3 | 3 | — |

**Recall is about a third**, and the misses are structural. On a deterministic sample of items it left alone, roughly one in six should have moved: "We're launching Lyria 3.5", "TimesFM-3: A zero-shot foundation model", "Aurora 1.5: Extending open foundation models" are all model launches whose names are not on the hand-maintained family list. What would raise recall is a maintained model-name list or a summariser that reads the item — not a cleverer regex.

### The rule I measured and reverted

A generic "capitalised name + version number" rule, meant to remove that limit, found 15 more items and **got 8 of them wrong**, taking MODEL precision from 90% to 71%. The shape is too common: a date ("State of Open Models: Summer 2026"), newsletter numbering ("Import AI 454"), software versions ("Apple releases iOS 27", "Datasette 1.0a39"), an incidental figure ("Training Text-to-Image Models 3.6x Faster"). It is recorded in the source with those titles, so the next person reaching for the obvious improvement finds the measurement first.

### TREND stays unreachable, on purpose

Trend pieces share no vocabulary with each other. "The AI data center boom is colliding with cities scarred by big industry" and "The AI graveyard: a running list of projects and startups that didn't make it" are both trend pieces with no word in common that a non-trend headline lacks. #41 said five types were unreachable; after this, one still is, and now the code says why.

### Words rejected after measuring them

Kept as `REJECTED_KEYWORDS`, each with the title that convicted it — `merge` ("Anthropic merges Claude chat and Cowork" is a product story), `stake` ("What's at stake in AI's trillion-dollar gamble"), `enterprise`, bare `partner` ("A multi-agent AI partner"), bare `funding`, bare `policy` ("Conformal Policy Learning" is reinforcement learning), and three unproven ones.

One was only findable by measuring: `valuation` fired on "double-blind AI **evaluations**". The shared matcher does substring matching above four characters, which topic tagging needs and this does not, so `matchesPhrase` sits beside it — the same boundary rule the short-token branch already applies, at every length.

## Design decisions

**Precedence: adapter-declared > classified > source default**, owned by `decideContentType` in one place. The adapter wins because it reads structured data — arXiv's PAPER and Hacker News's DISCUSSION are facts about the item, not inferences from wording.

**Provenance is recorded, not inferred.** `raw_items.content_type_source` is one of `adapter | classifier | default`. See finding 1 below for why the obvious inference is wrong.

**`refreshStory` is unchanged.** It already recomputes `contentType` from the story's primary item, so a better-typed item joining later already moves the story's type once it becomes primary. Deliberately *not* added: taking the type from a better-typed non-primary item. The primary item is the story's identity for title, `firstSeenAt` and type alike, so taking the type from elsewhere would produce cards whose badge disagrees with their headline. That is correct behaviour, not a limitation.

**The classifier can never cross a content family.** `assignStory` groups by `contentFamily` (PAPER against everything else) and reads the *story's* recomputed type — the interaction found in #30. `CLASSIFIABLE` excludes PAPER by construction and a test guards both directions.

**The classifier must override RESEARCH, which the ticket did not anticipate.** The four RESEARCH-default sources are lab blogs, and lab blogs are where model launches are published. 14 of the 40 come from them — google-deepmind 11, google-research 2, microsoft-research 1.

**Title only, not title plus excerpt.** Several feeds carry a whole article body as the excerpt, and at 600 characters almost every rule fires on something.

**`contentFamily` moved** from `run.ts` into the content-type module. A move with no behaviour change — the module that owns the boundary should own the rule that must not cross it.

## Backfill

`npm run backfill:content-types`, dry run by default, `-- --apply` to write. A command someone runs, not a step that fires on deploy.

Against a corpus restored to its pre-#41 state: **40 items, 35 stories, 102ms in one transaction**, 152ms total, 60 adapter-declared items left untouched. Story badges went from 4 distinct types to 8:

| | NEWS | RESEARCH | PAPER | RELEASE | MODEL | BUSINESS | REGULATION | TOOL |
|---|---|---|---|---|---|---|---|---|
| before | 279 | 114 | 60 | 50 | — | — | — | — |
| after | 255 | 105 | 60 | 49 | 15 | 9 | 7 | 3 |

Retype, refresh and re-score happen in **one transaction**, so no interruption can leave a story wearing a badge its score was not computed from. Only touched stories are re-scored, because a story's score does not depend on another story's type. Story types are recomputed by `refreshStory` rather than written directly, so the backfill cannot disagree with the pipeline about how a story gets its type.

One detail worth a sentence: a reclassified item whose story stayed NEWS, because that item was not the story's primary item. That is `refreshStory` taking the type from the primary item, visibly behaving as designed.

## Merge note: 0002 was edited in place, and the tooling will not tell you

**If you pulled this branch at `75bf28e` and ran `npm run db:migrate`, your database has the wrong provenance data and a re-run will not fix it.** Nobody else is affected; 0002 has never been on `main`, which is why editing it in place is legitimate rather than a history rewrite.

Why the tooling stays silent: drizzle selects migrations with `Number(lastDbMigration.created_at) < migration.folderMillis` (`pg-core/dialect.js:62`) — strictly greater. `drizzle/meta/_journal.json` still carries 0002 at `when: 1789609526195`, unchanged, while the file's contents changed completely. So `1789609526195 < 1789609526195` is false, nothing runs, and the command reports success. The hash is computed and stored (line 67) but **never compared**, so it cannot raise the alarm — though it does make the situation *detectable* after the fact:

```sql
select hash from drizzle.__drizzle_migrations where created_at = 1789609526195;
-- 3529919e…  the version on this branch now
-- b2b329f8…  the intermediate version; you are affected
```

**Repairing an affected database is not "bump the journal and re-run."** The second statement is `WHERE content_type_source <> 'adapter'`, and the rows the intermediate version got wrong are exactly the ones it wrongly marked `adapter` — so a re-run skips precisely the rows needing repair. Reset the column first, then apply:

```sql
UPDATE raw_items SET content_type_source = 'default';
```

Or, for a throwaway database, drop and recreate.

### This PR's own verification was run on a clean database

Stated because the question is fair and the answer has to be evidence rather than recollection. The proof ran against `ai_radar_mig`, a clone of a development database that has **never applied 0002 in either version** — confirmed from both databases' migration tables, which show only entries 1 and 2 in the source. After migrating, the clone carries the third entry with hash `3529919e`, the current file.

Independently of the hash: the intermediate version contained a single `UPDATE` that could only ever write `'adapter'` — it had no statement capable of producing `'classifier'` at all. The observed post-migration state was 416 `default`, 60 `adapter`, **40 `classifier`**. Only the statements on this branch can produce that row.

## Review findings

**1 — provenance.** Fixed by recording it. `raw_items.content_type_source` is written by `decideContentType`; the migration backfills it with the old inference exactly once, at the only moment that inference is still correct, and says so. Verified against the reviewer's own scenario: with `acquire` added to `REGULATION_WORDS`, the backfill now reports `reclassified: 1  BUSINESS -> REGULATION` where it previously reported 0. Re-running after an apply reports `adapter-declared: 60` both times — the 1 → 3 drift is gone — and reverting the rule reverses the change, so the script is idempotent with respect to the current rules.

**2 — controls measured on a partial suite.** Correct, and the cause was mine: I ran controls with `npx vitest run src/pipeline/normalize/`, which is a subset, not a skip. Every control below is re-measured on the **whole suite with a database, 318 tests, 0 skipped**. The comment at `content-type.test.ts` that claimed "nothing else in the file" is corrected in place, with a note about why a control measured on a subset understates its blast radius.

**3 — re-rank outside the transaction.** Fixed; see Backfill above.

**4 — `getSql().end()` not in a `finally`.** Fixed. An error used to leave the postgres handle open and the process hanging rather than exiting.

**5 — `matchesPhrase` keeping `.`, `+`, `-`.** Taken, not deferred. It was recall-only but real: "regulation" was not found inside "AI-regulation", and several feed titles end in a full stop. Re-measured afterwards — the classifier produces the same 40 items with the same distribution, so accuracy is unchanged.

**6 — the plain-words opening described a badge nothing renders.** True at the reviewed head `d2b744f`; no longer true. This branch is now merged with 15 commits of `origin/main`, which bring `src/components/story/story-card.tsx` rendering `<ContentTypeBadge>` and `src/lib/api/labels.ts` labelling all ten types. The sentence stands as written.

That merge also made two comments in `src/lib/api/fixtures.ts` false — both said the product could not reach BUSINESS or REGULATION, which is precisely what this PR changes. Corrected, and the Bloomberg fixture moves to BUSINESS because the classifier genuinely produces BUSINESS from its own title on "valuation". Same stale-artefact family as finding 6, found by looking for it.

## For the UI lane: which badges this makes real

After this merges the pipeline can emit **MODEL, TOOL, BUSINESS and REGULATION** ("Policy") from live data for the first time. Three facts, so nobody has to go looking:

- **A blank badge is not reachable.** `ContentTypeBadge` has no per-type branch — one uniform `<span>` — and `CONTENT_TYPE_LABELS` is a total `Record<ContentType, string>`, so the compiler requires a label for every enum value and the label test derives its list from `CONTENT_TYPES`. There is no fall-through to style or to fill.
- **No new width record.** "DISCUSSION" (10 characters uppercase) is the longest of all ten labels and is already drawn by a fixture. The widest thing this PR makes reachable is "BUSINESS" at 8. Whatever the header does at 10 characters, it already does.
- **Drawn by no fixture today: MODEL, TOOL and REGULATION.** Fixtures currently cover RELEASE ×2, PAPER, NEWS, DISCUSSION and — newly, from this PR — BUSINESS. So those three labels have been exercised only by the label unit test, never rendered in a screen. RESEARCH is in the same position and is not this PR's doing: four seeded lab blogs already default to it.

## A shared helper with three callers and no tests

`keywords.ts` had no tests of its own. Not thin ones — none. It was exercised
only through two of its three callers, and neither asserts anything about how
it treats punctuation. **That absence is the defect.** What follows is only how
it surfaced.

Finding 5 of the review was that the matcher preserved `.`, `+` and `-`, and
that keeping them could only cost recall. That is true of `matchesPhrase`, which
the classifier uses. Applied to `matchesKeyword` — which is what "fix the
matcher" means to someone tidying, and what sweeping the family would have
argued for — it breaks the Hacker News AI gate silently. Measured with a
control that applies the change and names the casualties:

```
a.i.
text-to-video
text-to-image
```

Three of the forty AI keywords, unmatchable forever, with no existing test
failing. `"a.i."` is four characters and matches through the whole-word branch
against a haystack that still has its dots; the hyphenated pair breaks the same
way. Text-to-image and text-to-video tools are precisely the Show HN material
the product is being pointed at, so the version of this that shipped would have
cut AI recognition on the items the new direction is about.

Three callers, three incompatible requirements, and no two of them explain the
third:

| caller | needs |
|---|---|
| topic tagging | substrings — "machine learning" must match inside a sentence |
| content classification | boundaries — "stake" must not match inside "Mistake" |
| the Hacker News AI gate | punctuation kept — `"a.i."`, `"text-to-image"` |

`keywords.test.ts` now pins the divergence in **both** directions, so unifying
the two functions either way reddens it. The thing that would make someone merge
them is that they look like a copy-paste and nothing in the code said otherwise.

The load-bearing assertion is general rather than anecdotal: *every keyword in
the gate's vocabulary must be findable by the gate's matcher.* It survives
changes to the list, and the control fails it on all three keywords at once.
Under it sits a floor — at least one keyword must contain punctuation — so that
deleting `"a.i."` cannot leave a guard that passes while guarding nothing.

**"Did this fix sweep its family?" and "does this fix belong to the other
callers?" are different questions, and the second has to be answered first.**

## Deletion controls

Re-measured on the full suite with a database, one at a time, tree restored between each. "Named test" is the one the control exists to prove.

| Control | Named test reddens | Total red | Files |
|---|---|---|---|
| A — PAPER in `CLASSIFIABLE` + a rule matching any title | `never moves an item across a content family` | **12** | 3 |
| A-narrow — the same, but PAPER emitted for one corpus title | same | 2 | 1 |
| B — a PAPER default may be overruled | same, plus `leaves an adapter-declared fact alone` | 2 | 1 |
| C — `stake` and `enterprise` restored | `does not classify BUSINESS on a near miss` | 1 | 1 |
| D — MODEL position/verb guard removed | `does not classify MODEL on a near miss` | 1 | 1 |
| E — adapter-declared no longer wins | `prefers what the adapter declared over the title` | 4 | 2 |
| F — classifier not called | `prefers the title over the source default` | 4 | 3 |

Control A's blunt form reddens twelve because an always-matching rule also breaks every test expecting a different classification; the narrow form isolates the invariant to two. Control F reddens `run.test.ts`'s takeover test, which is the test this PR modifies — on the subset instrument it looked like one, alone.

Floors on counts that could be empty: the family invariant asserts it checked `CONTENT_TYPES.length × CORPUS_TITLES.length` pairs; the coverage test asserts all 18 seeds are present before concluding every enum value is accounted for. Every title in the tests is a real one from the live ingest.

## One existing test changed

`lets a blog post arriving after the forum thread take over the story` expected `RELEASE`; it now expects `MODEL`, because the primary item's type is classified from its own title and "Introducing GPT-6" is a model launch. The assertion's subject is unchanged — that the type follows the item that is now primary — and it now says so explicitly by additionally asserting the type is no longer the forum item's.

## Verification

All with `DATABASE_URL` set, exit codes read:

- `npm run test` — **325 passed, 25 files, 0 skipped**, exit 0
- `npx tsc --noEmit` — exit 0
- `npm run lint` — exit 0 (two warnings, both pre-existing and outside this PR's files)
- `npx prettier --check .` — exit 0
- `npm run build` — exit 0
- `npm run routes:check` — exit 0, 12 public routes

One full-suite run during this work reported 7 failed files and 82 skipped. It did not reproduce on three subsequent runs, `pg_stat_activity` showed 6 connections, and the likely cause is connection exhaustion from running six controls back to back. Reporting it because it happened, not because I can explain it.


zarghammir:

Reviewed at head `d2b744f`, in a clean worktree of my own, against `origin/main` at `a2d6a6a` as of 2026-09-17T01:31Z. The classifier itself is the strongest part of this branch and I could not break it: the content-family invariant holds, the rules are measured rather than imagined, and every test title is a real one. The changes I am asking for are in the backfill script and in claims this PR makes about its own measurements — including two of the six deletion-control rows, which I believe were measured with a third of the suite silently absent.

---

## 1. The backfill's central premise is falsified by this very PR — and I ran it twice to show it

`scripts/reclassify-content-types.ts:36-38` decides an item was adapter-declared with `stored !== sourceDefault`, justified by the comment above it: before this change `normalizeItem` wrote `item.contentType ?? defaultContentType` "and nothing else", so a difference can only have come from an adapter.

That is true of the past and false from the moment this merges. From then on a stored type differs from its source default because **this classifier** moved it.

Demonstrated on a 5-item corpus in a scratch database, real titles, nothing else touched:

| | run 1 (dry, then apply) | run 2 (dry, immediately after) |
|---|---|---|
| `adapter-declared: N left untouched` | **1** | **3** |
| `reclassified` | 2 | 0 |

The two extra are exactly the two items run 1 itself reclassified (`RESEARCH -> MODEL`, `NEWS -> BUSINESS`). No adapter ever touched either one. The script files its own previous output under a label that names a different cause.

Then I made it bite rather than leaving it as arithmetic. I added `acquire` to `REGULATION_WORDS`, so the classifier now answers `REGULATION` for "NVIDIA to Acquire Hugging Face" (printed from the module to be sure). The stored value is `BUSINESS`, so the correct answer has changed. The backfill's verdict after that rule change:

```
adapter-declared:   3 left untouched
reclassified:       0 (0.0%)
```

**The tool whose entire purpose is to re-apply the classifier to stored items cannot re-apply a changed classifier to any item it has ever moved.** Given that the PR itself documents a rejected rule and argues the next person will reach for these rules again, that is the expected second use of this script.

There is a second path to the same wrong answer: `src/db/seed.ts` upserts `defaultContentType` (it is in the `onConflictDoUpdate` `set:`), so editing a source's default in the catalogue retroactively makes every stored item from that source look adapter-declared.

The fix is to record *why* a type was set rather than infer it from a comparison — a provenance column, or a stored marker for classifier-derived types. Whatever shape it takes, `stored !== sourceDefault` cannot carry this meaning once anything but an adapter can move a type.

## 2. Two of the six deletion-control rows were measured with 88 tests skipping

Control F is listed as "RED, alone". With `DATABASE_URL` set it reddens **two** tests:

- `prefers the title over the source default` (the named one)
- `lets a blog post arriving after the forum thread take over the story` in `src/pipeline/run.test.ts` — the very test this PR modified

I then ran control F again with `DATABASE_URL` unset. 6 files and **88 of 274 tests** skip, and it reports exactly **one** red — the claim in the table. So the "alone" claims were taken on an instrument that was silently omitting a third of the suite. To be fair to the repo, it is not silent about it: `run.test.ts:31-39` prints a loud warning. But the run still exits 0, so the claim survived the warning.

Control A is affected the same way, and separately overstated. `src/pipeline/normalize/content-type.test.ts:157` says the PAPER control "reddens THIS test, named, and nothing else in the file". Implemented exactly as that comment prescribes — PAPER added to `CLASSIFIABLE`, plus a rule matching any title — it compiles clean (`tsc` 0 new errors) and reddens **seven tests across three files**:

```
content-type.test.ts  never moves an item across a content family   <- the named one
content-type.test.ts  covers every classifiable type
normalize.test.ts     takes the content type from the source when the adapter does not set one
normalize.test.ts     builds a row with canonical url, excerpt and fingerprint
normalize.test.ts     falls back to the source default when the title says nothing
run.test.ts           does not let a paper and a news report cluster on title alone
run.test.ts           makes a paper first seen through a forum link a paper story
```

**The invariant is sound and the control is real — I want that on the record.** The guard reddens, it reddens for the right reason, and it is not a type error. It is the *description* that is wrong, and a false description of a load-bearing control is how the next person deletes the control. Please correct the comment and the table rather than the code. Controls B, C, D and E reproduce your table exactly.

## 3. The re-rank sits outside the transaction, so the half-applied state the comment forbids is reachable

`scripts/reclassify-content-types.ts:113-129`: items and `refreshStory` run inside `db.transaction`, then `rankAllStories` runs after it commits. The comment at line 125 says leaving scores alone would be "a half-applied migration where labels move and ordering does not" — which is exactly what an interruption in that window produces.

I made the process die between the commit and the re-rank:

```
after the crash:  item = MODEL   story = MODEL   score = 11.5  components {"contentType": 0}
after a real re-rank: score = 38.1  components {"recency": 12.1, "contentType": 6, "primarySource": 20}
```

The story wears the new badge and carries a score computed without it.

## 4. Any error leaves the process hung, because `getSql().end()` is not in a `finally`

`scripts/reclassify-content-types.ts:146,149-152`: the connection is closed only on the success path. `main().catch` logs and sets `process.exitCode = 1`, but the open postgres handle keeps node alive. I did not find this by reading it — my crash simulation above never exited and had to be killed after 300 seconds. On a failed backfill the operator sees a stack trace and a command that never returns.

## 5. Minor: `matchesPhrase` keeps `.`, `+` and `-`, which can only cost recall

`src/pipeline/normalize/keywords.ts:39-47` strips punctuation to spaces but preserves `.`, `+` and `-`, then requires a space on both sides. None of the five vocabularies contain those characters, so keeping them cannot help a match and does prevent some: "regulation" is not found inside "AI-regulation", and any phrase followed by a full stop never matches. Recall-only, and this PR argues precision is worth roughly ten times recall, so I raise it as a note rather than a change I need.

## 6. Minor: the plain-words opening describes a badge nothing renders on this branch

"Every story on the brief carries a badge" — no file under `src/components` or `src/app` references `contentType` at all. On this branch the classifier's effect is on ranking and on the API payload (`src/api/stories.ts:201`). The effect is real; the sentence describes a surface that is not here yet, and the owner reads these descriptions.

---

## Verified myself

- Head SHA `d2b744f`; base currency: forks from `97f26ef`, **6 commits behind `origin/main` (`a2d6a6a`) as of 2026-09-17T01:31Z**. Main moved during my fetch. The six commits touch none of these nine files, so I *expect* a clean merge — stated as a prediction, not a fact.
- File count two independent ways, agreeing: `gh pr diff --name-only` = 9, `git diff --name-only 97f26ef..d2b744f` = 9, sorted lists identical.
- CI at the exact head SHA (run `35169906818`): 3 check-runs, all `completed`/`success`, and I read **every step of all three jobs**. Every step executed. The only non-success steps are the two `if: failure()` log dumps, correctly `skipped`. No empty conclusions.
- Locally, with a real database and nothing skipping: `npm test` **274 passed / 21 files / 0 skipped, exit 0**; `npx tsc --noEmit` exit 0, **0 errors**; `npm run lint` exit 0 (the one warning is in `rank-all.ts`, which is not among this PR's nine files); `npm run format:check` exit 0; `npm run build` exit 0; `npm run routes:check` exit 0, 12 public routes.
- All six deletion controls, one at a time, each reverted to a clean `git status`, each gated on the `tsc` error set against a zero-error baseline: **no control introduced a type error**, so no control failed to compile.
- Findings 1, 3 and 4 reproduced on the machine as described above.
- Adjacent-surface sweep: one `contentFamily`, one PAPER test, `normalizeItem` the only writer of an item's type, `arxiv` and `hackernews` the only declarers. No second surface re-expresses the boundary.
- The product rules I can check statically hold: no new route, no paid call, no key outside env, no reading analytics, nothing in `src/config/brand.ts`.

## Taken on trust

- **The 516-item live corpus and every number derived from it** — 90% precision, the 40/36 table, recall ~1/3, the 15-found/8-wrong generic rule, the rank 104 -> 14 movement and ranks 124/127. I did not re-ingest 18 live sources. My corpus work used 5 items I inserted myself.
- **The backfill's reported production figures** (40 items, 35 stories, 328ms, the before/after story-type table).
- **That the two named stories are genuinely misclassified** — an editorial judgement, and I agree with it on reading, which is not the same as verifying it.
- **My local toolchain is not CI's**: `.nvmrc` pins Node 22 and no Node 22 is installed on this machine, so every local number above was taken on Node 25.8.1 / npm 11.11.0. My results agreed with CI at the same SHA, so I have no reason to suspect my instrument — but the difference is real and I am naming it.
- Whether the six commits now on main merge cleanly with this branch.

zarghammir:
All six taken. Head is now `75bf28e`, CI run `35172067020` green on that exact SHA, all three jobs read individually.

**The most useful thing this review did was not finding 1 itself — it was that chasing finding 1 showed the same defect had corrupted this PR's own headline measurement.**

To decide which stories the classifier had retyped, my ranking analysis used `stored !== sourceDefault`: the identical inference you showed cannot carry that meaning. It undercounted badly — 9 retyped stories where the true figure is 34. So my claim that "the Model weight's entire visible effect at the top of the brief was to promote two misclassified stories" was an artefact of a broken instrument, and I have withdrawn it in the body. Re-measured against recorded provenance, ten genuine model launches moved from ranks ~100–124 into ranks 20–30, and the largest single move into the top ten is still a false positive at 81 → 8. The precision argument survives and is now properly grounded; the conclusion I drew from it did not.

I would not have found that by fixing the script alone. Thank you for making it bite with `acquire` rather than leaving it as arithmetic — that is what sent me back to the other place I had used the same reasoning.

**1 — provenance.** `raw_items.content_type_source` (`adapter | classifier | default`), written by a new `decideContentType` that owns the precedence rule in one place. The migration backfills it with the old inference exactly once, at the only moment that inference is still correct, and the SQL comment says why it can never run again. Verified against your scenario:

```
with "acquire" in REGULATION_WORDS:   reclassified: 1   BUSINESS -> REGULATION
apply, then dry re-run:               adapter-declared: 60 both times, reclassified: 0
revert the rule, dry run:             reclassified: 1   REGULATION -> BUSINESS
```

The 1 → 3 drift is gone and the script is idempotent with respect to whatever the rules currently say. Your second path is closed by the same change: `seed.ts` upserting `defaultContentType` no longer affects provenance, because provenance is not derived from the default any more.

**2 — controls on a partial suite.** You were right and the cause was worse than skipping: I ran them with `npx vitest run src/pipeline/normalize/`, so a third of the suite was never *selected*, let alone skipped. All seven rows are re-measured on the whole suite with a database, 318 tests, 0 skipped. Control A reddens **12** across three files as you said, not "nothing else in the file"; I have also recorded a narrow variant that isolates the invariant to 2, so the table shows both the blunt control the comment prescribed and the sharp one. The comment is corrected in place and now explains that a control measured on a subset understates its blast radius — which is the actual lesson.

**3 and 4.** Both fixed. Retype, refresh and re-score are one transaction now, and only touched stories are re-scored since a story's score does not depend on another story's type. `getSql().end()` is in a `finally`, with a comment naming the hang rather than calling it tidiness.

**5.** Taken rather than deferred — it was recall-only but real, and I re-measured afterwards: same 40 items, same distribution, so accuracy is unchanged.

**6.** True at `d2b744f`, no longer true. I merged 15 commits of `origin/main` (I was 15 behind, not 6 — main moved again between your fetch and mine), and they bring `story-card.tsx` rendering `<ContentTypeBadge>` and `labels.ts` labelling all ten types, so the sentence now describes something that exists.

That merge also falsified two comments in `src/lib/api/fixtures.ts` which said the product could not reach BUSINESS or REGULATION — exactly what this PR changes. Both corrected, and the Bloomberg fixture moves to BUSINESS because the classifier genuinely produces it from that headline on "valuation". Went looking for those because of your finding 6; they are the same family.

**On your "taken on trust" list** — all fair. Two notes. The corpus is re-ingested and slightly different (515 items / 503 stories), and every number in the body is re-derived from it rather than carried over. And the two stories I called misclassified are still an editorial judgement; I added a third to the wrong column while writing this up ("Training a 4B model to produce 81% faster query plans" is a technique write-up, not a release), which is what makes 36 of 40 add up.

One thing I could not reproduce and am reporting anyway: a single full-suite run mid-work reported 7 failed files and 82 skipped. Three subsequent runs were clean, `pg_stat_activity` showed 6 connections, and the likely cause is connection exhaustion from six back-to-back control runs. It is in the body.

zarghammir:

Re-review at head `75bf28e`, new comment rather than an edit of the last one. Base currency: the branch now carries `origin/main` at `1c0e5ac` — **0 commits behind** as of 2026-09-17T01:55Z, and GitHub reports `MERGEABLE/CLEAN`.

Everything I asked for in the previous review is fixed, and I re-proved each fix on the machine rather than reading the diff and nodding. The provenance column is the right shape and `decideContentType` is the right place for it. I am requesting changes for one reason only, and it is a new finding that the fix itself created: **the migration writes the old bug into the data, and the owner's own development database is already in the state where it does.**

---

## BLOCKING — the migration's "one moment" has already passed on this machine

`drizzle/0002_slimy_fallen_one.sql:3-16` backfills provenance with the old inference, justified by: "This is the ONE moment at which `content_type <> default_content_type` correctly means an adapter declared this: every row here was written by a normalizeItem that could only copy the adapter's value or the source default."

That sentence is true of rows written by `normalizeItem`. It is false for rows written by **the previous head of this very branch**, whose backfill wrote `contentType` straight onto the row with `tx.update(rawItems).set(...)`, bypassing `normalizeItem` entirely. The PR body records that it was run for real against the 516-item corpus.

**I checked the development database on this machine, read-only, and it is in exactly that state:**

```
ai_radar:  516 raw_items, no content_type_source column yet
           40 rows where content_type <> default_content_type
           of those 40, PAPER or DISCUSSION (the only two types an adapter declares): 0
           breakdown: MODEL 20, BUSINESS 9, REGULATION 8, TOOL 3
```

That breakdown is the previous head's backfill table, exactly. So when `npm run db:migrate` is run against it, the migration will mark **40 rows `adapter` of which zero are adapter-declared**, and it will do so permanently — provenance is now stored rather than inferred, so there is nothing left to recover it from.

Reproduced end to end in a scratch database, pre-migration, mirroring that shape:

| row | stored | truth | migration recorded | |
|---|---|---|---|---|
| m1 google-deepmind | MODEL | classifier (old backfill) | `adapter` | **wrong** |
| m2 verge-ai | BUSINESS | classifier (old backfill) | `adapter` | **wrong** |
| m3 google-deepmind | RESEARCH | default | `default` | ok |
| m4 arxiv-ai | PAPER | adapter | `default` | **wrong** |
| m5 hackernews-ai | DISCUSSION | adapter | `adapter` | ok |

Three of five wrong, in both directions. Then the consequence, which is the original finding restored through the data:

```
$ npm run backfill:content-types
adapter-declared:   3 left untouched
reclassified:       0 (0.0%)

# add "acquire" to REGULATION_WORDS — the classifier now answers REGULATION
# for m2's title, and m2 is stored as BUSINESS:
$ npm run backfill:content-types
adapter-declared:   3 left untouched
reclassified:       0 (0.0%)
```

The code path is fixed and the data path is not. **The SQL comment guards against the statement running a second time; the damage is done by it running the first time, on a database that has already been backfilled.** Anyone reading that SQL and nodding would not see it, because the claim it makes is about `normalizeItem` and the thing that broke it was not `normalizeItem`.

The other direction matters less but is the same inference failing: all 60 arXiv items in `ai_radar` are genuinely adapter-declared and will be recorded `default`, because arXiv's declaration equals its source default. Harmless today — `isDeclaredFact` keeps PAPER from being reclassified regardless — but it is now frozen into a column that is meant to be the source of truth.

**What would fix it:** stop guessing in SQL. The titles are stored and the classifier is deterministic, so provenance can be *derived* rather than inferred from a comparison — a row whose stored type is PAPER or DISCUSSION and differs from its default is `adapter`; a row the classifier reproduces from its own title is `classifier`; everything else is `default`. That is a TypeScript step, not a SQL one, and it is correct on a clean database and on an already-backfilled one alike. Whatever shape you choose, it has to handle the already-backfilled case, because that is the case the owner has.

## Minor — control A's "12" is not reproducible from its description

The table says control A reddens 12 across 3 files. I could not get 12 from either natural reading of "PAPER in `CLASSIFIABLE` + a rule matching any title":

- rule **appended last** in `RULES`: **9 failed**, 3 files
- rule **placed first** in `RULES`: **16 failed**, 3 files

Both compile (`tsc` 0 new errors) and in both the invariant test is among the reds, so the guard is real either way. The gap is that the description does not say where the injected rule goes, and the count swings by seven depending on the answer. Either pin the position in the comment, or record it as a range. This is the same class of defect as the one just fixed, one notch smaller.

**Every other row reproduces exactly**, measured on the whole suite with a database, one at a time, tree restored between each, each gated on the `tsc` error set against a zero-error baseline:

| Control | body says | I measured |
|---|---|---|
| A-narrow | 2 / 1 file | **2 / 1** |
| B | 2 / 1 | **2 / 1** |
| C | 1 / 1 | **1 / 1** |
| D | 1 / 1 | **1 / 1** |
| E | 4 / 2 | **4 / 2** |
| F | 4 / 3 | **4 / 3** |

No control failed to compile, so none of them is a type error wearing a control's clothes.

## Minor — `decideContentType` records `default` for a genuine classifier hit if a source ever defaults to a classifiable type

`src/pipeline/normalize/content-type.ts`: the decision is `classified === sourceDefault ? "default" : "classifier"`. Today no seeded source defaults to MODEL, TOOL, BUSINESS or REGULATION, so this cannot fire. Add one that does, and a rule firing correctly on the title is recorded as `default` — which then lets the backfill overwrite it. A note in the function, or keying off whether a rule matched rather than off value equality, would close it.

## The cross-lane edit: required, not smuggled — and I checked rather than assumed

`src/lib/api/fixtures.ts` is the UI lane's file. My ruling:

- **Both comment corrections are required.** They asserted that the product "cannot currently reach" BUSINESS and that "nothing can emit REGULATION today". This PR is the merge that falsifies both. A comment asserting a now-false property is not housekeeping, it is a defect this PR creates if it leaves it.
- **The Bloomberg value move NEWS to BUSINESS is a behaviour change in another lane's file, and it is correct.** I ran the real classifier over the actual fixture titles rather than trusting either comment: "OpenAI weighs funding at a $1.2T–$1.5T valuation" gives `BUSINESS/classifier`, and "China's Foreign Ministry rejects Amodei's call to curb its AI" gives `NEWS/default`. Both new comments are accurate. `src/lib/api/labels.ts` has "Business" and renders REGULATION as "Policy", so nothing renders blank.
- Nothing pins it: `fixtures.test.ts:56` only asserts more than two distinct types across the fixtures, so no test forced this and none would have caught it.

I am **not** asking for it to be split out. The file's own governing rule — set by the UI lane's own commit "two fixtures described states the product cannot produce" — is that a fixture must describe a state the product can produce, and after this merge a story with that headline is produced as BUSINESS. Leaving it NEWS would have left the fixture inconsistent with the pipeline under the UI lane's own principle. I am naming it as a behaviour change rather than letting it pass as a comment fix, and it should carry that lane's acknowledgement.

## Fixes from the last review, each re-proved rather than read

- **Provenance recorded, not inferred** — `decideContentType` owns the whole precedence rule and returns `{type, source}`; the backfill branches on `item.storedSource === "adapter"`. Verified by the three new unit cases and by control E/F reddening as expected.
- **Re-rank moved inside the transaction** — proved by making `rankOneInTransaction` throw mid-transaction: item stayed `RESEARCH/default`, story stayed `RESEARCH` at score 11.5, nothing half-applied. Before, the same experiment left a MODEL badge on a score computed without it.
- **`getSql().end()` in a `finally`** — proved by making the script throw: **exit 1 after 1 second**. The same experiment against the previous head never returned and I had to kill it at 300s.
- **The control comment** at `content-type.test.ts:157` now states the real blast radius and says plainly that the earlier claim came from running one directory. That is the right correction and it is better than what I asked for.

## Verified myself

- Head `75bf28e`; 0 commits behind `origin/main` (`1c0e5ac`) at 2026-09-17T01:55Z; `MERGEABLE/CLEAN`.
- Files two ways, agreeing: `gh pr diff --name-only` = 15, `git diff --name-only 1c0e5ac..75bf28e` = 15, sorted lists identical.
- CI at the exact head SHA: 4 check-runs, all `completed`/`success` (run `35172067020` plus a Vercel preview check). I read every step of all three GitHub jobs; every step executed, the only non-success steps being the two `if: failure()` log dumps, correctly `skipped`.
- Local, real database, nothing skipping: `npm test` **318 passed / 24 files / 0 skipped, exit 0** · `tsc --noEmit` exit 0, 0 errors · `lint` exit 0 · `format:check` exit 0 · `build` exit 0 · `routes:check` exit 0, 12 public routes.
- Lint reports **2 warnings, both pre-existing on `1c0e5ac`** and neither introduced here: `ScoreComponents` in `rank-all.ts`, and an unused `bloomberg` const at `fixtures.ts:41` — this PR's only hunks in that file are at 208 and 244, so it did not cause it. Flagging it because the count went 1 to 2 since my last review and that is the merge of main, not this work.
- The `ai_radar` inspection above was strictly read-only: `SELECT` only, no writes, no schema change. All of my own writes were in a scratch database, which I found empty and left empty.

## Taken on trust

- **The withdrawn headline and its replacement.** The project has withdrawn the earlier claim that the MODEL weight's only visible effect was promoting two misclassifications, and re-measured it as 9 undercounted against 34 real, with ten genuine launches moving into ranks 20-30 and the largest move into the top ten still a false positive at 81 to 8. I have verified **none** of those numbers — I did not re-ingest 18 live sources. I record that the correction was found by the lane chasing the provenance fix, not by me.
- The 90% precision figure, the 40/36 accuracy table, and the rejected generic rule's 15-found/8-wrong.
- The disclosed flake: one full-suite run reporting 7 failed files and 82 skipped, not reproducing across three later runs, with connection exhaustion the suspected cause. I treat that as **suspected, not established**. My own runs showed no such signature and `pg_stat_activity` was at 0 connections to the review databases after my suite. Disclosing it was the right call and worth more than a clean-looking body.
- **My toolchain is still not CI's**: `.nvmrc` pins Node 22, none is installed here, so every local number above is Node 25.8.1 / npm 11.11.0. My results agree with CI at this SHA.

zarghammir:
REVIEW VERDICT: APPROVE

Third pass, at head `329e2c5`, new comment rather than an edit. Base currency: 0 commits behind `origin/main` (`1c0e5ac`) as of 2026-09-17T03:07Z, `MERGEABLE/CLEAN`. Five checks at the exact head SHA, all passing — four check-runs plus a Vercel commit status.

**The blocking finding is fixed, and the fix is better than the one I proposed.** Mine needed the stored titles and a run of the classifier. This one needs neither: exactly two adapters declare a type, both declare a type `CLASSIFIABLE` cannot emit, so provenance is decidable from the stored value alone. That is a stronger argument resting on a smaller premise, and the premise is guarded by a test rather than by a comment.

I proved it rather than accepting it.

## The migration, proved on the database that convicted the old one

I reset my review database to the pre-0002 state — dropped the column and the type, removed the applied row — so it had genuinely never seen migration 0002 in either version. Then I seeded the shape of an **already-backfilled** database: four rows the previous head's backfill had moved, one untouched row at its source default, and two genuinely adapter-declared rows. Then I ran `npm run db:migrate`.

| row | stored | truth | recorded | |
|---|---|---|---|---|
| x1 google-deepmind | MODEL | classifier | `classifier` | ok |
| x2 verge-ai | BUSINESS | classifier | `classifier` | ok |
| x3 ars-technica-ai | REGULATION | classifier | `classifier` | ok |
| x4 openai-blog | TOOL | classifier | `classifier` | ok |
| x5 google-deepmind | RESEARCH | default | `default` | ok |
| x6 arxiv-ai | PAPER | adapter | `adapter` | ok |
| x7 hackernews-ai | DISCUSSION | adapter | `adapter` | ok |

**Seven of seven.** The old migration got three of five wrong on the same shape. The hash recorded afterwards is `3529919e`, the new file.

And the loop closes — the failure that started all of this no longer happens:

```
backfill, no rule change:      adapter-declared: 2    reclassified: 0
add "acquire" to REGULATION:   adapter-declared: 2    reclassified: 1
                               by transition:  BUSINESS -> REGULATION  1
```

A changed rule now reaches a row the classifier had previously moved. Under the old design that same experiment reported `reclassified: 0` twice, and `adapter-declared` counted rows no adapter had touched.

## Control A reproduces exactly, now that the position is pinned

Injected as the **second** rule — after MODEL, before REGULATION — I measure **12 failed across 3 files**, matching the table. With the same description and the rule appended last I get 9, and placed first 16. Pinning the position was the right way to close that finding; recording a range would have left the next person unable to tell which control they were running.

Every other row still reproduces: A-narrow 2/1, B 2/1, C 1/1, D 1/1, E 4/2, F 4/3. No control introduced a type error.

## The keywords work checks out exactly as described

The claim is that applying my own punctuation finding to `matchesKeyword` breaks three of the forty AI keywords. I applied it — changed line 18's character class to match `matchesPhrase` — and confirmed the mutation landed before trusting the result:

```
keywords now unfindable: a.i., text-to-video, text-to-image  (3 of 40)
reddens 2 tests, both named:
  keywords.test.ts > the AI vocabulary and the matcher that reads it > finds every keyword in a title that contains it
  keywords.test.ts > the two keyword matchers diverge on purpose > matchesKeyword keeps punctuation, so a dotted keyword still matches
```

Exactly as claimed. This is the best kind of test to add: my observation was correct about one matcher and destructive about the other, no existing test would have failed, and the new file pins the divergence in both directions so unifying them either way goes red.

---

## Findings — none blocking, all recorded

### 1. The gate drops "AI-powered X" entirely — already filed as #73

`matchesKeyword` keeps `-` in the haystack and short keywords take the whole-word branch. `DEFAULT_AI_KEYWORDS` contains `ai`, `gpt`, `llm`, `rag`, `gpu`, `xai` — all four characters or fewer. Measured:

```
"ai"  in "AI-powered code review"            -> MISSED
"gpt" in "GPT-5 is out"                      -> MISSED
"llm" in "LLM-based agents..."               -> MISSED
"rag" in "RAG-based search"                  -> MISSED
"gpu" in "GPU-accelerated inference"         -> MISSED
"ai"  in "AI powered code review"            -> found
```

At the gate, three of seven realistic Hacker News titles are dropped outright. Two of the four that pass are rescued by an unrelated keyword — "agent" and "inference" — which is luck, not coverage. A miss here is not a wrong badge, the story never enters the brief.

**Pre-existing on `main` and untouched by this PR, so it does not block it.** I raise it here because this PR is the first to test that file and reasons explicitly about punctuation inside it — that makes it the right place to say so, not the wrong one. And the fix is **not** to strip punctuation: this PR has just proved that destroys `a.i.`, `text-to-video` and `text-to-image`. The root cause is that `keywords.ts` answers *word boundary* and *character class* with the same `replace`. A hyphen needs to be a boundary for the short-token branch while dotted keywords stay intact.

### 2. The new keyword property test tests only the easy case

`keywords.test.ts:56`, "finds every keyword in a title that contains it", builds `Show HN: a ${kw} project I built` — every keyword placed between two spaces, which is the one context where the whole-word branch cannot fail. The name claims a title that contains the keyword; the body tests the space-delimited case only, and the hyphenated case is both the realistic one and the broken one. A test written today that already promises more than it performs is worth more attention than a defect from months ago.

### 3. The migration's guard is narrower than the claim it makes

`content-type.test.ts`, "no adapter declares a content type outside PAPER and DISCUSSION", globs `*adapter.ts`. Its docstring says it fails "if an adapter ever declares a type outside that pair" — but a declaration anywhere else is invisible to it. Demonstrated: I planted `contentType: "NEWS"` in a new `src/sources/rss/feed-type.ts`, and the guard passed. The floor of 2 cannot catch it either, because the two known declarations still satisfy it.

Today the glob is correct — the only two declarations live in `arxiv/adapter.ts` and `hackernews/adapter.ts`, and `rss/parse.ts` and `http.ts` declare nothing. So this is latent, and the exposure is narrow: a fresh install migrates an empty table, so the risk is only a database holding rows written by a third declaration added before it was migrated. Widening the glob to all of `src/sources/**/*.ts` excluding tests is a one-line change and would make the test match its own sentence. I would have blocked on it if the migration could still go wrong on the author's database; it cannot, so I have not.

### 4. Anyone who migrated at `75bf28e` is silently stuck — a merge note, not a defect

Migration 0002 was edited in place, which is correct practice for an unmerged migration. But I read the installed library rather than trusting memory: `node_modules/drizzle-orm/pg-core/dialect.js:62` selects with `Number(lastDbMigration.created_at) < migration.folderMillis`, strictly greater, and the hash it stores on line 67 is never read back. The journal entry still says `when: 1789609526195` while the file's contents changed completely, so a database that applied the intermediate version will never receive the new one. No error, no warning, and `db:migrate` reports success.

**The author's database is safe and I checked rather than assumed** — read-only on `ai_radar`: the column is absent, two migrations recorded, the last at `1789589594536`, which is 0001. It has never seen 0002 in either version, and when it is migrated it will get 60 rows to `adapter` and 40 to `classifier`.

Worth knowing for anyone who pulled this branch and migrated mid-review: **re-running does not repair it**, because the second statement is `WHERE content_type_source <> 'adapter'` and excludes exactly the rows the old one got wrong. Reset the column to `'default'` first, or recreate the development database.

### 5. Minor, still open from the last pass

`decideContentType` records `default` whenever the classified answer equals the source default. No seeded source defaults to a classifiable type, so it cannot fire today; add one that does and a rule firing correctly is recorded as `default`, which then lets the backfill overwrite it.

---

## Verified myself

- Head `329e2c5`, 0 commits behind `origin/main`, `MERGEABLE/CLEAN`, files 16 two ways with identical sorted lists.
- Five checks at the exact head SHA, all pass. Reading only the check-runs API shows four — Vercel posts a **commit status**, not a check-run. I had this wrong on my previous pass and have corrected it.
- Local, real database, nothing skipping: `npm test` **325 passed / 25 files / 0 skipped, exit 0** · `tsc --noEmit` exit 0, 0 errors · `lint` exit 0 · `format:check` exit 0 · `build` exit 0 · `routes:check` exit 0.
- Lint's 2 warnings are both pre-existing on `1c0e5ac` and neither is introduced here.
- The migration table above, the backfill rule-change result, control A at 12/3, the six other control rows, the keywords control at 3-of-40 and 2 named reds, and the planted-declaration guard demonstration.
- All writes were in a scratch database. The `ai_radar` inspections were `SELECT` only.

## Taken on trust

- **The withdrawn ranking headline and its replacement** — 9 undercounted against 34 real, ten launches moving into ranks 20-30, the largest move into the top ten still a false positive at 81 to 8. I verified none of it; I did not re-ingest 18 live sources. The project withdrew that claim itself, found while chasing the provenance fix rather than by me.
- The 90% precision figure and the 40/36 accuracy table.
- The lane's `ai_radar_mig` proof. I did not need to take it on trust in the end — my own reset-and-migrate reproduces the same result independently — but its second argument is the stronger one and stands on its own: 40 rows came out reading `classifier` when the intermediate file contained no statement capable of writing that value.
- **Contention disclosure:** at release I found a live `vitest` run belonging to another worktree on this machine, overlapping at least my final run. Every number above is a pass/fail identity rather than a timing, and a green suite taken under contention is stronger evidence rather than weaker, so I am not re-running — but it is disclosed rather than smoothed.
- **My toolchain is not CI's**: `.nvmrc` pins Node 22, none is installed here, so every local number is Node 25.8.1 / npm 11.11.0. Results agree with CI at this SHA.

PR #58, opened 2026-09-17
ci: prove the README's quick start works on a machine with nothing (#18)

zarghammir:
Part of #18 — the clone-and-run proof. The architecture overview is the remaining half and is not here.

## What this is

A workflow that follows the README's own instructions **literally**, on a runner that has never seen this repository, and asserts the README's own promises. Both documented paths: self-host with Docker, and "needed to work on the code".

The whole promise of an open-source project is that a stranger can clone it and run it. **Nobody ever had.** Every green this project has is on a tree that already had `node_modules`, a database, or a warm cache somewhere.

## The editing rule, which is in the workflow file

> It may only run commands the README gives, in the order the README gives them, and when a job here needs a step the document does not contain, the README is what changes.

A job that quietly learns our habits would pass forever and prove nothing. That rule is why the contributor path was allowed to go red rather than fixed first.

---

## Findings, labelled by how they were established

The distinction matters for what you can re-run and what you must re-derive.

### Proved by a red, with a run id — re-runnable

**Run `35170475688`**, job *work on the code, the README verbatim*, failing step `npm run db:migrate`:

```
DrizzleQueryError: Failed query: CREATE SCHEMA IF NOT EXISTS "drizzle"
  cause: AggregateError [ECONNREFUSED]
```

That single red proves three gaps in the contributor path:

1. **It never says you need a Postgres.**
2. **Nothing creates what `.env.example` expects** — a database, user and password all named `ai_radar` on `localhost:5432`.
3. **It never seeds**, so anyone who got past the first two ended up with a working app and an empty catalogue — while the Docker path, four paragraphs earlier, explicitly promises the opposite.

The `self-host` job passed in that same run, so the Docker path stayed proven while the contributor path was red.

### Proved green, also with a run id

**Run `35170125563`** — the self-host quick start works as written, on a machine that had nothing:

```
docker:  Docker version 28.0.4
http://localhost:3000 answered 200 after 81s
sources loaded by the first start: 18
```

Stated as a result rather than as an absence of findings: every instruction after the clone, in order, with nothing added, and the catalogue was there.

### From reading, not measurement — you must re-derive these

I cannot be this README's clean-room reader: I have read it, the Dockerfile, the compose file and the CI today, and I will supply a missing step from memory without noticing. These three came from reading it as carefully as I could, and the job cannot confirm any of them.

4. **The document is wrong as a whole while no single sentence in it is wrong.** The README explains that Compose deliberately does not publish the Postgres port. It also ships a contributor `DATABASE_URL` pointing at `localhost:5432`. Both statements are true; they are never joined; and the consequence — that the obvious move, using the database already running, cannot work — exists only in the space between them. **I verified this against `docker-compose.yml` rather than reasoning from the README: it publishes `3000:3000` and nothing else.** A job cannot find this class of defect, because every individual instruction it runs is correct.

5. **Docker was never named as the self-host prerequisite**, while the Node and npm requirements a self-hoster does not need sat immediately above the command that needs Docker.

6. **`docker compose up` runs in the foreground** while the next instruction says to open a browser — so a reader following it literally needs a second terminal the document never mentioned.

---

## The fixes, as their own commit

`0335763` changes the README only, so the diff shows exactly what a stranger would have hit. `0f549c1` then updates the job to follow the corrected document, because that is the rule.

**The seed is not added silently.** The document says what it is — the step the Docker path performs for you — and what skipping it leaves you with: an app that runs and has nothing to read, which looks like a broken build and is not one. The job floors the catalogue count for the same reason, so a seed that seeded nothing cannot pass as one that worked. A document that leaves someone with an empty app works, and teaches the wrong model.

---

## One deliberate substitution

`git clone` is replaced by the checkout. Cloning would fetch **main**, so a pull request fixing the quick start would be graded against the unfixed copy — the job would pass or fail on the wrong document, and would look perfectly faithful doing it.

The comment names **#59**, which carries the condition under which the literal clone becomes right: today the repository is **private**, so the README's first command cannot work for a stranger at all. When #59 flips it public, restoring `git clone` proves the first command by the same instrument that proves the rest.

## Its own workflow, and a path filter

A job-level filter still pays a billed minute to check out and decide; a workflow that does not trigger costs nothing. The filter is safe here in a way it would not be on the browser job: that one would have to guess which files can change a rendered page and the honest answer is most of them, while this enumerates the files that can break **the documented instructions**, which is a small closed set. Anything else that breaks the image is still caught by `compose-smoke` on every pull request.

**Measured cost** (run 35170125563): `clone-and-run` 86s, against `ci` 108s, `compose-smoke` 115s and `browser checks` 106s. A pull request now costs **8 billed minutes**; with this filter the common case returns to **6**.

**On the overlap with compose-smoke, since two jobs building the same image invites the question:** they largely do prove the same artifact. The unique coverage is not the artifact, it is the **command form** — `compose-smoke` builds only `web` and starts services explicitly with `--wait`, and never runs plain `docker compose up`. What this workflow has that nothing else can is that **its input is the document**: `compose-smoke` would keep passing while the README rotted.


zarghammir:

Reviewed at head `0f549c1`, two files — `.github/workflows/quick-start.yml` and `README.md`. Reviewed by reading the workflow, the README, the Dockerfile, the compose entrypoint and the tree. **No part of this needed the machine and I did not take one.** Nothing below is unverified.

Base currency: the branch is **15 commits behind `origin/main`**. The document under test has not moved, though — `README.md` is unchanged on main since this branch's merge-base (`97f26ef`), so the thing this job grades is the same thing it was written against.

This workflow is the best-reasoned CI file in the repository. Its editing rule — *"it may only run commands the README gives, in the order the README gives them… a job that quietly learns our habits would pass forever and prove nothing"* — is the right rule, and it is followed. Where it deviates it says so and says why. One thing blocks, and it is the claim in the PR's own title.

---

## BLOCKING — the "machine with nothing" is a machine with Node, and the step that would notice only prints

```yaml
- name: Show what this machine has before we start
  run: |
    echo "docker:  $(docker --version 2>/dev/null || echo 'absent')"
    echo "node:    $(node --version 2>/dev/null || echo 'absent')"
    echo "openssl: $(openssl version 2>/dev/null || echo 'absent')"
```

The step above it explains what it is for: *"No Node setup step on purpose: the README says the Node toolchain is not needed to self-host because Docker builds it all, and this is here to find out whether that is true."*

It cannot find that out. `ubuntu-latest` ships Node, so `node --version` prints a version, the step exits 0, and the job proceeds. If any part of the documented self-host path ever reached for the host's Node, this job would stay green and a genuinely Docker-only machine — which is the machine the README is addressing — would fail. The step records the hazard instead of testing it.

**This is the same defect as the floor in `scripts/verify-shell.mjs`, one layer up, and I want to name it as a pattern rather than as a second bug.** There, a correct diagnostic sat four lines below something that threw, so the answer was computed into a variable while CI reported a Playwright timeout. Here, the answer is computed into a variable and echoed to a log nobody reads on a green run. In both cases the instrument is correct, the value is present, and nothing is gated on it. Two unrelated files, the same habit: **observing a condition and reporting it, where the whole point was to fail on it.**

The fix is small and it would make the title true. The documented self-host path is `cp`, `openssl`, `sed` and `docker compose` — none of it should invoke Node on the host. Putting a failing `node` stub first on `PATH` for the duration of those steps turns the printed observation into an assertion:

```yaml
- name: Make the host's Node unavailable, since the README says it is not needed
  run: |
    mkdir -p /tmp/no-node
    printf '#!/bin/sh\necho "the self-host path used the host node; the README says it should not" >&2\nexit 1\n' > /tmp/no-node/node
    chmod +x /tmp/no-node/node
    echo "/tmp/no-node" >> "$GITHUB_PATH"
```

If you would rather not, the honest alternative is to say in the file that this job does **not** test the Node-free claim and to carry it on an issue — but as written the comment asserts it does, and that comment is the kind of thing this repo has been burned by three times tonight.

## Minor — the trigger omits a file the job itself reads

The path filter lists `README.md`, `Dockerfile`, `docker-compose.yml`, `docker/**`, `.env.example`, `package.json` and the workflow. The contributor job resolves its Node version from **`.nvmrc`**, via `node-version-file: .nvmrc`, and `.nvmrc` is not in that list. Change `.nvmrc` and the job that depends on it does not run.

**I want to be precise about how narrow this is, because my first reading of it was wider and wrong.** I also suspected `package-lock.json`, `drizzle/**` and `src/db/**` were uncovered, and they are not: the Dockerfile copies `package-lock.json` and runs `npm ci` at line 36, and `docker/web-entrypoint.sh` runs migrate and seed before the server — so compose-smoke, which asserts the sources table is non-empty, genuinely does catch breakage in all three. The comment's rationale — *"anything else that breaks the image is caught by compose-smoke"* — is substantially true. `.nvmrc` is the one documented input that the image does not read, because the Dockerfile pins `node:22-alpine` directly.

So: one missing entry, low consequence, and the reasoning around it holds up better than I expected.

## Minor — the version under test is not the version documented

The README states the floor precisely and explains it: *"Node 22.13 or newer… 22.13 is the lowest Node that every dependency accepts: `vite` and `rolldown` need 22.12, and `eslint-visitor-keys` needs 22.13."* The job resolves from `.nvmrc`, which contains `22`, so `setup-node` installs the **latest** 22.x. The stated floor is never exercised. A contributor on exactly 22.13 — the version the README tells them is enough — is the one case this job cannot speak to, and it is the case the sentence exists to support.

Not a blocker, and arguably not worth a second matrix leg for a project at this stage. Worth knowing that the number in the document and the number under test are different numbers.

---

## Credited rather than filed

- **`actions/checkout` standing in for `git clone`** is the right call and is disclosed with its reasoning: cloning would fetch `main`, so a pull request fixing the quick start would be graded against the unfixed document — passing or failing on the wrong file and looking faithful doing it. The condition under which the literal clone becomes correct is named, and #59 carries it. I would not change this.
- **`setup-node` is used without `cache: npm`.** On a workflow whose entire purpose is a cold machine, a warm cache would be precisely the thing that makes it prove nothing. Leaving it off is deliberate and correct, and easy to "helpfully" add later — worth a comment saying so.
- **The contributor job deliberately withholds a database service**, with the reasoning in the file: a stranger's laptop does not have one, and whether the document tells them to get one is what is under test. That choice already produced a real red at `db:migrate` with ECONNREFUSED on run 35170475688, and a real README change adding the `docker run … postgres:17-alpine` block. That is the instrument working exactly as designed, and the README is better for it.
- **`docker compose up` is run in the foreground as written**, backgrounded only so the job can continue, with the honest note that a literal reader now has a blocked terminal and needs a second one the document never mentions — *"That is a finding, not a reason to write `-d` here."* The README change addresses it in prose rather than by quietly changing the command. Right on both counts.
- **The sources assertion uses the README's own command.** `docker compose exec postgres psql -U ai_radar ai_radar` is documented at README line 67, and the job uses it with only `-T` added for a non-TTY runner. The editing rule holds where it would have been easiest to break.
- **Both jobs floor their success condition** rather than trusting an exit status: `sources` must be greater than zero, so a seed that seeded nothing cannot pass as one that worked.

## Verified by reading

The path filter contents against the files each job actually consumes · `.nvmrc` is `22` against the README's `22.13+` · the Dockerfile runs `npm ci` over `package-lock.json` at line 36 and pins `node:22-alpine` at lines 11 and 81 · `docker/web-entrypoint.sh` runs migrate then seed · `README.md` is unchanged on `main` since this branch's merge-base, so the graded document has not moved · the psql command matches README line 67 · the concurrency group and `permissions: contents: read` match the pattern `ci.yml` settled on · both `failure()` log steps and the `always()` teardown are present.

## Taken on trust

That run `35170475688` is the red described — I read the reasoning and the resulting README change, not the run itself. And that `ubuntu-latest` ships Node, Docker and openssl: this is documented runner behaviour and the job's own printed output would confirm it, but I did not open a run log to read the printed values.

zarghammir:
REVIEW VERDICT: APPROVE

Re-review at head `bbcf52a`, 0 commits behind `origin/main`, `MERGEABLE/CLEAN`. **Six check-runs plus the Vercel status, all `completed/success`** — including both quick-start jobs, which is the first run of this workflow that means anything. Reviewed by reading; no machine taken.

**The artefact-currency check, which is the one that can make this whole workflow meaningless:** `README.md` is unchanged on `main` since this branch's merge-base, so the job still grades the document it was written against.

All three findings are addressed, and the blocking one is addressed by testing the claim rather than by softening it.

## The blocking finding is genuinely closed, and the job proves the README's claim

The observation became an assertion:

```yaml
- name: Take the host's Node away, since the README says it is not needed
  run: |
    mkdir -p /tmp/no-node
    printf '#!/bin/sh\necho "the self-host path invoked the host node; …" >&2\nexit 1\n' > /tmp/no-node/node
    chmod +x /tmp/no-node/node
    echo "/tmp/no-node" >> "$GITHUB_PATH"
```

And the print step now says what it is — *"This step ASSERTS NOTHING — it is a record of what the runner shipped with, and the step below is what tests the claim"* — which is the right repair for a step whose name implied otherwise.

I checked that it actually ran rather than trusting the job's conclusion. Every step of `self-host, the README verbatim` executed, the stub sits at position 4, and every documented command after it succeeded: `cp`, `openssl`/`sed`, `docker compose up`, the wait for 200, and the seeded-catalogue assertion. The only skipped step is the `if: failure()` log dump.

**So this run is now a real result rather than a formality: the documented self-host path completes with the host's Node shadowed by a failing stub.** The README's claim that Docker builds it all is tested, and it holds. The reasoning about the checkout being the only JavaScript action is also borne out — its `Post` step ran successfully after the stub, because the runner invokes its own Node directly rather than through `PATH`.

### The one thing I would still add: the guard has no floor

The assertion is negative — *nothing invoked `node`* — and nothing establishes that the stub was in force. A green run is equally consistent with "the documented path never called Node" and with "`/tmp/no-node` never made it onto `PATH`". If a future edit moves the stub, reorders the steps, or `GITHUB_PATH` behaves differently than expected, the job goes green and the file still claims it tested the claim — which is a worse state than before this commit, because the comment now asserts coverage.

This is the pattern I have been filing all night, arriving in the fix for it: a negative assertion needs a positive beside it. Two lines in the same step close it:

```bash
command -v node | grep -qx /tmp/no-node/node || { echo "the stub is not in force; this job is not testing the claim" >&2; exit 1; }
```

placed in the *following* step, since `GITHUB_PATH` takes effect from the next step onward.

**I am not blocking on it**, and I want the reasoning visible rather than felt. All three findings are addressed; the mechanism is documented GitHub behaviour rather than a guess; the job is empirically green with every step executed; and the failure mode is the guard going quiet, not the product breaking or a false claim about the app. That is a hardening of a new guard, not a defect in the documented path. If it were my file I would add the check, and I would rather say so plainly than hold the merge for it.

## The other two, both closed well

**`.nvmrc` is in the path filter**, both on `pull_request` and on `push`, with a comment giving the reason nothing else covers it — *the image does not read it, the Dockerfile pins `node:22-alpine` directly*. That is the correct reason, and it is the half of my finding that survived checking; I had originally claimed `package-lock.json`, `drizzle/**` and `src/db/**` were also uncovered and they are not, because the Dockerfile runs `npm ci` and the entrypoint runs migrate and seed.

**The version gap is carried by #81** rather than absorbed. The comment states the problem precisely — the README's floor is 22.13, `.nvmrc` is `22`, so the job installs the latest 22.x and *"a contributor on exactly 22.13 — the version the README tells them is enough — is the one case this job cannot speak to"* — and it names the three possible decisions rather than leaving the reader to invent them. A known limitation with an owner and a decision list is the right shape; a known limitation in a comment with neither is the one this project has been bitten by.

**And `cache: npm` now has a comment saying why it is absent** — *"On a workflow whose whole purpose is a cold machine, a warm cache is precisely the thing that would make it prove nothing. Do not add it as a speed-up."* That was the deliberate choice most likely to be helpfully undone by someone optimising CI minutes, and it is now defended in place.

## Verified by reading

Six check-runs and the Vercel status, all conclusions read individually rather than counted · every step of the self-host job, in order, with the stub at position 4 · `README.md` unchanged on `main` since the merge-base · `.nvmrc` present in both path filters · the Dockerfile still pinning `node:22-alpine` at lines 11 and 81, which is what makes the `.nvmrc` reasoning correct.

## Taken on trust

That `ubuntu-latest` ships Node, Docker and openssl — documented runner behaviour, and the job's own print step records it, but I did not open the run log to read the printed values. And the earlier red at run `35170475688` that produced the README's `docker run … postgres:17-alpine` block, which I know from the reasoning rather than from the run.

Issue #16, opened 2026-09-16
Saved, Settings and first-run onboarding


zarghammir:
**Problem.** No way to save, no preferences, no onboarding.

**Solution.** Saved page: list with tags, note editing, archive, read/unread. Settings: interests (topic chips grouped by company / domain / field), morning brief time, timezone (auto-detected, editable), notification preference (push / email / neither, with email optional; push and email themselves arrive in Phase 3, the preference is stored now), theme, brief length. Onboarding: on first launch (no `onboardedAt`) a short three-step flow: interests → time and timezone → notifications; skippable; writes `user_preferences`.

**Expected output.** Playwright: fresh database → onboarding shows, choosing three interests and a time lands on Today; reload does not show onboarding again; ranking reflects chosen topics (a followed-topic story gains the topic component). Saved items survive reload; archive hides them from the default list. Screenshots attached.

zarghammir:
From the PR #24 review (design note): the seed overwrites sources.config on every run. Correct today, but the moment Settings lets the owner edit a source's config (or enable/disable through the UI is joined by config edits), that field must leave the seed's conflict set or the next seed reverts the owner's edit. This ticket owns that change when it adds source editing.

Issue #15, opened 2026-09-16
Story detail page with full provenance


zarghammir:
**Problem.** A reader must always be able to answer "where did this claim originally come from". No page shows it.

**Solution.** `app/(app)/story/[slug]/page.tsx` from `GET /api/stories/:slug`: title, executive summary (or excerpt), why it matters, key points, Original source (name, tier label, link, published time) shown first and unmistakably, timeline of when each source picked it up, related sources with roles, discussion links (HN thread), related research/repositories when present, verification chip with its plain-language note, and a collapsible developer panel "Why is this ranked here?" listing score components with labels and the total. Save/share/hide actions. Mark read on open.

**Expected output.** Playwright: for a story with a blog post and an HN thread, the page names the blog as the original source and the HN thread under discussion; the why-ranked panel's numbers sum to the score shown. Screenshots attached.

PR #63, opened 2026-09-17
docs: architecture — how it fits together, and why the odd parts are odd (#18)

### LINKED ISSUE #18: Clone-and-run proof plus architecture and source-adapter docs

zarghammir:
Closes #18 together with #58, which carries the clone-and-run proof.

## What it is

`docs/architecture.md` — how the project is put together, and why three decisions that look wrong are load-bearing.

## The split this was written under, because it should be checked differently

**Every structural claim was derived from the tree, not remembered.** Route count, module layout, the ten tables, what a page actually imports. Those are checkable and you should check them.

**Lines marked "Why" are reasons that are not visible in the code.** They come from the decisions behind it — PR bodies, review threads, incidents from today — and they are the half a fresh reader cannot recover by reading the tree, because the tree does not record why anything is the way it is.

A long memory of this repository is a liability for *what is there* and an asset for *why it is there*. Mixing the two silently is how a document ends up describing a shape that no longer exists. Hence the marking.

## What it records that the code cannot

- **`output: "standalone"`** is what the Docker runtime image ships, and the Dockerfile asserts `.next/standalone/server.js` at build time. It reads like a leftover; removing it breaks self-hosting immediately.
- **The internal ingest route imports the database client inside the handler.** `next build` imports every route module to collect page data, and `src/db/client.ts` throws without `DATABASE_URL` — so a top-level import fails any build with no database, which is exactly what the image build is.
- **`getDb()`/`getSql()` are functions** so that importing a module never opens a connection.
- **`routes:check` has no exceptions list**, with the incident that produced it: nine of twelve routes once reached the adapters through one import line, no route *called* any of it, so the rule held and could not be *verified*.
- **One writer, one lock, one pass** — including why the lock reserves a connection, and why `runIngest` and `rankAllStories` are a single call rather than two jobs.

## One thing stated because it reads as a defect otherwise

**The screens do not consume the API yet.** `src/app/radar/page.tsx` imports components and nothing else. The API is finished and exercised by tests before it has a caller, and a reader who does not know that would reasonably conclude a route is dead.

## How to review this

A cold reader is the right instrument here, and the useful question is not whether it is well written: **if the document teaches you something the tree does not support, that is the finding.** The structural claims are all checkable against the files they name. The "Why" lines are the ones that cannot be confirmed from the code — they are assertions about history, and if one of them is wrong it will be wrong in a way that matters, because those are the sentences that stop someone tidying a load-bearing decision away.

## Review round: two claims were false on main, and the cause was general

Both confirmed by reading the merged tree rather than taken from the verdict.

- **The `src/api/` module list named seven files; there are eight.** `reading-budget.ts` arrived with the Today page.
- **"The pages are shells" described a tree that no longer exists.** `src/app/page.tsx` now imports `BriefList`, `ReadingMode`, `briefSummary` and `getBrief`.

**The insight underneath the second one is true and is what survives:** no screen calls the route handlers. Nothing under `src/app`, `src/components` or `src/lib` fetches `/api/…` — the Today page reads through `src/lib/api/client.ts`, which serves fixtures under an explicit `NEXT_PUBLIC_USE_FIXTURES` mode rather than a fallback. That is the reason the routes look dead, and it is the part worth keeping.

**The general defect was stating fragile facts in the same confident voice as durable ones.** A file count and a reason read identically on the page, and only one of them survives a normal week's work. So every count is now either pinned by a test or phrased so ordinary work cannot falsify it, and where a list would rot the document names the directory instead of its contents.

| claim | what it is now |
| --- | --- |
| ten tables | **pinned** — `schema.test.ts` floors the count |
| route handlers | phrased durably, with `npm run routes:check` named as the live source of the pair |
| the `src/api/` modules | no longer enumerated — the two shared modules are named, the directory is the list |
| the pages | replaced by the durable claim about route handlers and the fixtures mode |
| what is not built | no longer a list of issue numbers; the Phase 1 milestone stays current and this does not |

**Control on the new floor, run in the same window:** adding an eleventh table to the schema reddens that test. A floor that has never been seen to fail is not yet known to work.

**Scope of that control, stated precisely:** I ran `npx vitest run src/db/schema.test.ts`, so what I measured is that the control reddens exactly one test *within that file* and leaves the other five green. I did **not** run the whole suite under the control, so "and nothing else" — which an earlier version of this body claimed — is broader than what I measured.

**One thing I did not do, because it is not my file.** The reviewer noted that `scripts/check-route-cost.ts` already prints "12 public routes … 1 internal" and could assert that pair against the document in one line. It came with #42 and belongs to the Intelligence lane, so I have asked rather than edited. The document points at the command instead, which needs nobody's permission.

## Second review round: a false absolute, and a pin that promised too much

**The read-path sentence was wrong, and wrong in the category this document exists to avoid.** "Nothing under `src/app`, `src/components` or `src/lib` fetches `/api/…`" is false: `src/lib/api/client.ts` requests `/api/brief`, `/api/saved/:id` and `/api/hide/:id`, each immediately after an `if (USING_FIXTURES) … return`. **What makes them quiet is an environment variable, not a property of the tree** — a runtime condition stated as a structural fact. I replaced a stale snapshot with a stronger absolute, and the absolute was wrong the day it was written, where the snapshot had at least been true once.

It also contradicted the bullet two sections below it, which was the correct one.

The replacement states the intended structure rather than a census — screens go through `client.ts`, whose calls are inert while the flag is on — so a future direct caller reads as an architecture violation rather than falsifying a count, and it survives #74 adding three more call sites.

**Why my sweep missed it, since the mechanism is the transferable part:** my pattern included a bare `/api/`, which matched every `@/lib/api/...` **import path**, and I piped the result through `head -5`. The noise filled the window and the real call sites were below the cut. A truncated, noisy result read as "nothing".

**And the document claimed more for the table floor than the floor delivers.** The test asserts a literal that a human keeps in step with the prose; it cannot read the document. Bump the literal, leave the prose, and it rots as before with a green suite over it. The honest claim is that the count can no longer drift **silently**, and the test now says that about itself in its own comment.

## The three "Why" lines I would most want challenged

A cold reader cannot know which of these I am confident about. I can, so here they are ranked by how little I actually verified myself.

**1. "Two concurrent passes can pick the same slug and collide on the unique index."** This is the reason given for the single-writer lock, and it is the one I am least sure of. It reached me as a carry-in from the #30 review rather than from my own reading: I confirmed `uniqueSlug(tx, title)` exists in `src/pipeline/run.ts` and I never traced the collision. The lock is right regardless — it is also what closes the vanished-story race in the ranker — but **if the slug mechanism cannot actually collide, this sentence teaches a wrong reason for a correct guard**, and someone will eventually act on the reason.

**2. "Nine of twelve routes once transitively imported the adapters, through `stories.ts` → `rank-all.ts` → `run.ts` → the registry → all three adapters → `src/sources/http.ts`."** The import chain and the number both came to me second-hand from #42's finding. I did not walk the graph. It is the justification for `routes:check` having no exceptions list, so it is load-bearing for a rule, and the chain is checkable against the tree at the commit before #42 merged.

**3. "A story nobody scored cannot appear on Today."** This is the reason `runIngest` and `rankAllStories` are one call, and it is a claim about a screen that **does not exist yet** — #13 is open. It is forward-looking by construction, so it cannot be verified today and may turn out to be wrong about how Today actually queries.

The other Why lines I established myself and would defend: the pooled-connection failure (reproduced in a scratch probe, unlock returned false with the lock still held in `pg_locks`), the `next build` page-data collection (reproduced locally, then caught in CI on another lane's PR), and the `output: "standalone"` assertion (read in the Dockerfile).


zarghammir:

Reviewed at head `7f5df14`, one file, 151 added lines. Reviewed entirely by reading the tree — no build, no suite, no browser, and I did not take the machine. **There are no claims in this document I was unable to settle by reading.**

This is an unusually careful document and almost everything in it is true. Its preamble sets the bar itself: *"Structural claims here were derived from the tree — file paths, imports and counts are things you can check."* I checked them. Two fail, and they fail against `main` rather than against the branch — which is the point below.

**The branch is 9 commits behind `origin/main` (`1c0e5ac`).** For code that is a merge question. For a document whose subject is the current tree, it is a correctness question, because the tree it describes has moved underneath it.

---

## 1. The module list is already wrong on the branch this merges into

Line 80: *"`src/api/` holds the work: `brief`, `radar`, `stories`, `catalogue`, `reader`, plus `params` and `http`."*

At this PR's own base that enumeration is exactly right — seven files, seven named. On `origin/main` there are eight:

```
brief.ts  catalogue.ts  http.ts  params.ts  radar.ts  reader.ts  reading-budget.ts  stories.ts
```

`reading-budget.ts` arrived with the Today page. The sentence was true when written, is false where it is going, and nothing anywhere will ever say so. This is the cleanest example in the document of the thing the document is otherwise very good about.

## 2. "The pages are shells" describes a tree that no longer exists

Lines 84-87: *"**The screens do not consume these yet.** `src/app/radar/page.tsx` imports components and nothing else — the pages are shells while the Today page is built."*

At this PR's base, `src/app/page.tsx` imported `PageShell` and `LocalDate` and nothing more — a genuine shell, and the sentence was accurate. On `origin/main` it imports `BriefList`, `ReadingMode`, `briefSummary`, `getBrief` and `USING_FIXTURES`. **The Today page is built.** A reader arriving at this document on main is told the opposite of what they will find.

The narrower claim survives and is worth keeping: the screens still do not consume the *route handlers* — Today goes through `@/lib/api/client`, which is fixtures. That distinction is the useful part and it is currently buried inside a sentence whose framing has expired. Separating "no screen calls the API routes yet" (durable, and the reason the routes look dead) from "the pages are shells" (a snapshot) would keep the insight and drop the part that rots.

Related, same cause — line 149, *"Not built yet: The screens (`#13`, `#14`, `#15`, `#16`)"*. The Today page is merged, and `#16` is PR #74, in flight with Saved, Settings and onboarding. This line is partly wrong now and more wrong the day #74 lands.

## 3. What #57 and #74 will do to it, since you asked

- **#74 invalidates nothing new.** It touches no file under `src/api/` and adds no route handler, so "thirteen route handlers" and the `src/api/` sentence are unaffected by it. It does push item 2 further out of date.
- **#57 invalidates nothing at all.** It touches `src/db/schema.ts` and `drizzle/`, but it adds a *column* and an *enum type*, not a table — so "Ten tables" survives it. And the paragraph most exposed to it is the one it strengthens: verification level and content type stay separate columns with separate vocabularies, which is exactly what #57's provenance work preserves.
- I checked the document against your #66/#77 correction and it does not apply: the only mention of content types is lines 133-135, which is about the two columns staying distinct. It states no badge count and names no individual type, so there is nothing here to be wrong about five-versus-four or about TREND being unreachable.

## 4. Which claims can rot, and what could pin them

The document already draws the right line in its preamble — structural claims versus **Why** lines — and the **Why** lines are the durable half. A reason does not expire; a count does. What is missing is that nothing pins the counts, so they all rot silently and identically:

| claim | verified | can it rot silently? |
|---|---|---|
| thirteen route handlers, twelve client + `internal/ingest` | **true** | yes — nothing asserts it |
| ten tables, and the ten names | **true** | yes |
| the `src/api/` module list | **false on main** | yes, and it has |
| `src/app/radar/page.tsx` imports components only | **true** (plus a `Metadata` type import) | yes |
| the five documents in "Where to look" all exist | **true** | yes |

Two of these are nearly free to pin, and both hooks already exist. `scripts/check-route-cost.ts` already walks and counts the routes — it prints "12 public routes … 1 internal route" — so asserting that pair against the number in this document is a one-line addition to a script that already has the data. Likewise `src/db/schema.test.ts` already imports the schema and could floor the table count. I am not asking for either as a condition of merge; I am saying that where a hook already exists, an unpinned number in a document is a choice rather than a necessity.

---

## Verified true, by reading

Everything below I checked against the tree rather than accepted:

- **Thirteen route handlers** under `src/app/api/`, twelve client plus `internal/ingest` — 13 at the doc's base and 13 on main.
- **Ten tables**, and all ten names exactly as listed: `sources`, `raw_items`, `stories`, `topics`, `story_topics`, `user_preferences`, `saved_items`, `read_state`, `ingest_runs`, `llm_usage`.
- **`output: "standalone"`** at `next.config.ts:8`, and the Dockerfile really does assert it at build time — `test -f .next/standalone/server.js` at line 77. Removing it would break the self-host path, as claimed.
- **The internal ingest route imports the client inside the handler**: `const { getDb, getSql } = await import("@/db/client")` on line 14, with the module's only top-level import being the trigger.
- **`getDb()` and `getSql()` are functions**, `src/db/client.ts:65` and `:71`.
- **`routes:check` has no exceptions list** — and the script says so itself, in the same words and for the same reason.
- **`schema.test.ts` fails if verification and content type are merged** — it asserts both enum column sets independently and floors their lengths.
- **All five documents** in "Where to look" exist: `docs/sources.md`, `docs/api.md`, `docs/DESIGN.md`, `docs/cost-protection.md`, `README.md`.
- **`ingest.yml` runs `npm run worker:once` on `*/30 * * * *`** — every thirty minutes, as stated.
- The Processes table's four entry points all exist at the paths given.

The **Why** sections are the best part of this document and I have no quarrel with any of them. "A carve-out is a hole with a comment on it" and the advisory-lock-across-pooled-connections explanation are both the kind of thing that is invisible in the code and expensive to rediscover. They are also, correctly, the claims the preamble invites you to challenge if the code disagrees — and the code does not.

## Taken on trust

- **The historical claim at lines 115-120** — that nine of twelve routes once transitively imported the adapters through `stories.ts` → `rank-all.ts` → `run.ts` → the registry → `src/sources/http.ts`. That is a statement about a tree that no longer exists, so it is not checkable by reading this one. It is a **Why** line and reads as one; I flag it only because it is the single number in the document with no present-day referent.
- That `LLM_MAX_STORIES_PER_DAY` and `llm_usage` are read by nothing — I confirmed the table exists and is listed, but did not trace every consumer to prove the negative.

zarghammir:

Re-review at head `76faa6b`, new comment rather than an edit. **0 commits behind `origin/main`** — merging main was the right first move, and it is the reason both of my previous findings could be checked against the tree this actually lands in. All five checks pass at this SHA. Reviewed by reading; no machine taken.

Both previous findings are genuinely fixed, and the remedy programme is better than what I asked for. Three fragile numbers got three different treatments — pinned, made durable, or deleted because the directory is the list — and deleting the `src/api/` enumeration is the strongest of the three, because a list that cannot be wrong beats a list that is checked.

One thing blocks, and it is in the sentence that replaced my finding 2.

---

## BLOCKING — the new absolute is false today, and #74 makes it more false

Line 23 of the read path:

> **No screen calls these route handlers.** Nothing under `src/app`, `src/components` or `src/lib` fetches `/api/…`

The first sentence is true and valuable. The second is not. `src/lib/api/client.ts` is under `src/lib`, and on this branch it contains three:

```
src/lib/api/client.ts:91    json<BriefResponse>(`/api/brief?length=${…}`)
src/lib/api/client.ts:102   json(`/api/saved/${story.id}`, { method: saved ? "POST" : "DELETE" })
src/lib/api/client.ts:113   json(`/api/hide/${story.id}`, { method: "POST", … })
```

Each sits immediately after an `if (USING_FIXTURES) { … return }` guard. So the code that fetches `/api/…` exists in one of the three named directories, and what makes it quiet is `NEXT_PUBLIC_USE_FIXTURES`, which is an environment variable, not a property of the tree. **The claim's truth is a runtime condition stated as a structural fact** — which is the exact category this revision set out to eliminate.

**#74 lands before this and takes it from three to six.** At `e10842d`, `client.ts` adds `/api/saved?archived=`, `/api/saved/${storyId}` and `/api/read/${storyId}`. The sentence gets further from true by the merge that precedes it.

**The document already contains the accurate version of this, two sections down**, which is how I am confident about the intent rather than guessing at it:

> **Nothing consumes the API yet.** The flip is `NEXT_PUBLIC_USE_FIXTURES=0`, and it is the single change that turns the screens from a prototype into the product.

That is right, and it implies what the read-path sentence denies: the calls exist, and one variable activates them. The two passages contradict each other and the bullet is the correct one.

The fix is one sentence and loses almost nothing:

> **No screen calls a route handler directly.** The only caller is `src/lib/api/client.ts`, and its `/api/…` calls are inert while `NEXT_PUBLIC_USE_FIXTURES` is on — which is why the API is finished and exercised by its own tests before it has a live caller.

That keeps the warning the paragraph exists for — *worth knowing before you conclude a route is dead and delete it* — and it survives both #74 and #65 without editing.

I swept for this rather than eyeballing it: every `.ts`/`.tsx` under `src/app`, `src/components` and `src/lib`, excluding tests, for `"/api/`, `` `/api/ ``, `'/api/`, `fetch(`, `sendBeacon`, `EventSource`, `new Request(`, `useSWR`, `useQuery`, `axios` and `action="/`, on **both the branch and `origin/main`**. The only hits outside `client.ts` are `RouteContext<"/api/…">` type parameters inside the handlers themselves, which are declarations rather than calls.

## The pin is real, and it pins something narrower than the sentence claims

The mechanism is sound: `Object.values(schema).filter((v) => is(v, PgTable))` against `toHaveLength(10)`. That genuinely counts the exported tables, and an eleventh breaks it. Good.

But the document says:

> Ten tables — a count `src/db/schema.test.ts` now floors, **so this sentence fails a test rather than rotting quietly.**

Strictly, the sentence does not fail a test. The *schema* fails a test whose `10` is a literal that a human keeps in step with the prose. Add a table, update the literal, forget the document, and the sentence rots exactly as before — with a green suite. The test's own comment is more careful than the document's claim: *"Adding a table is fine — update both."*

This is not a request to parse markdown from a test. It is a request that the document not claim a stronger guarantee than the guard provides. The honest version is that the count can no longer drift **silently** — a change to the schema now stops and tells you to update both — which is a real improvement over the previous nothing, and is what the test actually delivers.

**Not verified:** I did not run the control. The claim relayed to me is that adding an eleventh table reddens exactly that test and nothing else. By reading, an added export changes only the length, so I expect it to hold — but "and nothing else" is a whole-suite property and I have not measured it. `npm test` with a throwaway `pgTable` added, then reverted, settles it in under a minute.

## Verified against the merged tree

Item by item, because nine commits arriving at once is exactly when a correction lands on a stale version of the new state:

- **Both previous findings are fixed at the root.** Eight modules under `src/api/` including `reading-budget.ts`, and `src/app/page.tsx` importing `BriefList`, `ReadingMode`, `briefSummary` and `getBrief` — so the enumeration and the "pages are shells" snapshot are both gone rather than patched.
- **The route pair holds:** 13 handlers under `src/app/api/`, exactly one under `internal/`, so twelve and one. And it is now phrased as a convenience with `npm run routes:check` named as the source of truth, which is the right shape for a number that moves.
- **Ten tables holds**, and **#74 does not change it** — `schema.ts` at `e10842d` still has ten. #74 also adds no route handler, so the route pair survives it too.
- **All five "Where to look" targets exist.**
- **Nothing else in the document is falsified by #74** as far as I can see: it adds screens and client modules, not tables, routes or processes. The read-path sentence above is the only casualty, and it is already blocking.

## Credited

- **Deleting the `src/api/` list** rather than correcting it. A list that cannot be wrong is better than a list that is right today, and it is the harder call to make about your own prose.
- **Dropping the enumerated issue numbers from "Not built yet"** in favour of pointing at the milestone, on the stated ground that the milestone stays current and a document does not. That is the same reasoning applied to a second instance without being asked.
- **Three fragile facts, three different remedies** — pin, durable phrasing with a live source, deletion — chosen per fact rather than one mechanism applied uniformly.
- **The fixtures-are-a-mode reasoning**, quoting `client.ts`: a client that tries the network and quietly uses fixtures on a 404 cannot tell "this route does not exist yet" from "this save failed". That is the most useful paragraph added in this revision and it is the kind of thing only the author knew.

## Taken on trust

The control on the table-count pin, as above. And the historical claim at the routes-check section — that nine of twelve routes once transitively imported the adapters — remains a statement about a tree that no longer exists and is not checkable from this one.

zarghammir:
REVIEW VERDICT: APPROVE

Third pass at head `305c218`, 0 commits behind `origin/main`. **All five checks pass at this exact SHA** — four check-runs plus the Vercel status. `browser checks` was still `in_progress` when I started and I polled it to `completed/success` rather than counting its presence. Reviewed by reading; no machine taken.

Both findings are closed, and the first is closed better than the replacement I suggested.

## The read-path claim is now structural, and I measured that it survives #74

> **Screens do not call route handlers directly.** They go through `src/lib/api/client.ts`, and its `/api/…` calls are inert while `NEXT_PUBLIC_USE_FIXTURES` is on — the calls exist, and one variable activates them.

This is better than what I proposed. My version still described a census — *the only caller is client.ts* — which a second caller would falsify. This describes the intended **structure**, so a future direct call from a screen reads as an architecture violation rather than as a documentation error. That is the difference between a sentence that has to be maintained and one that has to be obeyed.

I checked the thing you would want checked, against #74's head rather than this branch, because #74 lands first and adds call sites:

```
/api/… literals at e10842d, outside the handlers' own RouteContext type params:
  src/lib/api/client.ts   ×9   (brief, saved ×3, hide, read, preferences ×2, topics)
  anywhere else            0
```

Nine call sites where there were three, every one inside `client.ts`, and **no screen calling a route handler directly**. The claim holds unchanged across a merge that triples what it describes — which is exactly what the rephrasing was for.

## The pin's limitation is now stated where someone about to break it will read it

`schema.test.ts` carries it in its own comment:

> *Note what this does NOT do: it cannot read the document. Bump the literal below and leave the prose alone and it rots exactly as before, with a green suite over it. A pin on the code is not a pin on the prose that cites it.*

That is the right location — the person editing `toHaveLength(10)` is the person who needs to know, and the document's reader does not. The prose is correspondingly narrowed to what the guard delivers: the count *can no longer drift silently*, adding a table stops the suite, and keeping the two in step is still a person's job. The previous "this sentence fails a test" is gone.

Three-way agreement, checked: `schema.ts` has **10** tables, the pin asserts **10**, the prose says **Ten**.

## Verified at this head

- Thirteen route handlers, exactly one under `internal/` — the twelve-and-one pair, still phrased as a convenience with `npm run routes:check` named as the source of truth.
- All five "Where to look" targets present.
- The "Not built yet" flip bullet is intact and consistent with the new read-path paragraph.

## One observation, with an owner, not a finding

The final clause — *"the API is finished and exercised by its own tests before it has a live caller"* — is what **#65** falsifies, since flipping `NEXT_PUBLIC_USE_FIXTURES` to live by default gives it exactly that. It is paired with the "Nothing consumes the API yet. The flip is `NEXT_PUBLIC_USE_FIXTURES=0`" bullet, so the two go stale together and the document is internally consistent about the state it describes.

I am not asking for a change. A document describing the pre-flip architecture should not pre-write the post-flip one, and the merge that makes it false is the merge that should update it. Naming it here so it is a known handoff to #65 rather than something discovered afterwards.

## Not verified

The control on the table-count pin. I declined to assert it last round because "reddens nothing else" is a whole-suite property I had not measured, and I still have not run it. My understanding is that it was since run against one file, which is a narrower claim than the original and matches what was actually done — that correction is the right handling and I am recording that I did not re-derive it.

The historical claim in the routes-check section — nine of twelve routes once transitively importing the adapters — remains a statement about a tree that no longer exists.

Issue #48, opened 2026-09-17
Screenshot harness: named Playwright projects, npm run screenshots, docs/screenshots/<area>/


zarghammir:
Split out of #17, which delivers the CI job that runs the existing browser checks. **This is the half #17 does not deliver**, and it is the half the owner sees.

What exists today: the a11y audit already captures screenshots as a side effect of its sweep (six routes × two widths × two themes), and PR #31 shipped nine of them by hand. What does not exist: a named, repeatable way to produce them — a Playwright config with named projects, an `npm run screenshots` entry point, and the `docs/screenshots/<area>/` layout so a PR can show the screen it changed without the author assembling it manually.

**Why it matters beyond tidiness:** the owner reads this project through pictures. Every PR touching a screen should be able to produce its own evidence cheaply, and every PR that does not touch a screen should be able to show the screen it protects. Today that takes a lane remembering to capture, name and attach them.

Owned by **lane:ui** because these are its scripts and its capture logic — do not fork a second screenshot path in the CI lane. Do it after #13.

Requirements carried from the #17 work:
- names are stable across runs, so a diff between two PRs' screenshots is meaningful rather than a reshuffle
- the set is derived from the routes the audit already sweeps, not a hand-written list that rots
- a floor: the run fails if it captured fewer images than routes × widths × themes, so an empty or partial capture cannot pass as a clean run
- no network dependency — see the fixture ruling on #17: these pages render seeded fixture content, never a live ingest

Not urgent. Filed so the missing half has an owner rather than being assumed into #17's close.

zarghammir:
**Adding a requirement to this ticket by the PM's ruling, from #74's review: regeneration is not enough on its own — something has to go RED when a screen changes.**

## The hole, found by walking into it

While deciding whether to commit four per-PR screenshots to #74's branch, we settled a distinction — a *documented set* (`docs/screenshots/<area>/`) earns its bytes by staying true, per-PR evidence does not. Then the justifying half turned out not to exist.

**Nothing regenerates the documented set today.** No script, no `package.json` entry, nothing in `scripts/` references `docs/screenshots` at all. There are **18 tracked images, 2.4 MB**, across three areas — `docs/screenshots/shell/`, `docs/screenshots/today/` and a `design/screenshots/` set that had not come up in any of this. They are a hand-maintained claim wearing the authority of evidence.

[...]

PR #65, opened 2026-09-17
Today reads the live API, and an unreachable database is not a quiet morning

### LINKED ISSUE #62: Today: read from the live API by default, and correct three stale explanations
**Flip the Today screen from fixture stories to the live API — and fix three sentences that tell the next contributor to wait for something that already happened.**

#52 was built while the twelve API routes were unmerged, so it reads from fixtures behind an explicit flag (`NEXT_PUBLIC_USE_FIXTURES`). The routes landed in #42 during that PR's review. The flag design stays — a client that tries the network and silently falls back on a 404 cannot tell "this route does not exist yet" from "your save failed", and would report a failed write as a success. What changes is the default.

**The stale sentences, which are the reason this is a ticket rather than a note.** #52's body was corrected to say the routes landed. The same claim survives in three other places:

- `src/lib/api/client.ts:7` — "The routes on feat/api-routes are not merged"
- `src/lib/api/client.ts:13` — "once /api/brief exists"
- #52's own summary line — "Fixture mode until the routes merge"

The code comment is the one that matters: it is what the next contributor reads when deciding whether to flip the flag, and it tells them to wait for something that is already done.

**This is the third instance of one family inside a single PR**, which is why it is worth stating rather than quietly fixing: first the fixture DATA claimed reading times the pipeline cannot produce; then the selection RULE diverged from the server's; now the EXPLANATION describes a world that ended during the review. Each fix landed on its own instance. The family is "an artefact describing a state that has moved" — and the review caught all three separately rather than as one.

**Acceptance:**
- Today reads from `/api/brief` by default; the fixture flag still works and is still explicit.
- All three sentences above are correct, and a grep for "not merged" / "once /api/brief exists" returns nothing stale.
- The empty state and the error state are both exercised against the real route — an empty database must look like a quiet morning, not a broken screen.
- A failed save still announces itself rather than silently reverting.

Owner: UI lane. Raised by the reviewer on #52 with the note that a deferred item without an owner is a dropped item on this project.

### LINKED ISSUE #91: ⛔ Saves are per-browser today and become shared the moment #65 lands — decide before it merges
**⛔ THIS BLOCKS #65. Raised by the owner on 2026-09-17 after reading the cost analysis.**

## Today it works. The live-data switch breaks it.

`src/lib/api/client.ts:70` says it plainly:

> *"Applies the reader's own saved/hidden state to a brief. In fixture mode that state lives in localStorage; **against the real API it arrives on the card**."*

So **right now saves are per-browser** — every visitor has their own, because the app is not talking to its own database. That is correct behaviour arrived at accidentally, as a side effect of fixture mode.

**#65 turns the live API on.** From that moment saves come from `saved_items`, where `storyId` is `.unique()` and there is no user column — **one saved list for the entire application.** `read_state` is the same shape, and `user_preferences` is a single row by design (`schema.ts:206`: *"V1 is single-user (self-hosted, no login). One row with id = 1."*).

**The first thing real data would do is break saving.** Person A saves a story, person B sees it saved; person B marks something read, it is read for everyone.

## This is not a bug — it is a design being outgrown

Single-user, self-hosted, no login is the **right** design for the product as conceived: an open-source app people deploy themselves. Nothing here was done wrong. **The owner deciding to put it on a public URL is what changes the requirement**, and that decision was made today.

## Four ways out, and they are not close in cost

**1. Saves stay on the device — localStorage only, even after the live switch.**
Each visitor gets their own saves. No login, no accounts, no personal data stored anywhere, no new table. The server still serves one shared *catalogue* — which is correct, the news is the same for everyone — and the *reader's* state stays with the reader.
- *Cost:* small. Keep `applyLocalState` on the live path instead of reading `saved` from the card.
- *Limit:* saves do not follow a person between phone and laptop, and clearing browser data loses them.

**2. Anonymous device identity.**
A random id in a long-lived cookie; the server keys saved/read rows by it. Same experience as 1, survives a cache clear, still no login and still no personal data.
- *Cost:* a column on three tables, a cookie, and a migration.
- *Limit:* still per-device, not per-person.

**3. Real accounts.**
Email or OAuth login, sessions, cross-device sync.
- *Cost:* large, and it is not mostly the login — it is password reset, account deletion, a privacy policy, and becoming the custodian of other people's data.

**4. Self-host only.**
Do nothing. The public URL stays the owner's personal instance; anyone who wants it deploys their own free copy.
- *Cost:* none. It is what the app already is.

## Recommendation

**Option 1 now, before #65 merges.** It is small, it is honest, it costs nothing, it stores no personal data, and it works for a hundred readers or a hundred thousand. **Option 2 when saves-across-devices actually matters.** Option 3 only if the product becomes one where an account earns its keep.

**Option 4 remains true regardless** — self-hosting is the designed path and stays supported either way.

## Acceptance

- After #65, two different browsers saving different stories **do not see each other's saves**. Tested with two contexts, not reasoned about.
- **A control:** make both browsers share an identity and the test goes red — otherwise it is asserting something that was already true.
- `read_state` gets the same treatment as `saved_items`, or the same defect ships one field over.
- Whatever is chosen, `schema.ts:206`'s comment is updated. **It currently states the design accurately; the day this changes it becomes the most misleading line in the file.**

Related: #65 (blocked by this), #74 (built the Saved screen), #16.

### LINKED ISSUE #94: ⛔ user_preferences is one shared row — including the reader's email address. Blocks #65
**⛔ BLOCKS #65. Found by the UI lane while applying #91's ruling — it is the same defect one table over, which is the sweep-the-family rule doing its job.**

## The fact

`user_preferences` is a **single shared row with `id = 1`**, and every field in it belongs to the **reader**, not to the instance:

| field | whose it is |
|---|---|
| `topicKeys` | what **you** follow |
| `briefTime`, `timezone` | when **you** want your brief |
| `briefLength` | how long **you** want it |
| `notificationChannel` | how **you** want to be told |
| `theme` | **your** appearance choice |
| `onboardedAt` | whether **you** have seen the welcome |
| **`email`** | **your email address** |

On a public URL with no accounts, that means: person 47 sets their brief time and person 12's window moves. Person 47 finishes onboarding and person 12 never sees the welcome screen.

**And person 47 types their email into Settings, where person 12 can read it back from `GET /api/preferences`.**

**That last one is not a preference collision. It is one person's personal data served to another** — the exact category of obligation the owner ruled he does not want on #91.

## Severity: prospective, not live — verified

`GET /api/preferences` on production right now returns `email: null`, `onboardedAt: null`, everything else at defaults. **Nobody has entered anything.** The exposure begins the moment #65 ships and a second person opens Settings.

### 1. The email field goes, rather than being protected

**#72 already records that nothing sends a brief: the notification preference is stored and read by nobody.** So the email field has no function today. It collects personal data to serve a feature that does not exist.

**You cannot leak what you do not collect.** Remove the field and the control that fills it. When #72 gives it a purpose, it comes back **with** whatever identity model that feature needs — which is the right time to decide it, not now.

### 2. Everything else becomes reader state on the device

Same treatment as saves and `read_state` under #91: `topicKeys`, `briefTime`, `timezone`, `briefLength`, `theme`, `onboardedAt` all live client-side.

**`onboardedAt` is the one that cannot be argued** — UI already built a client-side half for it on #74, precisely because the server could not see it. This finishes that.

**`briefTime` and `timezone` are arguably instance config** — the schedule a self-hoster sets for their own deployment. **Decide that explicitly rather than by default.** If they stay server-side, say why, where the field is defined.

## Acceptance

- After #65, **two browser contexts setting different preferences do not see each other's** — tested with two contexts.
- **A control: share an identity between the contexts and the test goes red.**
- `GET /api/preferences` returns **no field belonging to a reader**, or the route goes.
- The `email` column's removal is a migration, and the reason is recorded where the column was.
- `schema.ts:206`'s comment covers preferences as well as saves.

Related: #91 (the ruling this extends), #65 (blocked), #72 (why the email has no purpose yet), #74 (built the client-side onboarding half).

zarghammir:
Closes #62.
Closes #91.
Closes #94.

**Today stops showing examples and starts showing real news.**

The fetcher has been collecting from your sources every half hour, the ranker has been putting them in order, and the endpoints have been serving them. The screen was still reading a prepared sample. This is the switch. Nothing underneath changes; what changes is that you can finally see it.

**Working without a database is still possible.** It becomes something you turn on deliberately — `NEXT_PUBLIC_USE_FIXTURES=1` — rather than the thing that happens by default. That matters for anyone building a screen on a laptop with nothing installed.

## The defect this found, which is bigger than the switch

**Today was never reaching the API at all, and the screen looked fine.**

The page is rendered on the server. It was calling `fetch("/api/brief")` — a relative address, which has no base on the server, so **every single load threw** while the route itself answered correctly. The page then rendered its "cannot reach your stories" state. Against the built-in examples nothing touched the network, so nothing showed. Against an empty database it would have looked exactly like a quiet morning.

The only reason it was caught is that the three states name themselves in the page, so a check could tell "we could not reach the stories" apart from "nothing arrived today". A server component fetching its own API is a round trip through the process it is already inside. It now calls the same functions the route calls.

## The three states this ships honest about

| State | What it means | Why it is not collapsed into the others |
|---|---|---|
| **brief** | There are stories | — |
| **quiet** | The database answered and nothing arrived | A fact about the day. It says so, and says the worker will fill it |
| **unreachable** | The database could not be read | A fact about the app. It says what to check, and **does not print the error** |

**The error stays in the server log.** A driver message is the failed SQL with column names, and a connection error can carry the host and the credentials out of `DATABASE_URL` onto a page somebody screenshots. The person who needs that detail is the one with the log. Every screen added in #16 answers a failed read the same way, so this is the app's rule now rather than this page's habit.

| Still a placeholder | Why it is not hidden |
|---|---|
| **Radar**, **Research**, **Releases** | Routes and design exist, the data does not. `docs/architecture.md` now names these three rather than claiming nothing consumes the API |

## Measured, not claimed

Production build, standalone server on 3210, real browser.

**Re-measured by CI on the rebased branch, not by me on my laptop.** The earlier numbers in this section were taken at a commit a rebase later rewrote away, so they were replaced rather than re-presented. These come from the `browser checks` job at `326d861`:

| what | measured |
|---|---|
| seeded brief length | **11 minutes** against a shortest budget of 5 |
| reading-length switch | `all=11` → `10 min=10` → `5 min=5` — it demonstrably shortens |
| browser checks derived | **5**, and every browser-driving script accounted for |
| controls that reddened as required | `a11y`, `verify:isolation`, `verify:saved` — all exit 1 |
| checks that ran and passed | `a11y`, `verify:isolation`, `verify:saved`, `verify:shell`, `verify:today` |
| declared no control, named not absorbed | `verify:shell`, `verify:today` (#113) |

**The green is quoted as artefacts rather than as a verdict on purpose.** An earlier run on this branch was green because two of these checks **did not run at all** — the job died in the control loop before reaching them — and an absence reads exactly like a pass from the outside. The derived count, each control's result line and the per-check `MEASURED` digests are what distinguish the two, so they are what this table reports.

**This sentence deliberately does not say how far behind `main` the branch is.** It has been wrong in both directions: it claimed `0` when the branch was four behind, then `4` after a rebase had made zero true again — the correction outliving the error, in a line whose whole job was to stop that. `git rev-list --left-right --count origin/main...origin/feat/today-live` is the answer that cannot go stale.

**The two states that must never look alike, proved in LIVE mode against a real database** — this is the claim the PR is actually making, so it is measured in the mode it ships in rather than on fixtures:

| Database | `data-screen-state` | What the screen says |
|---|---|---|
| Reachable, **empty** | `quiet` | "quiet morning" |
| **Not there** | `unreachable` | "Cannot reach your stories" |

And the leak check on that second one: **0** occurrences of the connection string, the host, the port or `ECONNREFUSED` anywhere in the rendered HTML, with the reason logged **4** times in the server log where it belongs.

| Check | Result |
|---|---|
| `next build` | **0** |
| Unit tests | **390 passed, 32 files** |
| `tsc --noEmit` / `eslint` / `prettier --check` | **0 / 0 / 0** (2 eslint warnings, both on `main` at `aa5e9c8`) |
| `verify:shell` | **0** — 3 radios, one tab stop, arrows move focus, worker controlling, live 404 |
| `verify:today` | **0** — 5 stories at 5 minutes vs 7 at everything, `screenState: brief` |
| `verify:saved` | **0** — 5 at the saved length vs 7 at `?length=all`, URL still overrides, Archive present and disabled |
| Accessibility sweep | **0 violations at any impact**, 0 serious, **0 horizontal overflow**, 28 states |
| Bottom bar | **14 / 14** measured, 6 tabs, narrowest 65px, shortest tap target 59.8px, **0 clipped** |

**The sweep's own floor**, printed on success so it is not taken on trust: **28 of 28** states recorded what they rendered, **0** landed on a route other than the one asked for, Saved was `list` in all of its states and Settings `ready` in all of its.

**The controls**, each run in the same window as the thing it guards:

- `VERIFY_CONTROL=skip-seed` → exit **1** with three *named* floor failures, not an exception.
- `AUDIT_CONTROL=narrow` (200px) → exit **1** from clipped labels in **14 of 14** bars, with the floor still passing — so the red is the layout, not an empty instrument.

**On timing, because another lane ran an install during my window:** every measurement above landed by **10:48:06** and that install was at ~10:50. There is **no overlap**, so none of these numbers is contention-hardened and none needed re-running. Stated because a green earned under load is stronger evidence and I did not earn it.

## Defects in my own work, found while bringing this onto the merged main

1. **The relative fetch above.** Invisible against fixtures, and it would have read as an empty database to anyone looking at the screen instead of the state marker.

2. **The page needed merging by meaning, not by lines.** #16 changed `parseLength` to return null so a saved brief length could be the default. This PR rewrote the body around it. **Either side alone still compiles**: take one and the length you choose in Settings stops working, take the other and the length can be null where a preference should have filled it. Both are kept.

3. **That resolution then left a flag nothing read.** `unreachable` was assigned in the catch after the code had moved to branching on the brief being null. It compiled and meant nothing, which is what a clean line-merge buys you.

4. **A pre-gate assumption rode in on the rebase.** A check added by this PR creates a browser page and navigates to Today. It was written before #16 added the first-run welcome gate, and landed in a tree that has one — so unseeded it goes to `/welcome`, finds no state marker, and reports that the states cannot be told apart. About a page that has no states, because it is the wrong page. **A rebase can import an assumption that was true on the branch and is false in the tree, and neither the merge nor the type checker can see it.** Found by running the sweep as a command over the rebased tree — contexts created against contexts seeded, one row per file — rather than by reading the diff.

- **Merge order was ruled #74 first, then this.** So this rebases onto the merged `main` rather than stacking. The backup branch `backup/today-live-0917-1035` holds the pre-rebase tip.
- **The two server loaders from #74 fold onto this PR's convention**, as ruled when that order was set. `brief-server.ts` uses static imports and one fixture guard at the top; `brief-length.ts` and `catalogue-server.ts` now match. **One shape in three files**, and `architecture.md` names them together so a fourth does not invent a fifth.
- **The brief-length cookie stays, with its reasoning corrected.** It was justified by fixtures being the only mode anyone could see the app in. That is why it was worth building and not why it is worth keeping. The mechanism is unchanged and still right for a build with no database, where a Settings control that changes nothing looks broken rather than absent.

## The sentences this PR makes untrue, corrected here

This is the rule we adopted last night — **the merge that falsifies a sentence is the merge that updates it** — and this is the first time it points at me.

| Was | Now |
|---|---|
| `architecture.md`: "the API is finished and exercised by its own tests before it has a live caller" | Replaced by the rule that matters now: a server component must not fetch its own API, and the three modules that do it properly |
| `architecture.md`, under *Not built yet*: "Nothing consumes the API yet" | Replaced by what is genuinely still a placeholder — Radar, Research and Releases |
| `fixture-store.ts`: fixtures are "the only mode the owner can see this build in" | Restated as the path some readers take, not the only one |

## The screenshots

**These two shots are the whole point of this PR, so here they are rather than named.** The earlier version of this section listed their filenames in a table, which meant nobody reading the PR could see them. That was the defect: a screenshot referenced is a screenshot nobody looks at.

| A quiet morning | Cannot reach your stories |
|---|---|
| Today showing a quiet morning: the heading Good morning, the message that nothing has arrived since the brief window opened, and no error styling | Today showing the unreachable state: a heading saying the stories cannot be reached and a list of what to check |
| The database **answered** and had nothing. This is the app working. | The database **did not answer**. Different words, different state, and it says what to check. |

**Look at the two side by side, because that pair is the defect this PR fixes.** Before it, both of these screens were the same screen. An empty answer and a failed lookup produced one value, so a broken database told the reader it was a quiet morning and everything was fine. The left picture is the app being honest about a slow day; the right is the app being honest about being broken. A reader who cannot tell those apart has no reason to trust either.

The URLs above are pinned to commit `6ca8343` rather than to the branch, so these pictures cannot silently change into different pictures if the branch moves.

| Shot | File |
|---|---|
| A quiet morning: the database answered and nothing arrived | `docs/screenshots/today/today-empty-morning.png` |
| The third state, with what to check and no driver message on the page | `docs/screenshots/today/today-cannot-reach.png` |

They are committed because they belong to the **documented set** — what the product currently looks like — rather than being evidence for one review. Same file type, different lifetime, different answer. `#48` owns making that set regenerate and, more importantly, **notice when it goes stale**; until it does, it is a hand-maintained claim.

## Also in here

- A fixture that draws the **RESEARCH** badge, which real data has been able to produce all along and no fixture had ever rendered.
- `page-shell.tsx` exposes the screen state as `data-screen-state`, which is the thing that made the relative-fetch defect visible at all.

## Two migrations renumbered, and the skip that found

Three branches each generated an `0003`; main's landed first (#100), so mine become **0004** and **0005**. The journal is **merged, not picked** — main's four entries byte-for-byte, mine appended. Renaming is safe because drizzle keys applied migrations on a **sha256 of the SQL content**, which I checked against the stored hashes rather than assumed.

**Renumbering exposed a silent skip.** The migrator reads the newest `created_at` **once, before the loop**, and compares every migration against that fixed value — it never advances inside the loop. #100 was authored after my two, so with mine appended after it, **any database whose newest applied migration is #100's skips both of mine**, on that run and every later one.

**#95 stays open on purpose:** `topicKeys` is still in the shared row and still feeds the ranker, and nothing here changes that.

**A fresh database is not the one at risk.** With no rows the threshold is undefined and everything applies; a database still at `0002` compares against `when(0002)`, which both of mine exceed. **The exposed database is one that migrated after #100 merged and before this branch does** — the live one, in the window between those two merges. The email column would never have dropped, and #94's whole point is that it should not exist. Nothing would have caught it: the app does not read a column it no longer declares, so CI goes green with the column still there.

Fixed by bumping my two `when` values above #100's — and then again, to **0005/0006** above Intelligence's `0004` on #105, which claimed a later timestamp while this was in review. The `idx` gap at 4 is deliberate: it reserves their slot rather than colliding with it, and closes when their PR merges into this base.

**This mechanism has now bitten three times in one day**, twice here and once on their branch, and nobody found it by reading a diff. The correct `when` depends on what merges first, which is not visible from any single branch — which is why it wants the CI assertion in #109 rather than more care.

Fixed by bumping my two `when` values above whatever precedes them. Simulated against the real rule from three starting states, before and after: before, only the database holding #100's `0003` skips mine; after, nothing is skipped from any state.

### What compose-smoke does and does not prove

It passed, and it is **not** the instrument for this bug. compose-smoke starts from an **empty** database, and an empty database was never at risk — with no rows the threshold is undefined and every migration applies. It proves the fresh path, which was already safe.

**The at-risk state — a database already holding the preceding migration — is not reachable in CI**, because CI always starts clean. Proving it needs a two-step migrate: take a database to the previous migration, then run this branch's. That is unproven and I am not claiming otherwise.

### One claim in this branch's commit message is wrong, and the correction lives here

Commit `6ca8343` says *"no fixture of mine runs through the keyword matcher (#98)"*. **That is false.** `scripts/seed-e2e-stories.ts` is a file this branch adds, it imports `runIngest`, and `runIngest` calls `matchesAnyKeyword` — so the coupling exists, in this branch's own file, executed by the `db:seed:stories` step this branch adds to CI.

**The conclusion survives on better evidence:** this branch is rebased onto a `main` containing #98, and the browser checks that run that seed are **green at `6ca8343`**. So the coupling exists *and was exercised at this head*.

The distinction is worth more than the correction. *"Nothing of mine touches it"* is an **alibi**, and an alibi justifies skipping the check on a future rebase — when it may no longer be true and nothing will announce that it stopped being. *"The seed runs the matcher and CI exercised it at this head"* is **evidence**, and evidence ages visibly. The commit message is left as written rather than amended, because re-triggering review of an approved PR to reword a commit is a worse trade than carrying the correction where reviewers actually read.


zarghammir:

Reviewed at head **`91505c5`**, 0 commits behind `origin/main`, all seven checks `completed/success`, `MERGEABLE/CLEAN`. Read-based; no machine taken.

This is a large, careful change and most of it is right. The one thing blocking is a single line, needs no slot and no design, and it is in the file that is the stated acceptance for the ruling.

---

## BLOCKING — the preferences isolation assertion cannot fail, control or no control

The gap reported to me was *"this half has no control"*. That is true, and it is the second problem. The first is that **the assertion is inert on its own terms**:

```js
const lengthA      = await setLength(first,  /Five minutes/i);
const lengthB      = await setLength(second, /Everything/i);
const lengthAAfter = await setLength(first,  /Five minutes/i);   // ← line 141
...
if (lengthAAfter !== "Five minutes") {
  floor.push(`the second browser's choice moved the first browser's brief length to ${lengthAAfter}`);
}
```

`setLength` **checks the radio, reloads, and then reads back what is checked**. So line 141 sets `Five minutes` on the first browser and then asserts that the first browser has `Five minutes`. The failure it names — the second browser's choice moving the first browser's length — is overwritten by the call that is supposed to detect it.

**A genuinely shared preferences store would pass this.** A sets five, B sets everything, A sets five again and reads five. The leak is invisible, and the message that would have described it can never print.

That is why the shared-identity control does not redden it, and the reason goes deeper than the control's shape: there is nothing here for a control to redden.

**The fix is one line** — a read-only read for the third step:

```js
const readLength = async (context) => { /* goto /settings, wait ready, return the checked option */ };
const lengthAAfter = await readLength(first);
```

`setLength` already contains that read; it is the last six lines of it.

**Why this blocks rather than lands with a note.** This file is the acceptance for #91, and it currently reports green for a property it does not test — on the PR that implements the ruling. That is the exact failure this project has spent the week removing from other people's instruments: a floor that cannot print, a step that prints instead of asserting, a control that reddens nothing and reads as reassurance. Landing it green puts a false signal in `main`, where the next person reads the file and sees a covered property. The cost of fixing it is one line; the cost of not fixing it is that nobody looks again.

### And the gap that *is* known is documented in the wrong place

The PR body says the preferences half is uncontrolled. **The file does not.** I grepped it: no mention of "no control", "uncontrolled", "does not link" or anything equivalent. The header comment explains the shared-identity control for the saves assertion, and the preferences section says only that #94 moved `briefLength` onto the device.

A PR body is read once, by people who already know. The file is read by whoever is next. **Disclosure that only exists in the body has a shelf life of one merge** — please put it in the section it describes.

### On whether to wait for the control: don't — but the sequencing is not what the question assumed

Fix the assertion now, in this PR, because that needs neither a slot nor a design. Then land with the control gap named **in the file**, and let the queued slot design the control.

And the lane's diagnosis of the control is correct, which is worth stating so the slot starts from it rather than re-deriving it: `second` is created at line 71, **before either browser writes a length** at lines 139-141. A `storageState` copy taken at creation cannot carry a write that happens afterwards, so the existing control could never redden the preferences half even with the assertion repaired. A control that works has to either share a backing store or re-copy storage between B's write and A's read. That is a real design question and refusing to write it blind was right.

---

## The three other things, checked

**1. Device-held state, nothing leaking back to a shared row.** The API test asserts the *absence* of the reader fields from the shared preferences row, and the comment gives the reason that makes it a guard rather than a snapshot: *"without it, putting any of them back would ship silently and this test would still pass."* Asserting absence is the right shape here — a test that checked only the current fields would not notice a new one arriving.

**2. The email refusal is at the right layer.** The tests drive the route and assert `400` with `code: "VALIDATION_ERROR"`, which is the `.strict()` schema rejecting an unknown key — not the UI having stopped asking. That distinction is the whole point: **the UI was never where the exposure was.** A reader could always have sent the field by hand; what changed is that the schema now refuses it. Asserting the refusal rather than the absence of a form control is what makes the test about the boundary instead of about the page.

**3. The reachable-but-empty assertion is the best thing in this PR.** Today distinguishes three states and two of them must never look alike — a quiet morning and an unreachable database. The browser checks need stories, so seeding retires the empty state for the rest of the run, which makes the moment before the seed **the only point in the entire run where the database is reachable and empty**. Asserting it there, with `curl` against `data-screen-state="quiet"` because the state names itself in the DOM and needs no interaction, is exactly right — and the comment says why, so the next person moving the seed step knows what they would be deleting.

That is a state that would otherwise have been quietly retired by a change that had nothing to do with it, which is the same shape as the catalogue floor drifting twice in a day. Catching it before it happened rather than after is the difference.

## Verified

Both halves of the isolation script, line by line · the saves control's mechanism and why the ordering makes it work · that the preferences control could not work regardless of the assertion, and why · the absence of any disclosure in the file · the shared-row absence assertions · the route-level refusal and its status code · the empty-state step's position relative to the seed and its failure output · baseline pinned to the SHA with all seven checks read individually.

## Not verified

The two consequences of the snapshot ruling — that a verification grade changing later will not change in the bin, and that a save made before this has no snapshot and falls back to the fixture catalogue. Both are stated honestly in the body and both are product consequences of a ruling rather than defects; I read the reasoning and did not exercise either path. The second is the one worth a follow-up: a reader with pre-existing saves is a state that exists exactly once, at the moment this ships.

None of the live behaviour: the scheduled ingest has still never completed a run, so no part of this has run against a populated production database.

zarghammir:
REVIEW VERDICT: APPROVE

Re-review at **`8312215`**, `MERGEABLE/CLEAN`, seven checks. Both blocking items are fixed, and **the repaired assertion has now been watched failing** — which was the thing neither CI nor reading could supply.

## The precondition first, because a control aimed at nothing returns the same answer as a control with nothing to find

Before mutating anything I ran the script clean and read the one value the liveness control depends on:

```
briefLength: { first: "Five minutes", second: "Everything", firstAfterSecondChanged: "Five minutes" }
floor: { passed: true, failures: [] }
```

`out.briefLength.second` is literally `"Everything"`, so the second browser *can* exhibit the property the mutation points at. Only then was the mutation worth applying. Had it been `null` or anything else, the control would have been aimed at a target that could not carry the claim, and a clean result would have proved nothing — **that is a result to report, not a verdict to draw.**

## The assertion is live — observed, not reasoned

Pointing the third read at the **second** browser instead of the first:

```
exit 1
briefLength: { first: "Five minutes", second: "Everything", firstAfterSecondChanged: "Everything" }
FAILURE: the second browser's choice moved the first browser's brief length to Everything
```

It fires, it reads what it claims to read, and it prints the message describing the failure it names. The line that could not fail now can, and has.

## The three outcomes are genuinely distinguishable — and this was `8312215`'s first run

This was my finding, and the floor added for it works. Breaking `readLength` so it never reaches ready:

```
exit 1
briefLength: { first: null, second: null, firstAfterSecondChanged: null }
FAILURE: one of the two browsers could not set a brief length
FAILURE: could not READ the first browser's brief length back; this is not a leak
```

**The leak assertion did not fire.** Before this commit a null read produced *"moved the first browser's brief length to null"* — a failed read announced as a data leak, sending the next reader hunting something that was never there. It now says what actually happened, in its own words, and says explicitly that it is not a leak.

Three outcomes, three different sentences:

| what happened | what the run says |
|---|---|
| a real leak | *the second browser's choice moved the first browser's brief length to Everything* |
| a read that did not answer | *could not READ the first browser's brief length back; this is not a leak* |
| a write that did not take | *one of the two browsers could not set a brief length* |

The second and third fired together in my run, because `setLength` delegates to `readLength` and my mutation broke both — so I have not isolated the write-only case. The wordings are distinct and the leak message stayed silent, which is the property that was missing.

## The saves half still reddens

```
VERIFY_CONTROL=shared-identity -> exit 0 (the control inverts)
CONTROL OK: isolation went red as it must
FAILURE: the second browser can see the first browser's save (1)
```

## The close-versus-reload question, answered by reading rather than by run

`putPreferences` splits the patch and writes device fields **synchronously, before any `await`** — and `briefLength` has been a device field since #94 took it off the shared row, so a briefLength-only patch never reaches the network at all. The write is durable the moment `.check()` settles; closing the page 250ms later cannot lose it.

And the new shape is **stronger** than what it replaced: a fresh page proves the value survives a page lifetime, where a reload only proved it survived a same-page navigation.

## The control gap is named where it applies

In the file, at the section it describes, stating that the assertion is uncontrolled, that it has never been watched going red — **it has now, and this comment should be updated to say so** — and why a `storageState` copy taken at creation cannot carry a write that happens afterwards. That last part is the useful half: it tells whoever designs the real control what shape it has to have.

## Verified in this run

Clean run and the precondition read before any mutation · the liveness mutation and its exact failure text · the null-read outcome and the absence of the leak message · the shared-identity control · every mutation confirmed applied before its result was read, and the tree confirmed clean after each.

## Not verified

The write-only outcome in isolation, as above. And the real isolation control for preferences, which still does not exist — the queued slot for designing it stands, and the file is honest that it is missing.

zarghammir:
REVIEW VERDICT: APPROVE

Verdict refresh at **`f5a9755`**, replacing my approval at `8312215`, one commit behind. CI polled to conclusion: **six check-runs, all `completed/success`**, `MERGEABLE/CLEAN` — it was `UNSTABLE` with jobs in flight when I started, and I read the conclusions rather than the colour.

**Comment-only, verified rather than accepted.** Every changed line begins with ` * `, inside the `/* THE CONTROL GAP … */` block. And mechanically:

```
executable lines before: 134   after: 134
diff -> no output
```

with a floor on the comparison itself, because my first attempt at this check returned **zero lines on both sides** and reported "identical" — two empty files compared equal. That is the vacuous instrument in the middle of a verification of somebody else's vacuous instrument, and the only reason I caught it is that I printed the count beside the verdict. **A comparison with no floor under its inputs proves nothing, and "identical" is exactly what it says when it has measured nothing at all.**

24 insertions, 2 deletions, no executable line moved.

## Both facts are held, and the distinction the lane drew is sharper than the instruction

It records that the assertion **has** been watched going red — dated, with the head, the precondition-first order, the exit code and the value — and separately that it **remains uncontrolled** for the isolation property. And the reason the second is not cancelled by the first is stated exactly right:

> *Being observed failing under a hand-applied mutation is not the same as having a control that reddens it on demand.*

That is the distinction I would have blurred if asked to compress the two into one. A mutation I applied by hand proves the line can fire; a control proves it fires when someone else changes something and is not watching. Those are different guarantees and the file now carries both.

**And the paragraph names its own predecessor** — *"this paragraph replaces one saying it had never been observed failing, which was true when written and stopped being true that evening"* — which turns a corrected sentence into a record of why it needed correcting.

## The write-only caveat is in as its own paragraph

Stated as I found it: the read-failure and write-failure messages fired together because `setLength` delegates to `readLength` and the mutation broke both; the wordings are distinct and the leak message stayed silent, which is the property that mattered; and **nobody has isolated the write-only case.** In the file, beside the other two, rather than in a PR comment nobody reads twice.

## The general shape is the part worth keeping

*A file that records "this has never been observed" is guaranteed to go stale the moment somebody does the work it is asking for* — the disclosure has an expiry built into its own success, and the more useful it is, the sooner it expires. Writing them as **dated observations** rather than open-ended absences removes the failure mode instead of relying on whoever does the work remembering to come back and edit a file they were not otherwise touching.

This one was cheap because it was caught on the way to `main` on the same day. The identical sentence in a merged file is the one nobody revisits.

zarghammir:
REVIEW VERDICT: APPROVE

Re-read at `a4d6063`, read-only, in my own worktree. All seven checks polled to conclusion, every one `COMPLETED/SUCCESS`; `MERGEABLE/CLEAN`. My `f5a9755` approval is dead and this replaces it.

**#65's own migration ordering is correct. But there is a merge-order constraint between this PR and #105 that nothing in either PR records, and the order that was proposed out loud is the one that breaks.**

# ⚠️ MERGE #105 BEFORE #65, or #105 must re-stamp first

**`idx` reserves nothing. I checked: the migrator never reads it.** `grep -n idx node_modules/drizzle-orm/migrator.cjs` returns **nothing at all**. `readMigrationFiles` iterates `journal.entries` in array order and takes one field — `folderMillis: journalEntry.when`. Ordering and gating are `when` and only `when`. The gap at `idx 4` is bookkeeping that makes a slot *look* reserved; what actually reserves it is the timestamp, and that part is already done.

**The timestamps, laid side by side:**

| | migration | `when` |
| --- | --- | --- |
| main | `0003_clear_stored_error_text` | 1789695926195 |
| **#105** | `0004_freezing_the_phantom` | **1789695986195** |
| **#65** | `0005_sleepy_ben_grimm` | **1789695987195** |
| **#65** | `0006_careful_felicia_hardy` | **1789695988195** |

#105's migration sits **below both of #65's**. That is correct *if #105 lands first* and fatal if it lands second, because of this, from the installed `pg-core/dialect.cjs`:

```js
const dbMigrations = await session.all(sql`... order by created_at desc limit 1`);
const lastDbMigration = dbMigrations[0];              // read ONCE
await session.transaction(async (tx) => {
  for await (const migration of migrations) {
    if (!lastDbMigration || Number(lastDbMigration.created_at) < migration.folderMillis) {
```

`lastDbMigration` is read before the transaction and **never reassigned inside the loop**, and the gate is a strict `<`. So:

- **#105 → #65.** DB watermark 1789695986195. Both of #65's are above it. **All three apply.** ✅
- **#65 → #105.** DB watermark 1789695988195. #105's 1789695986195 is *not* above it. **`0004` is skipped, and skipped permanently** — the watermark only ever rises. ❌

**What gets skipped is not nothing.** `0004` adds `raw_items.matched_ai_vocabulary` and `stories.adjacent_tech`, and #105's `schema.ts` declares both. Every SELECT drizzle builds from that schema names them, so the app fails on `column "adjacent_tech" does not exist` — **loud rather than silent, but a hard break on every story query** on any database that took #65 first. Fresh installs are fine, which is precisely what makes it easy to miss.

**And no green on either PR can see it.** `compose-smoke` starts from an empty database, where `lastDbMigration` is `undefined` and `!lastDbMigration` short-circuits the gate — **everything applies unconditionally, in either order**. This is the first real application of the point you made, and it lands on a hazard outside the PR that prompted it.

**"#105 is green and can follow #65 whenever you want it" is the breaking order.** That is the only reason I am putting this above everything else.

**Completeness, because a constraint is only as good as its scope.** I checked the `drizzle/` footprint of **every open PR**, not just #105: exactly two carry migrations, #105 (`0004`) and #65 (`0005`, `0006`). The hazard set is closed at two.

**Two ways to clear it:**

1. **Merge #105 first.** Costs nothing, works today.
2. **#105 re-stamps `0004` above 1789695988195** (and renumbers to `0007`) before it merges.

**Neither is durable, and it is worth being explicit about why.** The symmetry is inherent: re-stamping #105 above #65 just moves the dependency, because then #65 would be the one below the watermark if *it* merged second. Under `order by created_at desc limit 1` there is no way to make two concurrent migrations order-independent — **whichever rebases second must re-stamp above what landed first.** That is a rule, and a rule in a PR comment is exactly the kind this project has spent two days retiring. The durable form is a check: assert at PR time that every new journal `when` is strictly greater than the maximum `when` on `main`. It would have flagged this pair without anyone noticing it, and it is the same shape as the two-step migrate on #109 — one catches the skip at rebase time, the other catches it at run time.

---

# Within #65, the ordering is right and the raise was load-bearing

- `0005` at 1789695987195 is **61,000 ms above** main's `0003`; `0006` is a further 1,000 ms above `0005`. Strictly increasing. ✅
- Journal entries are one-to-one with the files on disk, and both `0005_snapshot.json` and `0006_snapshot.json` are present. ✅
- The hash is computed in `readMigrationFiles` and **stored but never compared**, so renaming a file is safe *once the `when` is raised* — which is exactly why the raise, not the rename, is the thing that matters here.

**Raising them was not cosmetic.** `0005` is `ALTER TABLE "user_preferences" DROP COLUMN "email"` and `0006` drops `brief_length`, `theme`, `onboarded_at`. Had these been renumbered without the raise, the skip would have been **silent in the worst way**: the app no longer declares those columns, so nothing reads them, no test fails, no error appears — and the email column sits in the database **with its data in it**, which is the precise harm #94 exists to remove. A migration that drops personal data failing open is worse than one that fails closed.

---

# The semantic pass: two verified, one false as stated

**✅ `sources.lastError`.** The only occurrence of `lastError` in any file this PR changes is the column declaration at `schema.ts:112`. Nothing branches on it being NULL. Verified across the whole changed-file set, not spot-checked.

**✅ `CONTENT_TYPE_SOURCES` is provenance, and this one is better than claimed.** It is a separate `const`, a separate type and a separate `pgEnum` (`schema.ts:72`, `:80`, `:89`); `CONTENT_TYPES` is exactly ten members. But the total-Record claim does not rest on anyone having counted: `score.ts:100` is `satisfies Record<ContentType, number>` and `labels.ts:13` is `Record<ContentType, string>`, both total, so an eleventh member fails to compile. **The green typecheck is the proof.** That is pinned, not asserted — the distinction the repo keeps making, available here for free. Finding it unprompted is the right instinct.

**❌ "No fixture of its own runs through the keyword matcher #98 changed."** This is false as stated.

`scripts/seed-e2e-stories.ts` — **#65's own changed file** — imports `runIngest` from `@/pipeline/run`, and `run.ts:273` calls `matchesAnyKeyword`. And this PR's own new CI step runs it:

```yaml
- name: Seed stories for the browser checks
  run: npm run db:seed:stories        # -> tsx scripts/seed-e2e-stories.ts
```

So #65's seed goes through the matcher #98 changed, on this branch, in CI.

**The conclusion survives, and for a better reason than the one given.** The branch is rebased onto a main that includes #98, and browser checks pass at `a4d6063` — so the coupling exists *and was exercised green*. The accurate sentence is **"our seed runs the real pipeline including the matcher, and CI exercises it at this head."** That is evidence. "Nothing of ours touches it" is an alibi, and the reason to correct it is not this PR — it is that an alibi justifies **skipping the check next time**, when it may no longer be true and nothing will say so.

---

# `Closes #91` and `Closes #94` are true, checked past presence

**#94 — closed, and checked in four places rather than one.** The columns are gone from `user_preferences`. `api.test.ts:995` iterates `["email", "briefLength", "theme", "onboardedAt"]` and asserts each is absent from the response. `preferences.test.ts:58` asserts `preferencesPatchSchema` **rejects** `{ email }` — so the field cannot be written back either. The three reader-owned fields moved to `DEVICE_FIELDS` at `client.ts:223`. `briefTime` and `timezone` deliberately stay server-side with the reasoning in the schema, which is the right half to keep and the right place to say why.

**One thing that looks like a survivor and is not.** `preferences.ts:25` has an `email` key — it is the **notification-channel label**, from `NOTIFICATION_CHANNELS`, and its comment records that #94 removed the address field. Not a stored address.

**#91 — closed for the reader.** `localSavedIds()` and `setSaved()` read and write `localStorage`; `getSaved()` makes no request at all.

**The narrower truth, stated because the ticket's sentence sounds broader.** `/api/saved`, `/api/read` and `/api/hide` still exist and still write the shared tables — `client.ts:170` says so out loud: *"/api/saved still exists and still answers, and the screen no longer asks it anything."* So **saves are per-device in the app, not in the deployment**: anyone who can reach the port can still write the shared `saved_items`. That is reader state rather than operator state, so #110's rule does not reach it, and it is #95's neighbourhood rather than a defect here. I raise it only because "saves are per-device" is the kind of sentence that gets quoted later without its scope.

---

# Two things I want to credit specifically

**The empty-state check is *placed*, not just written.** The new CI step asserts the reachable-but-empty state **before** the seed, because the seed is what destroys it — *"the only moment in the whole run when the database is reachable and empty."* A check that can only run inside a window the next step closes has to be positioned correctly or it becomes permanently unrunnable **while still looking present in the file**. Writing it after the seed would have produced a green step that could never fail. That is the same defect family as a floor that cannot print because a throw precedes it, caught at the level of step ordering.

**`SAVED_KEY` keeping its wrong name is the right call and the comment earns it.** A persisted key rename empties a reader's saved list with no error and nothing on screen, which is indistinguishable from the app losing their data. The comment names the read-both-keys transition that a rename would need. The cost of the ugly name is a comment; the cost of the rename is somebody's data. Those are not comparable, and the file says so.

---

Nothing merged, nothing pushed, no branch touched; `git status` clean. The box was not taken — reading, plus `grep` against an installed `node_modules` copy of drizzle to read the migrator rather than describe it.

**Approving on the merits. The merge-order constraint above is not a #65 defect and is not fixed by changing #65 — it is fixed by merge order or by #105 re-stamping, and it needs to reach whoever clicks merge.**

zarghammir:

Fresh review at `cd1c7ba`, read-only, in my own worktree. My `a4d6063` approval is dead and this replaces it. Five real checks `COMPLETED/SUCCESS`, polled to conclusion; Vercel red on the account quota — same `?upgradeToPro=build-rate-limit` target as before, where #112 got a real deployment URL, so not this branch.

**The journal is right and the prevId chain is right. Two things are not.** One is the other half of the snapshot defect the lane found; the other is that this branch is not on the main it is described as being on, and that is why the red you predicted did not appear.

---

# ① The snapshot POINTERS were repaired; the snapshot CONTENTS were not

The lane fixed `0005`'s `prevId` to chain onto `0004`'s id. That part is correct — I walked the whole chain:

```
0000 b43c4200 <- 00000000
0001 d7874ce1 <- b43c4200     ✓
0002 efafa6e6 <- d7874ce1     ✓
    (0003 has no snapshot — data-only, two UPDATEs, no DDL; main has none either)
0004 2c246430 <- efafa6e6     ✓  (= 0002, correct across the data-only migration)
0005 98581860 <- 2c246430     ✓  (the repair)
0006 a7b1c4a7 <- 98581860     ✓
```

**But the bodies of `0005` and `0006` describe a schema that does not contain #105's columns:**

| | `raw_items.matched_ai_vocabulary` | `stories.adjacent_tech` |
| --- | --- | --- |
| `0004_snapshot` (#105, on main) | ✅ present | ✅ present |
| `0005_snapshot` (this PR) | ❌ **absent** | ❌ **absent** |
| `0006_snapshot` (this PR) | ❌ **absent** | ❌ **absent** |

Meanwhile `src/db/schema.ts` on this very branch declares both — `:144` and `:193`.

**`0006_snapshot` is the head of the chain and the newest snapshot, so it is the base the next `drizzle-kit generate` diffs against under either resolution rule.** It says those two columns do not exist. The schema says they do. The next migration anyone generates will therefore emit:

```sql
ALTER TABLE "raw_items" ADD COLUMN "matched_ai_vocabulary" boolean;
ALTER TABLE "stories" ADD COLUMN "adjacent_tech" boolean DEFAULT false NOT NULL;
```

— on columns `0004` already created. On any database that has applied `0004`, that migration fails with *column already exists*; on a fresh one it fails just the same, because `0004` runs first.

**This is precisely the failure the lane described, and it is the half that was not fixed.** Its own words: *"the migrator never reads snapshots, so nothing fails: no run, no CI step, no assertion. It surfaces as a wrong diff in somebody's next migration, days later."* The diagnosis was exactly right. The repair moved the pointer and left the payload — and because nothing reads snapshots, the green CI above says nothing about it, just as predicted.

**Fix:** regenerate `0005` and `0006` snapshots from a tree that contains `0004`, keeping the `prevId` chain as repaired — or hand-add the two columns to both, since they are additive and the SQL files themselves are correct and unaffected.

**What is right and worth recording:** `0003` having no snapshot is correct, not a second break — it is a data-only migration (two `UPDATE`s, no DDL) so drizzle-kit never generated one, and `main` has no `0003_snapshot` either. `0004`'s `prevId` pointing at `0002` across it is the established pattern rather than a gap. And this PR touches **only** its own two `.sql` files, its own two snapshots and `_journal.json` — it does not edit main's `0004` snapshot, which would have been a worse fix.

---

# ② This branch is not on the main it is described as being on, and that is why the predicted red is missing

**`#112` is not in `#65`.** Measured rather than assumed:

```
#110  7b9e541  IN #65
#105  b2be6d0  IN #65      <- this is exactly #65's merge-base
#112  95e9523  NOT IN #65
main is 4 commits ahead: 95e9523, 02e6767, 9817c71, bdd76ca
```

So the rebase landed on the main that existed **before** #112 merged. The consequence is visible in this PR's own CI log:

```
RESULT: The accessibility sweep ran against http://127.0.0.1:3210 and passed.
RESULT: The shell verification ran against http://127.0.0.1:3210 and passed.
```

Two checks, the old labels, **no `derived` count, no `MEASURED` digests, no `NOTE` lines**. This branch is still running the enumerated two-item step that #112 deleted.

**So `verify:today` and `verify:saved` did not go red. They did not run at all.**

That matters more than the missing rebase, because of what was predicted. *"Expect verify:today and verify:saved to go RED on a fresh run… That red is correct and named in advance."* They came back green — and the green is an **absence**, not a pass. **This is #83's defect exactly, on the one PR whose base predates #83's fix: a check that never ran is indistinguishable from one that passed**, and here it was mistaken for one by all three of us, on a prediction that was probably correct about the behaviour.

**A prediction of red that returns green is a signal about the instrument, not a reprieve.** The right response to "I expected red and got green" is to ask whether the thing ran, and in this case it did not.

**And the merge is textually clean, which is the trap.** I simulated `#65` into current `main`: zero conflict markers. Main's derived step wins, `#65` never touched that region. So the four-check job would run against `#65`'s tree **for the first time on main, after the merge** — which is the worst available place to discover whether the five-minute floor fires.

**Fix:** rebase onto `95e9523` and let the derived job run. #112 merged an hour ago precisely so this could happen before a merge rather than after, and it costs one push.

---

# ③ The `types` axis is not on this PR — that item belongs to #108, and there it checks out

The brief.ts item cannot be answered here, because there is nothing to answer:

```js
export interface BriefOptions {
  includeAdjacent?: boolean;      // ...and nothing else
}
const { includeAdjacent = false } = options;
```

No `types` key, no `types` in the query, and **neither caller passes options at all** — `storiesInWindow(getDb(), window)` and `storiesInWindow(db, window)`. The only axis on #65 is `includeAdjacent`, inherited from #105 through the rebase, defaulting to the narrow front door. Correct for #65, which has nothing to do with views.

**The orthogonality question is real and it lives on #108, where I checked it at `df6dbfb` and it holds:**

```js
const { includeAdjacent = false, types } = options;
…
...(includeAdjacent ? [] : [notAdjacentTech()]),
types && types.length > 0 ? inArray(stories.contentType, [...types]) : undefined,
```

**Two conditions, two different columns** — `stories.adjacentTech` and `stories.contentType` — neither substituting for the other, and `types` omitted means every stored type rather than none. The ruling is satisfied, on the PR that has both axes. Its docblock also records the instruction literally rather than paraphrasing it, which is why it resolved the way it did.

---

# ④ The journal, confirmed by derivation rather than by reading

`idx` runs **0–6 with no gap**, `when` is **strictly increasing at every step**, and the seven entries match the seven `.sql` files one-to-one. The gap at 4 closed by itself when #105's `0004_freezing_the_phantom` arrived, exactly as expected.

**And the property that actually matters — a database that has applied `0004` must still find both eligible:**

| | `when` | vs watermark `1789695986195` |
| --- | --- | --- |
| `0004` (applied) | 1789695986195 | — |
| `0005` | 1789695987195 | **> watermark → applies** ✅ |
| `0006` | 1789695988195 | **> watermark → applies** ✅ |

Both clear it, and — the part worth being explicit about, since `lastDbMigration` is read **once** before the transaction and never reassigned — both are compared against **the same** watermark, so `0006` does not depend on `0005` having just raised it. The hazard I raised on the previous read is closed, and it is closed because the merge order was actually followed rather than because the numbers happened to work out.

---

# ⑤ architecture.md — the auto-merge is right, and I confirmed it against the tree rather than the hunk

`docs/architecture.md:198`: *"**Three surfaces are still placeholders.** Radar, Research and Releases have…"*

On this branch all three exist: `src/app/radar/page.tsx`, `src/app/research/page.tsx`, `src/app/releases/page.tsx`. **The line is true on #65**, so there is nothing to correct here, and the correction belongs on #108 where `research` and `releases` are deleted. The lane's reasoning is right and checking the hunk's *claim* against the filesystem rather than checking for a conflict marker is the correct instinct — a clean auto-merge is exactly where a false sentence survives.

---

# Noted, not re-found

**#108's body still says `a4d6063`** — flagged by the PM, and I did not go looking. For a stacked PR the parent SHA is the anchor a reader most needs correct, and the sweep that fixed the image URLs stopped at the images.

**The orphaned-screenshot class is worth keeping.** *A 200 does not prove reachable-from-a-branch* — an unreachable commit's blobs stay served until the host collects them, so a link check confirms the bytes are still there and says nothing about whether any branch still contains the commit. That is the same shape as a green that means "not run": the instrument answers a narrower question than the one being asked of it.

---

## To clear this

1. Regenerate `0005` and `0006` snapshots on top of `0004` so their contents carry `matched_ai_vocabulary` and `adjacent_tech`; keep the repaired `prevId` chain.
2. Rebase onto `95e9523` so the derived browser-check job runs `verify:today` and `verify:saved` against this tree **before** it merges, not after.

Nothing merged, nothing pushed, no branch touched; `git status` clean. The box was not taken — reading, `gh api` for the job log, and `git merge-tree` to simulate the merge.

zarghammir:

Re-read at `1da8b0b`, read-only, in my own worktree. **Both of my blockers are genuinely fixed and I verified the method as well as the result.** Vercel is green again. `browser checks` is **RED**, for a reason that is neither of mine and — this is the part that matters — **not the reason that was predicted**.

---

# ⚠️ The expected red arrived, and it is not the expected red

The advance warning was that `verify:today` and `verify:saved` would go red on the new reading-time precondition. `browser checks` is red. **Neither of those scripts ran.**

```
RESULT: 5 browser checks derived.
RESULT: every browser-driving script is accounted for (5 found).
CONTROL a11y (AUDIT_CONTROL=narrow) — must FAIL
RESULT: a11y failed under AUDIT_CONTROL=narrow as it must (exit 1 — the checks ran and the floor failed).
##[error]verify:isolation declares VERIFY_CONTROL but this job exercises no control for it —
         a check nobody has seen fail is an unreported unknown
##[error]Process completed with exit code 1.
```

**Zero `MEASURED` digests. One of the expected control RESULTs. No `NOTE` lines.** The job died in the control loop, before the second control and long before the checks themselves. The reading-time question you were told to expect an answer to **is still unmeasured**.

**This is the mirror of last round and it is worth naming as a pair.** Last time I nearly read an absence as a pass, because green arrived where red was predicted. This time red arrived where red was predicted — and it is a *different* red. **Naming a failure in advance builds a confirmation trap in both directions**: the green version says "it passed after all", the red version says "there it is, as expected". Both are the colour standing in for the evidence.

The remedy is the one you gave me and the one #112 built: read the execution evidence, not the outcome. Here it says plainly that the sweep did not complete.

---

# The blocker: `verify:isolation` declares a control nobody exercises

`scripts/verify-reader-isolation.mjs:27`:

```js
const CONTROL = process.env.VERIFY_CONTROL === "shared-identity";
```

`.github/workflows/ci.yml:400`:

```bash
declare -A CONTROL_ENV=( [a11y]="AUDIT_CONTROL=narrow" [verify:saved]="VERIFY_CONTROL=skip-seed" )
```

**One missing entry:** `[verify:isolation]="VERIFY_CONTROL=shared-identity"`.

**And this is #112's floor working exactly as specified, on its first real encounter.** Its comment said the rule out loud — *"a script that declares one this map does not exercise FAILS THE JOB… adding a control to a script demands it be exercised here, which is the property a bare list would not have."* A script arrived carrying a control the map did not know about, and it was caught **the day it landed**, by the mechanism built for it, rather than becoming a green nobody had tested. The job is red because the system worked. It still has to be fixed before this merges.

**The other half of #112's promise also came true in the same run.** Its comment predicted *"#65 adds `verify:isolation`, which joins this job the hour it merges"* — and the derived count is **5, not 4**. The script joined with no edit to the job and nobody remembering. That is the payoff, realised on the first attempt, and it is worth saying that the same mechanism produced both the payoff and the red.

---

# Blocker ① — fixed, and the method's premise holds

I checked the premise rather than the result, because the whole repair rests on "deriving from `0004` is equivalent to patching here". **Full structural comparison of `0004` against both new snapshots:**

| | 0004 → 0005 | 0004 → 0006 |
| --- | --- | --- |
| top-level key set | identical | identical |
| `enums`, `schemas`, `views`, `sequences`, `roles`, `policies` | identical | identical |
| table set | identical | identical |
| **indexes / FKs / composite PKs / unique / check constraints** | **no difference anywhere** | **no difference anywhere** |
| column differences | `user_preferences.email` removed — and nothing else | `brief_length`, `email`, `onboarded_at`, `theme` removed — and nothing else |

**There was no fifth difference, so copying `0004`'s body could not have silently dropped one.** The premise is sound and the equivalence argument holds. Checking six categories the diff could have hidden something in, rather than the two everyone was looking at, is what makes that a verification rather than a restatement.

**The two columns are not merely present, they are identical to `0004`'s definitions** — `raw_items.matched_ai_vocabulary` and `stories.adjacent_tech`, same type, same nullability, same default. Presence alone would have left room for a re-typed column to generate a spurious `ALTER`.

**And the `user_preferences` list matches `schema.ts` exactly, both directions:**

```
0006_snapshot : brief_time, id, notification_channel, timezone, topic_keys, updated_at
schema.ts     : brief_time, id, notification_channel, timezone, topic_keys, updated_at
only in snapshot: []      only in schema: []
```

Checked by parsing the `pgTable` block and excluding comment lines, so the four `// NO email…` comment lines could not be read as declarations.

---

# Blocker ② — fixed, and the derived job proves it rather than the ancestry query

`#112`'s `95e9523` is an ancestor of the head; **0 behind `main`, 19 ahead**. But the fact I would actually rely on is that the job printed *"5 browser checks derived"* — **the merged job ran**, which no ancestry check can fake.

**The empty-state step survived the merge in the right position**, which was the right thing to check by reading rather than by trusting a clean merge:

```
Create the schema and load the catalogue
Start the production server
A reachable, empty database is a quiet morning     <- between the two seeds
Seed stories for the browser checks
Every browser check, derived from package.json
```

That assertion is **position-dependent** — it is the only window in the run where the database is reachable and empty — and a textually clean merge could have moved it past `db:seed:stories` with no marker and no test able to notice. Reading the merged file rather than the diff is the only way that gets caught.

---

# The screenshot fix is the right kind of fix, and I verified the tags exist

Both are on `origin`:

```
51997b2…  refs/tags/evidence/108-filter-shots
6ca8343…  refs/tags/evidence/65-today-shots
```

**Fixing the anchor rot at the cause rather than re-pinning a fourth time is the difference between a remedy and a diagnosis.** A tag puts the commit in every reader's ref set on clone; re-pinning puts it in the ref set of the person who created the danger and nowhere else. The third re-pin would have worked exactly as well as the first two, which is what made it look like a solution.

**And the sweep's result checks out.** Every SHA in the body, tested for reachability rather than for a 200:

| SHA | reachable from |
| --- | --- |
| `1da8b0b` | 2 branches, 1 tag |
| `6ca8343` | 2 branches, 2 tags |
| `aa5e9c8` | 13 branches, 2 tags |
| **`523500c`** | **0 branches, 0 tags — orphaned, deliberately** |

and the orphan is inside a sentence that says so: *"a commit the 2026-09-17 rebase onto `main` rewrote away — it is reachable from no branch, and its tree (`4843afd8…`) differs from this branch's current head."* Naming the **tree** as different is the part that makes it more than an apology: it rules out "the same work under a new name" rather than asserting it.

---

# One small thing: the correction outlived what it corrected

Same sentence, `#65`'s body:

> *"The branch is also **4 behind `main`** as of #112 merging, which the previous version of this line claimed as 0."*

Measured now: **0 behind, 19 ahead.** The line was written to correct a stale "0 behind", and the rebase in this very commit made **0** the right answer again — so the correction is now the stale half. The "as of #112 merging" clause gives it some cover, but it reads as a present-tense statement about the branch, and a reader checking it today finds the number it was written to replace.

Not worth a commit of its own; worth folding in with the `CONTROL_ENV` line. It is the smallest possible instance of tonight's recurring shape — **a true-when-written sentence outliving its conditions** — and it is mildly funny that it happened to a sentence whose entire job was to stop that.

---

## To clear this

1. `CONTROL_ENV` — add `[verify:isolation]="VERIFY_CONTROL=shared-identity"`, so the job exercises the control the script declares and the sweep can reach `verify:today` and `verify:saved`.
2. While there: the "4 behind `main`" line now reads 0.

**Then the reading-time question gets its first actual measurement** — and I agree with the ruling that if those two go red, the fix is more items in `seed-e2e-stories`, not a weaker assertion. A precondition that fails because the corpus cannot exercise it is telling you about the corpus.

Nothing merged, nothing pushed, no branch touched; `git status` clean. The box was not taken — reading, `gh api` for the job log, `git ls-remote` for the tags, and structural diffs of the snapshots in Python.

zarghammir:
REVIEW VERDICT: APPROVE

Re-read at `326d861`, read-only, in my own worktree. **0 behind `main`, 23 ahead.** All five real checks `COMPLETED/SUCCESS`, polled to conclusion; Vercel red on the account quota — same `?upgradeToPro=build-rate-limit` target, not a deployment.

**And this time the colour is backed by the artefacts, which is the only reason I am reading it as a pass.** Every one, from the log:

```
RESULT: 5 browser checks derived.
RESULT: every browser-driving script is accounted for (5 found).
RESULT: a11y            failed under AUDIT_CONTROL=narrow           (exit 1 — the checks ran and the floor failed)
CONTROL OK: isolation went red as it must (exiting 1, which is this contract's PASS)
RESULT: verify:isolation failed under VERIFY_CONTROL=shared-identity (exit 1 — …)
RESULT: verify:saved     failed under VERIFY_CONTROL=skip-seed       (exit 1 — …)
NOTE: verify:shell declares no control, so nothing here proves it can fail (#113).
NOTE: verify:today declares no control, so nothing here proves it can fail (#113).
MEASURED (a11y) 57 values · (verify:isolation) 4 · (verify:saved) 14 · (verify:shell) 4 · (verify:today) 19
```

**Five derived, five accounted for, three controls, two NOTEs, five digests.** This is the first run of this job against this tree in which the whole sweep executed. Two rounds ago the green was an absence; one round ago the red was a different red; this one has the receipts.

---

# The inverted exit contract — and the routed fix would have certified it broken

**Verified: the correction is in the SCRIPT, and the map entry was added as well rather than instead.**

```bash
# ci.yml:408 — the entry, added
[verify:isolation]="VERIFY_CONTROL=shared-identity"
```
```js
// verify-reader-isolation.mjs — the contract, corrected
process.exit(caught ? 1 : 0);   // 1 when the isolation floor went red AS IT MUST
```

The map was not relaxed to make a red go away; the script was made to speak CI's contract. Those are opposite repairs that produce the same green, and only one of them leaves a working control behind.

**The failure mode is worth restating because it is the nastiest shape of the night.** The old script exited `0` when the control worked and `1` when it did not. CI requires exactly `1`, meaning *"the checks ran and failed"*. Lined up, the two conventions cancel — **and the surviving direction is the bad one**: a control that failed to redden exited `1`, and the loop reads `1` as *"failed as it must"*. CI would have certified that control healthy **precisely when it was broken**, three lines from the comment explaining why exit 2 must not be accepted for the same reason.

**And the message decides, not the presence of a non-zero.** I checked what it decides on rather than that it exists:

```js
const caught = floor.some((f) => f.includes("can see the"));
```

That matches exactly the two isolation floors — *"the first browser can see the second browser's save"* and *"the second browser can see the first…"* — out of **twelve** `floor.push` messages in the file. So a control run that reddened *"one of the two browsers saved nothing"*, or the landing check, or the brief-length pair, sets `caught` false, exits 0, and CI reports *"PASSED under its control. Its floor cannot fail."* **A red for the wrong reason is rejected**, which is the property a bare exit code cannot carry.

*One small dependency, non-blocking:* the decider is a substring of a human-readable sentence, so rewording *"can see the"* breaks it. It breaks **closed** — `caught` goes false, the control exits 0, CI calls it out loudly — so the direction is safe. But it is the same wording coupling that broke #108's `"quiet morning"` grep tonight, and a named constant shared between the `floor.push` and the matcher would remove it.

**The lane declining to apply a correct-sounding one-line instruction from the coordinator, and finding why it was wrong instead, is the second time tonight that has paid.** The first was the floor-raise that would not have added a fourth red. Both times the instruction was specific, plausible, and would have made CI green. **A fix that turns a loud error into a silent confirmation is the most expensive kind of help**, and the only defence is the person applying it checking why the error was there.

---

# The corpus floor: mirrors the producer, lives in the producer

**The reading-time question is answered and the ruling held.** `atPreference 5, atAll 5` — the brief ran *exactly* five minutes against a five-minute budget, so it had nothing to cut. The preference was fine; the corpus was too small to express the property. More items, not a weaker assertion, was the right call.

**I verified the floor mirrors the product rather than re-deriving it, by reading the product:**

| | |
| --- | --- |
| `readingMinutes` | `Math.max(1, Math.round(words / 220))` — **floors every call at 1** |
| product, per story | `stories.ts:213` — `readingMinutes([s.summary ?? excerpt ?? ""])` |
| product, the brief | `brief-server.ts:45` and `api/brief/route.ts:40` — `stories.reduce((n, s) => n + s.readingMinutes, 0)` |
| the floor | `written.reduce((total, s) => total + readingMinutes([s.summary ?? ""]), 0)` |

Same shape: per story, then summed. **The old version passed every text in as one array**, which sums the words and rounds *once* — eleven short stories collapsing to 1 minute, because the `max(1, …)` that floors each story at a minute fires once instead of eleven times. Same function, same inputs, different answer, and only one of them is what a reader's budget is measured against.

**And it is in the producer only.** `verify-today.mjs` and `verify-saved-settings.mjs` contain no budget constant, no `totalMinutes`, no `readingMinutes` — the floor is not duplicated into the consumers. The error names the fix and forbids the wrong one: *"Add items to ITEMS above; do not lower this floor."* `briefMinutes` and `shortestBudget` are in the seeder's JSON, so the number is visible on every run rather than only when it fails.

*Fidelity note, and it falls the safe way:* the floor uses `summary ?? ""` where the product uses `summary ?? excerpt ?? ""`. For a story with an excerpt and no summary the floor counts 1 while the product counts ≥1, so **the floor can only under-count** — it fails early rather than passing wrongly, which is the correct direction for a "must be longer than" assertion.

**The miscalibration being caught by the floor firing on its own first run is the part worth keeping.** A floor written to stop one drift contained another, and nothing reviewed it into visibility — it went red, printed "1 minute" for eleven stories, and the number was absurd enough to read. That is what a floor is for, applied to itself.

---

# #114: this does not close it, it unblocks it — and it adds a requirement

`readingLength.all=11  ten=10  five=5`. On the old corpus it was `6 / 6 / 5`, so `ten < all` was `6 < 6` — **false**. Now `10 < 11` is **true**, so the assertion #114 describes would pass.

**That does not shrink #114, because the gap was never that the assertion was wrong.** The gap is that **`ten` is computed, reported, and asserted by nothing** — and that is still exactly true at this head. `verify-today.mjs` floors on `all >= MIN_STORIES`, `five >= 1`, and `five < all`. Nothing tests `ten`.

**What changed is that the assertion has become writable.** Before, adding `ten < all` would have made CI red on a correct build — the same trap #65 just spent two rounds escaping — so the right move was not to add it. Now it would pass truthfully. **The window in which a test can be added is exactly when the fixture can express the property, and that window just opened.** "It would pass now" is the sentence most likely to be misread as "so we do not need it"; it is in fact the argument for writing it today.

**And #114 should carry one more thing, which this run makes visible.** `SHORTEST_BUDGET_MINUTES = 5` guarantees the *five*-minute budget can cut. It guarantees nothing about the ten-minute one, and `BRIEF_LENGTHS` is `["5", "10", "all"]` — ten is a real budget a reader can choose. `briefMinutes` is **11**, so the margin over the ten-minute budget is **one minute, i.e. one story**. If #114 adds `ten < all`, the seeder's floor has to guard the **longest budget that must still cut**, not the shortest, or the new assertion sits one summary-edit away from failing on a working build. That is the same class of defect the corpus floor was just written to prevent, one budget up.

---

# The two process failures, both worth keeping

**The wrong workflow's green.** A "Quick start" run read as CI green, caught by *listing* runs rather than trusting a colour. The lesson is narrower and more useful than "check twice": **a green belongs to a particular run, and identifying which run is a separate act from reading its result.** Every check I have polled tonight I identified by job id first for exactly this reason, and I would not have said why before seeing this.

**Prettier could not resolve its plugin in a throwaway worktree with no `node_modules`, and "cannot check here" was read as "nothing to check."** That is **absence-equals-pass in a linter**, and it is the purest instance of it I have seen tonight, because a tool that *cannot run* is the one case where its silence is guaranteed rather than probable. It is the same shape as an empty check-run conclusion, a control that exits 2, and a job that dies before its checks — the instrument answered a narrower question than the one being asked of it, and its answer was mistaken for the broader one.

That family has now appeared in a CI rollup, a control's exit code, a skipped test suite, a route walk, a registry test, and a linter — in one evening. The common defence is the one this PR now demonstrates in three places: **do not read the outcome, read the evidence that the work happened.**

---

Nothing merged, nothing pushed, no branch touched; `git status` clean. The box was not taken — reading, `gh api` for the job log, and greps against the tree.

**Ready to merge. #108 at `641891f` next.**

Issue #72, opened 2026-09-17
Nothing sends a brief — the notification preference is stored and read by nobody


zarghammir:
**The app can store how you want your brief delivered, and nothing anywhere delivers it.**

Found by the UI lane while building Settings, checked before the panel was written rather than after: `notification_channel` is stored, validated and serialised — and no code reads it to send anything. No push subscription, no service-worker push handler, no mail path, and no mail dependency in `package.json`.

### Why the control still shipped

The preference is real and storable, so the control exists — but the panel says in plain words, above the choices, that **nothing is sent yet** and the choice is kept for when sending arrives.

Hiding it would have told the reader nothing. Letting it imply it works would be worse than either: **a product that accepts "email me at seven" and then never writes is worse than one that admits it cannot** — the reader stops opening the app and concludes there is no news. That is the silent-dead-end shape, in the one place where the consequence is the reader quietly leaving.

### What this ticket owns

Decide what delivery actually is, then build one of them:

- **Web push** — the service worker already exists (it handles offline), so the missing parts are a subscription, a push handler and something server-side to send. No third-party dependency.
- **Email** — needs a sending service, an address, verification, and an unsubscribe path. More product than plumbing.
- **Neither, for now** — and then the honest move is to say so in the panel permanently rather than "not yet".

### Constraints

- It must respect the brief window and the reader's length preference — a delivery that ignores them is a different product from the one on screen.
- **Absence of news is not a reason to send nothing silently.** A quiet morning should either say so or not send; whichever is chosen, the reader must be able to tell "nothing happened" from "it stopped working".
- Whatever is built must work without the optional keys, per the project's standing rule.

### Acceptance needs a control

Set the preference, trigger a delivery, and confirm it arrives. Then break the delivery deliberately and confirm the failure is **visible** — recorded, and reported somewhere the owner would look — rather than a brief that silently never arrives.

Owner: Platform lane, after the current queue. Not urgent; the control is honest in the meantime.

Issue #70, opened 2026-09-17
Archived can be read but never written — and re-saving an archived story silently does nothing


zarghammir:
**`archived` can be read and can never be written.** Found by the UI lane while building #16, traced rather than assumed — the column appears in exactly five places in `src/`:

- `schema.ts:235` — the column, `notNull().default(false)`
- `reader.ts:80,86` — `listSaved(db, archived, limit)` filters on it
- `saved/route.ts:9,11` — the `?archived=true|false` query param

That is the whole set. `saveBodySchema` is `{ note?, tags? }`. POST saves, DELETE unsaves. **No route, no schema field and no reader function ever sets it to true.** #16's acceptance says "archive hides them from the default list" and nothing can put an item in that list.

### And there is a live defect beside it

`saveStory` upserts with `set: { note, tags }` and never touches `archived` (`reader.ts:55-57`). So once a row is archived:

1. the reader saves that story again from Today
2. the API returns `{ saved: true }` and the bookmark fills in
3. **the item still does not appear in Saved**

A success for an action whose result is invisible — the absence-versus-failure family, in the one place where the reader is told it worked.

### Ruled: a separate endpoint

`POST /api/saved/[storyId]/archive` with `{ archived?: boolean }`, mirroring how `/api/hide/[storyId]` is already separate from read.

Rejected the alternative of growing `saveBodySchema` with `archived?`, for the UI lane's reason: archiving is a different act from editing a note, hide and read are already split for exactly that reason, and it keeps `saveStory` idempotent rather than giving it a second meaning.

**And a fresh save must un-archive.** The reader has just asked for the item back; leaving it archived is the defect above with a different cause. So `saveStory`'s upsert sets `archived: false` alongside note and tags.

### Acceptance needs a control

- Archive an item, confirm it leaves the default Saved list and appears under `?archived=true`.
- Then save it again from Today and confirm it **returns** to the default list — not merely that the API says `saved: true`. Assert through the database, not the API's own echo.
- Remove the `archived: false` from the upsert and confirm that second test goes red. If it stays green, it is asserting the API's reply rather than the state.

### Also in scope, small

`src/lib/api/types.ts:25` says StoryCard is returned by `/api/saved`. It is not — that route returns the wider `SavedCard` (StoryCard plus note, tags, savedAt), declared only in `src/api/reader.ts`, with no name in the UI contract file. The UI lane is adding `SavedCard` to types.ts and correcting the comment; this ticket owns keeping the server side in step.

[...]

PR #74, opened 2026-09-17
Saved, Settings and first-run onboarding (#16)

### LINKED ISSUE #16: Saved, Settings and first-run onboarding

zarghammir:
Closes #16.

Three screens a reader actually uses, and the plumbing behind them.

**Saved** is the bin. Everything you pressed Save on, newest first, with the grade and the link it came with. You can put a note on any of them, tag them, filter by a tag, and mark them read. Removing takes one off for good.

**Settings** is where you say when you want the brief, where in the world you are, how long it should run, what subjects you follow, how you want to be told, and which theme. It also explains what the four grades mean and lets you run the welcome again.

**The welcome** is the first thing a stranger sees. Three short questions, every one skippable, and it never asks twice.

## The three states this ships honest about

|---|---|---|
| **Archive** | Visible, **disabled**, with the reason beside it | The endpoint does not exist (#70). Faking it in browser storage would disagree with the server the moment you opened a second device, and a button that lies quietly is worse than one that arrives a week late. |
| **Notifications** | Choice is stored; panel says **nothing is sent yet** | There is no push subscription and no mail path in this repository (#72). A product that accepts "email me at seven" and then never writes is worse than one that admits it cannot — you stop opening it and conclude there is no news. |
| **Sources** | Described, not drawn as dead switches | Still no endpoint that can store the choice. |

## The dead control this removes

Today parsed `?length=` and fell back to a hardcoded ten minutes, so the brief length in Settings was written by the reader and read by nobody. The stored value is now the default and `?length=` is the override for one visit — exactly the split `ReadingMode`'s own comment had specified before the endpoint existed. The `localStorage` copy inside that component is gone with it: two answers to one question is how a length just chosen in Settings gets silently overridden by an older device memory, leaving a control looking broken with nothing wrong.

## Measured, not claimed

Production build, standalone server on 3210, real browser.

| Measurement | Result |
|---|---|
| States swept (7 routes x 2 themes x 2 sizes) | **28** |
| axe violations, **any** impact | **0** |
| Serious or critical | **0** |
| Horizontal overflow | **0** |
| Bottom bars measured | **14 / 14** — 6 tabs, narrowest 65px, shortest tap target 59.8px, 0 clipped |
| `npm run verify:saved` | exit **0**, floor passed |
| `npm run verify:today` | exit **0**, floor passed, all four behaviours |
| Unit tests | **343 passing**, 30 files |
| prettier / eslint / tsc | **0 / 0 / 0** |

**The sweep's own floor**, printed on success so it is not taken on trust: 28 of 28 states recorded what they rendered, 0 landed on a route other than the one requested, `/saved` was in state `list` in all 14 of its states, `/settings` in `ready` in all of its, and at least 18 controls were scanned per saved state. Without this the audit would have swept one page seven times — the first-run gate redirects an unseeded context to `/welcome` — and still reported seven routes.

**Controls, each run in the same window as the thing it guards, each reddening the named check and only that one:**

- `AUDIT_CONTROL=narrow` (200px): exit 1 from **clipped labels in 14 of 14 bars**, with the floor still passing — so the red is the layout, not a crash and not an empty instrument.
- `VERIFY_CONTROL=skip-seed`: exit 1 from three **named floor failures**, not an exception.
- `NOTE_MAX` 2000 → 2001: reddened only the limits test.
- Saved store's failed read collapsed to `{ ready: [] }`: reddened only "a read that failed is not a reader who has saved nothing".
- Server preference read reverted to the HTTP client: reddened only the two live-path tests.

## Four defects the browser found, all in my own work

1. **`/settings` and `/welcome` were prerendered** *(worth knowing beyond this PR — a CI build has no `DATABASE_URL`)*. A server component that only reads a database is not dynamic to Next without being told. In live mode the subject list would have frozen at build time — and on a build with no `DATABASE_URL`, which is how CI builds here, "the list of subjects could not be read" would be baked into a static page forever. Nothing would look wrong; the page would simply always say it.
2. **`addInitScript` runs on every navigation, reloads included.** My seed wrote the starting state back over the page on each reload, so every "survived a reload" assertion was measuring the seed being re-applied. **A check destroying the evidence it exists to look for** — and the note assertion was the one that would have hidden it, because a note silently re-seeded to `null` looks exactly like one that was never written. It only surfaced because the tag filter crashed loudly two steps later.
3. **My layout gate broke `verify-today.mjs`.** The first-run gate is in the root layout, so it applies to every route; a script written before it existed waited thirty seconds for a Save button on a page it was not on.
4. **A failed note save threw away what the reader had typed**, then showed an error about a note no longer on screen.

Plus one caught before the machine: seed ids were `1, 2, 3` and the fixtures use `412, 418, 421`, so every seeded run would have measured an empty screen. Ids are read off the page now, from a new `data-story-id`.

## One assertion retired, then given back — #75

Today resolves its default length **on the server**. On fixtures preferences live in `localStorage`, which the server cannot see, so the end-to-end assertion was retired: it was a true failure about the harness rather than about the app, and a green there would have been a lie about coverage.

Rather than leave that permanent, I raised the consequence — the owner has **no preview URL**, so fixture mode is the only way he can see a build, and a feature that looks broken in the one mode he can reach is a real cost. The PM ruled to fix it, and it is in this branch as **#75**.

**One cookie, one value.** The chosen length is mirrored into a single cookie the server can read, in fixture mode only. It carries the length and **nothing else** — deliberately not the preferences object, because that row holds the reader's **email address** and a cookie is the wrong place for it. A preference, never an identifier.

**On "self-hosted, nothing is sent anywhere":** that sentence is about a **third party** learning something. A cookie the app sets and the app's own server reads adds no third party, and nothing leaves the reader's machine. Settings says so in their own words, and says it **only on a fixture build**, because where there is no cookie the sentence would be false.

**The control found a hole in my own test.** Serialising the whole preferences row into the cookie reddened three of five checks — but *not* the one named "carries nothing that could identify the reader", because it searched the raw cookie string for `reader@example.com` while `encodeURIComponent` had made it `reader%40example.com`. The assertion could not see the leak it existed for. It now asserts the **decoded** value is exactly the length, and the control reddens all four.

The cookie is reader-writable, so it is treated as untrusted input: a value this build cannot render falls back rather than putting Today into a state it cannot draw, and the value is escaped so a length cannot smuggle in another cookie attribute. Both asserted.

**Verified in a browser.** Today shows **5 stories at the saved five-minute length and 6 at `?length=all`** — before the cookie it was 6 and 6 — and the URL still overrides. Control, because that assertion was new: stop writing the cookie, rebuild, re-run, and it goes back to 6 and 6 with **that assertion red and only that one**. The assertion this PR retired is back and passing, so the retirement was temporary and recorded rather than quietly permanent.

## Five things the screenshots showed that the DOM did not

**A screenshot is a measurement, not a deliverable, and the things it measures are the ones no test can see.**

I took the four screenshots to hand over, then looked at them instead of sending them. Five things were wrong — after a block in which every number came back green: 28 axe states at zero violations, zero overflow, the bottom-bar geometry gate passing, 343 unit tests. **Not one of the five was visible to any assertion in this PR.** All five are the same category: correct markup that reads badly.

1. **The disabled Archive did not look disabled.** It sat beside Mark read and Remove in identical weight. Everything above about a gated control staying visible and saying why is defeated if it does not *look* inert — the reader clicks, nothing happens, and the sentence explaining why is never the thing they blame. It now takes the **dashed border this design already uses for "not built yet"** in the empty states and the unwired Settings panels, so the solid-bordered buttons beside it are visibly the ones that do something.
2. **The archive sentence repeated on every card.** Twenty saved stories meant reading it twenty times. Said once for the list now, with every Archive button pointing at that one element through `aria-describedby`.
3. **`UTC (not in this list)` was the app's own default.** `Intl.supportedValuesOf` does not enumerate `UTC` here, so every fresh reader saw their default timezone flagged as an anomaly about a choice they never made. "Not in this list" is now said only when the platform genuinely rejects the zone.
4. **The step indicator read `1 OF 3  2 OF 3  3 OF 3`**, because each item carried the whole phrase. Said once; the items are numerals.
5. **The tag input was the width of the card**, so a one-word box read as the most important control on it.

All of this touches shared components, so every number in the table above was re-measured after the fixes rather than reported from before them.

- **`user_preferences.theme` stays unwritten.** `ThemeScript` reads `localStorage` before first paint, so a server-stored theme cannot beat the first paint without a cookie and would flash — and two writers for one fact is the disagreement that component's own comment warns about. Settings already says "this device only".
- **The first-run gate fires on `onboardedAt === null` and nothing else.** Not while preferences are loading, and emphatically not when the read **failed**: an unreachable store is not a new reader, and redirecting on a failed read would walk somebody who has used the app for months through a setup whose every answer then fails to save.
- **No `empty:hidden` on a live region.** `display:none` removes an element from the accessibility tree, so a region hidden while empty is added to that tree at the moment it receives its message and may not be announced. Every one of these screens reports the result of a write that way. The `empty:hidden` on `PageShell`'s summary wrapper is untouched — that is layout, not a live region.

## For #65's author

I had to write a server-side preferences reader (`brief-length.ts`) and a server-side catalogue reader (`catalogue-server.ts`) for the same reason your `brief-server.ts` exists: a server component fetching its own app at a relative URL has no origin to resolve it against and throws on every load. **Merge order is ruled #65 first**, and I will fold both of mine into whatever convention `brief-server.ts` sets rather than leave a second shape. If that shape turns out wrong for either of these, I will say so rather than diverge quietly.

The general rule this produced, which cost three instances to learn: **a fixture path cannot exercise a relative fetch, so every call that works in fixture mode is untested against the mode it will actually run in.**

## The four screenshots

Taken from a build I had just watched behave, and the script asserts which page each came from — a screenshot of the wrong page is still a perfectly good PNG and nothing downstream would notice.

| Shot | What it shows |
|---|---|
| **Saved, desktop** | Where it lives and what it holds: three stories, the tag filter, a note and tags on the first card. |
| **Settings, desktop** | Settings as a list of sections — brief, then what you follow — so interests and a watchlist drop in without a rewrite. |
| **The welcome, first screen** | The pitch is what the app *does for you*, not what it contains, which is why it survives the direction change. |
| **One card, dark** | The control that is not built yet: visible, dashed, inert beside two solid buttons, and saying why. |

They are not inline above because **`gh` cannot attach images** — it takes `--body-file` only, and GitHub's image upload is a web-UI endpoint with no REST equivalent. The files are at `/tmp/ai-radar-16-shots/` (332 KB for all four).

**Why they are not committed, as a rule rather than a preference.** This repository *does* track screenshots — 18 of them, 2.4 MB, under `docs/screenshots/` and `design/screenshots/` — so "we don't commit PNGs" would simply be false. The distinction that holds is **lifetime, not file type**: `docs/screenshots/<area>/` is a *documented set*, what the product currently looks like, and it earns its bytes by staying true. Per-PR evidence is true for one review and stops being interesting the day the PR merges, while every stranger who clones the repository pays for it forever.

So the distinction holds, but **conditionally**: a documented set earns its cost *only once something regenerates it and something else notices when it is wrong*. Until then it is a liability that looks like documentation — and today nothing does either. No script and no `package.json` entry produces `docs/screenshots/`; the 18 images are hand-maintained.

### ⚠️ Two committed screenshots are KNOWN STALE because of this PR

`docs/screenshots/shell/settings-laptop-light.png` and `docs/screenshots/shell/settings-phone-dark.png` no longer show Settings — it has four sections that did not exist when they were captured. Nothing in CI notices, and they were found by accident, while looking at a different screenshot.

They are **not regenerated here**. That needs a machine block, and a stale PNG does not justify taking the shared machine back forty minutes after I had wrongly reclaimed it. Recorded rather than silently left: **a known-stale artefact with a note is honest; a silently stale one is the defect** — the same distinction this whole PR is built on. The requirement that makes it impossible to happen again is filed on #48: the harness owes a staleness check, so that a screen changing visually makes something go red, not merely that someone *can* re-run a script if they remember.

## Also in here

`docs/DESIGN.md` gains section 11: the four named screen states and why the last two are never merged, never rendering defaults after a failed read, never gating on one, a control with nothing behind it staying visible and saying why, Settings as a list of sections, and every write reporting what happened to it.


zarghammir:

Reviewed at head `1cfd31e`, 0 commits behind `origin/main` (`1c0e5ac`), 40 files, counted two ways with identical sorted lists. This is strong work and most of this comment is confirmation rather than criticism — including two defects I went looking for and did not find. One thing blocks: **CI is red on this branch, the cause is this PR, and it reproduces deterministically on my machine.**

---

## BLOCKING — `browser checks` fails, and the family sweep stopped one script short

`gh pr checks` at this head: `lint, typecheck, test, build` pass, `compose-smoke` pass, Vercel pass, **`browser checks` FAIL** (run `35177950623`). Step 11, the accessibility sweep, passes. Step 12, "The four checks axe cannot see", fails:

```
locator.focus: Timeout 30000ms exceeded.
  - waiting for locator('[role="radio"]').first()
  at scripts/verify-shell.mjs:42:26
```

Reproduced locally against a clean build of this branch — `node scripts/verify-shell.mjs` exits 1 with the identical error at the identical line. Not a flake, not CI-specific.

The cause, measured rather than inferred:

```
seeded=false   asked for /settings -> landed /welcome    radios=0
seeded=true    asked for /settings -> landed /settings   radios=3
```

This PR's first-run gate lives in the root layout, so it applies to every route, and an unseeded context asking for `/settings` is sent to `/welcome`, where there are no radios. `scripts/verify-shell.mjs` does no seeding at all — it has no `markOnboarded`, no `addInitScript`, nothing — and **it is not among this PR's 40 files.**

This is the sweep stopping at the instances in front of it. `scripts/lib/seed.mjs` exists precisely for this hazard, and its own comment describes this exact failure: *"a script written before the gate existed then waits thirty seconds for a control that is on a different page. verify-today.mjs failed exactly that way the first time it ran against the gate."* That lesson was applied to `verify-today.mjs`, `verify-saved-settings.mjs` and `a11y-audit.mjs`. `verify-shell.mjs` is the fourth script in the same directory with the same dependency and it was not visited — most likely because it is the one the PR did not otherwise need to touch.

**A second observation worth as much as the fix.** Line 38 already carries the right floor:

```js
if (radioCount !== 3) floor.push(`expected 3 radios on /settings, found ${radioCount}`);
```

with a comment saying zero radios would otherwise read identically to a broken widget. That floor is correct and it never gets to speak, because `radios.first().focus()` four lines later throws first and the floor is only reported at the end of the script. CI therefore says "Timeout waiting for locator" — which points at Playwright — instead of "expected 3 radios on /settings, found 0", which points at the gate. The diagnostic that was built for exactly this case was defeated by ordering. Consider bailing on the floor before the first interaction, so the script explains itself when this happens again.

I have deliberately not verified a fix. My attempt to patch seeding into the script did not even import correctly, so that run proved nothing and I am not reporting it as evidence.

---

## Two defects I predicted and did NOT find

I would rather record these than quietly drop them, because both were reasonable and both are wrong.

### Today is not prerendered. I was wrong, and here is the control

I expected `/` to be statically prerendered despite reading a cookie, on the theory that Next signals its dynamic bailout by throwing and `defaultBriefLength`'s `try { ... } catch { return FALLBACK }` swallows it — which would mean #75 ships a cookie a built app never reads, with every test still green because they mock `next/headers`. The route table says otherwise:

```
┌ ƒ /            ← dynamic
├ ○ /saved
├ ƒ /settings
└ ƒ /welcome
```

I then isolated *what* holds it dynamic, since `searchParams` alone would be enough and would make the dynamism incidental — a later removal of `?length=` would silently make the page static again:

| tree | `/` |
|---|---|
| as shipped | `ƒ` |
| `searchParams` read removed, prop still declared | `ƒ` |
| `searchParams` removed from the signature entirely | `ƒ` |
| cookie read removed as well | **`○`** |

So the cookie access alone holds the page dynamic, the catch-all does not swallow the signal in this version of Next, and there is no incidental fragility. The deletion control identifies the exact line that holds it. On `main` the question does not arise — Today reads no cookie there, so `○` would be correct.

### The dark-mode contrast on Saved is the design, documented, and axe was right

I measured the card action buttons at 2.91:1 and 3.63:1 against the page background in dark mode and thought I had a contrast failure that axe had missed. I was computing against the wrong pair. The card's surface is `bg-paper`, which resolves to `#f3f3f1` in dark mode, and axe measured `#585c62` on `#f3f3f1` at 6.05:1. `globals.css:157` says so outright — *"paper stays light: you read on the print, the bench is what it sits on"* — and `story-card.tsx` on `main` uses the same `bg-paper text-ink`. This PR follows the established design. That file also already records that orange text on paper is 3.17:1 in dark and must not be used, so the palette has been reasoned about more carefully than my check was.

---

## The cookie — verified on both sides

**Writing.** The rewritten privacy assertion is the right shape and it does catch the control that defeated the old one: asserting the *decoded* value IS the length means serialising the whole preferences row fails on the first line rather than slipping through because `encodeURIComponent` hid `reader@example.com` as `reader%40example.com`. The escaping test asserts the exact encoded string for `5; domain=example.com`, so attribute smuggling is genuinely closed rather than described.

**Reading.** `asBriefLength` is a strict allowlist, so an unrecognised cookie falls back to ten minutes rather than putting Today into a state it cannot draw, and `brief-length.test.ts` drives that with `"42"`, `""`, `"<script>"` and `"ALL"` — plus a positive beside each negative, and a case proving the cookie is not read at all against a real database.

**End to end**, from the repo's own run against my build: `today.atPreference: 5`, `today.atAll: 6`, `urlStillOverrides: true`. The length chosen in Settings genuinely changes Today, and `?length=` still wins for one visit.

### Minor: the untrusted-input test misses the only input that throws

The four values above all *return* something wrong. Malformed percent-encoding — a bare `"%"`, or `"%E0%A4"` — makes `decodeURIComponent` **throw** instead, which is a different branch: it unwinds out of `readPreferences` into `defaultBriefLength`'s catch rather than reaching `asBriefLength`. The behaviour is correct either way, so this is a coverage gap rather than a defect, but it is the one class of input the test exists for and does not cover.

## The seeding floor bites, and I ran the control

The repo's sweep reproduces exactly on my build: **28 states audited, 0 axe violations of any impact, 0 horizontal overflow, 14/14 bottom bars measured, narrowest tab 65px, shortest tap target 59.8px**, and the floor that matters:

```
"seeding": { "statesRecorded": 28, "expected": 28, "landedElsewhere": 0 }
```

28 of 28 recorded what they rendered and none landed on a route other than the one asked for — so the audit is measuring seven routes rather than one page seven times. And the control fires: `VERIFY_CONTROL=skip-seed` exits 1 with three named failures (`Saved is in state "null", not "list"`, `Saved rendered 0 cards, floor is 2`, `Settings never reached "ready"`). A floor that fires when the thing it guards is removed is a floor.

`noteSurvivedReload: true` and `briefLengthSurvivedReload: true` now mean something, because `markOnboarded` writes only when the key is absent. The reload assertions are measuring the app rather than the seed being re-applied.

**Minor, and the same hazard one file over:** `scripts/lib/fixture-ids.mjs:22` writes its seed with no such guard. It is harmless today — that context does one `goto` and is closed in a `finally`, so nothing re-navigates and no persistence is asserted through it. But it is the identical pattern the sibling file fixed with a comment explaining why, and the day someone adds a reload there it goes quiet in the same way.

## My own browser pass

Independent of the repo's scripts, at 1440×900 and 390×844, light and dark, 16 states across `/`, `/saved`, `/settings`, `/welcome`:

- **console: 0 errors, 0 warnings. 0 page errors.**
- **my own axe run: 0 violations** in every state.
- **keyboard:** first `Tab` reaches a visible "Skip to content" link in all four viewport/scheme combinations, with a focus ring.
- every route landed where it was asked to land, with a seeded context.

## The disabled Archive — I judge it the design's own vocabulary

Measured on the rendered page: `border-dashed` against solid siblings, `cursor: not-allowed`, text `rgb(110,106,97)` against the siblings' `rgb(88,92,98)` so it sits lighter on the paper, the real `disabled` attribute, and one explanatory sentence with `aria-describedby` pointing three buttons at a single `#archive-not-built` node rather than repeating it per card.

`border-dashed` appears ten other times in this codebase and every one means *this is a note, or there is nothing here yet* — empty states, "not built" explanations, the interests placeholder. This is the first time it is applied to a control, so the element is new but the meaning is not: Archive is precisely a thing that is not built yet. I judge it the design's existing vocabulary rather than a new signal, and correct.

Two honest notes. The differentiation is carried almost entirely by border *style* — colour, size, font weight and icon treatment are identical to the live buttons — so on a phone or for a low-vision reader the cue is subtle. And the sentence doing the real explaining appears **once, above the list**: a screen-reader user gets it on every card through `aria-describedby`, but a sighted reader scrolled to the twentieth card has only the dashes. That is a fair trade against repeating it twenty times, and worth knowing you made it.

---

## Verified myself

Head `1cfd31e`, 0 behind main, files 15→40 two ways identical. Five checks polled to completion — four pass, `browser checks` fails as above; I did not treat the earlier `in_progress`/EMPTY conclusions as a pass. Locally, with a real database and nothing skipping: `npm test` **343 passed / 29 files / 0 skipped, exit 0** · `tsc --noEmit` 0 errors · `lint` exit 0 (2 warnings, both pre-existing on `1c0e5ac`) · `format:check` 0 · `build` 0 · `routes:check` 0, 12 public routes. The route table above, all four prerender controls, the repo's three browser scripts plus the skip-seed control, my own 16-state browser pass, and the rendered-style measurements of the Archive.

## Taken on trust

The four screenshots attached to the PR and the five design defects they are said to have surfaced — I looked at my own rendering of Saved rather than at those images, and I agree with the one I could judge. That the Archive endpoint is #70's work. My toolchain is not CI's: `.nvmrc` pins Node 22, none is installed here, so my numbers are Node 25.8.1 — worth naming here because CI's failure reproduced on my Node anyway, which makes the version difference irrelevant to the one finding that matters.

zarghammir:

Re-review at head `e10842d`, 0 commits behind `origin/main`, new comment rather than an edit.

**My blocking finding from the last round is fixed and I re-proved it rather than reading the diff**: `node scripts/verify-shell.mjs` now exits **0** against a clean build of this branch, where the same command at `1cfd31e` exited 1 with `locator.focus: Timeout 30000ms exceeded`. The `bailIfBroken` helper and the landing-path floors are both good work, and the comment in `floor.mjs` states the defect more clearly than my review did.

Two things block, and both are inside this commit's own stated scope: it set out to fix the CI failure **and the one it would have become after #65**, and the second half covers three of the four scripts.

---

## BLOCKING 1 — the post-#65 fix misses the script that most needs it

I tested the actual post-#65 shape rather than a resemblance of it: built with `NEXT_PUBLIC_USE_FIXTURES=0`, pointed at a real database, `onboarded_at` NULL, and asked for `/settings` three ways.

```
A  no seeding at all             -> landed /welcome
B  localStorage seed ONLY        -> landed /welcome
C  markOnboardedOnServer         -> landed /settings     (returned true)
```

**Row B is the proof that the mechanism was needed and row C is the proof that it works.** In live mode the localStorage half is exactly as inert as the comment in `seed.mjs` predicts, and the server half genuinely clears the gate. That part of the claim is confirmed by execution, not by resemblance.

The gap is which scripts got it:

| script | `markOnboardedOnServer` | drives gated routes |
|---|---|---|
| `verify-today.mjs` | yes, and reports it | yes |
| `verify-shell.mjs` | yes, and reports it | yes |
| `a11y-audit.mjs` | yes, **discards the result** | yes |
| `verify-saved-settings.mjs` | **no — not at all** | `/`, `/saved`, `/settings` |

`verify-saved-settings.mjs` seeds through its own inline `addInitScript` in `seededContext()` — localStorage only — and never writes the fact where a live build keeps it. Row B above is precisely that configuration, and it lands on `/welcome`.

So when #65 flips the default, the script whose entire subject is Saved and Settings is sent to `/welcome` on every seeded navigation: the same failure, one merge later, in the file that was not visited. That is the defect this commit exists to prevent, and it is the fourth instance of the pattern the commit message names.

It will fail loudly rather than silently — its floors at lines 156–157 would fire — so this is a CI break waiting one merge away, not a false green. It is still the thing the commit claims to have handled.

I did not run that script against the live build, and I want to be exact about why: in live mode this database has no ingested stories, so `collectStoryIds` fails first for a data reason and would mask the gate. The finding rests on row B plus reading, not on a run of that script.

## BLOCKING 2 — the audit captures the answer and throws it away, and the linter says so

`scripts/a11y-audit.mjs:84`:

```js
const onboardedOnServer = await markOnboardedOnServer(base);
```

Nothing reads it. `verify-today.mjs:182` and `verify-shell.mjs:193` both do `out.onboardedOnServer = onboardedOnServer;`. The audit does not, and `npm run lint` reports it:

```
scripts/a11y-audit.mjs
  84:7  warning  'onboardedOnServer' is assigned a value but never used
```

**Lint goes from 2 problems to 3 on this commit.** The other two are pre-existing on `1c0e5ac` — `bloomberg` in `fixtures.ts` and `ScoreComponents` in `rank-all.ts`, both confirmed against main. This one is new.

It matters more than an unused variable because of what the variable is. `seed.mjs` says of that return value: *"Returns whether it applied, so a caller can report it rather than assume it."* The audit assumes it. This is the observed-and-not-reported shape that the `floor.mjs` docstring was written about, in the same commit that fixes the other instance of it — a value computed, correct, and connected to nothing.

## The shared helper's blast radius, measured — not blocking, but it changes what a run tells you

`bailIfBroken(floor)` throws when the floor array is **non-empty**, and each script keeps one floor for all of its blocks, with the handler at the top level. So a bail fires on failures that have nothing to do with the check it guards, and ends the whole script rather than the section.

Demonstrated on `verify-today.mjs` — one synthetic failure injected into the first block, unrelated to anything later:

```
clean run    keys: readingLength, save, hide, states, onboardedOnServer, floor
one early failure  keys: readingLength, onboardedOnServer, floor
                   failures reported: ["SYNTHETIC early failure, unrelated to Save or Hide"]
```

`save`, `hide` and `states` — three independent blocks, including the four-viewport responsive sweep — are abandoned, and the report names a single failure, which reads as *one* thing being wrong. Before this commit the script ran every block and reported all of them.

**This cannot produce a false green**: the floor is still non-empty, the run still exits 1. The cost is diagnostic yield and the number of runs it takes to clear a red. Worth knowing because the change's purpose is diagnostics.

The helper's rationale — *"once the floor is broken, everything after it is measuring something else"* — is exactly right for the case it was written for, a count the next interaction depends on. It is not right for a failure in an unrelated earlier block. Bailing on the entries added **since that section started** rather than on the whole array would keep the fix and drop the side effect, and is a small change to `bailIfBroken`'s signature.

**The `SkipRest` absorption is behaviourally identical** and I checked rather than assumed: the one call site is inside `if (!ready)`, immediately after a `floor.push`, so the array is guaranteed non-empty and the conditional throw fires exactly where the unconditional one did. The latent edge — moving that call above its `push` would silently make it a no-op — is worth a comment, not a change.

---

## Verified myself

- `verify-shell.mjs` exits **0** (was 1 with a Playwright timeout at `1cfd31e`). `verify-today.mjs` exits 0, floor passed. `verify-saved-settings.mjs` exits 0, floor passed.
- The 28-state sweep reproduces exactly: **28 states audited, 0 serious-or-critical, 0 total violations, 0 horizontal overflow, 14 bars measured, narrowest tab 65px**, and `statesRecorded: 28` with `landedElsewhere: 0`.
- The `skip-seed` control **exits 1** with its three named failures, so the seeding floor still bites.
- Local suite with a real database and nothing skipping: `npm test` **344 passed / 29 files / 0 skipped, exit 0** · `tsc --noEmit` 0 errors · `format:check` 0 · `build` 0 · live-mode build 0, with `/`, `/settings` and `/welcome` all `ƒ`.
- Lint exit 0 with 3 warnings, one of them introduced here as above.
- The three-way gate experiment and the `bailIfBroken` control, both above.

## Taken on trust

That CI is green at this SHA — it was still `in_progress` with empty conclusions while I worked, and I did not treat that as a pass. The screenshots and the five design findings behind them. And the service-worker context being seeded in all three places, which I did not separately enumerate.

zarghammir:
REVIEW VERDICT: APPROVE

Round four at head `f7e70ac`, 0 commits behind `origin/main`. **All five checks pass at this exact SHA** — `browser checks` was `in_progress` when I started and I polled it to `completed/success` rather than counting its presence.

Both of my blockers are fixed, the third gap that neither of us named is fixed, and the `bailIfBroken` scoping is measured in both directions. I re-proved each rather than reading the diff.

## The bail still fires for its own case — the outcome that mattered most

Scoping a bail could easily have turned it into a no-op, and a diagnostic that improved its yield by no longer firing would have been the worst possible result. It fires. With seeding neutered so `/settings` redirects:

```
exit: 1
floor: "expected 3 radios on /settings, found 0 (landed on /welcome)"
timeouts in output:     0
stack traces in output: 0
```

The named floor, the landed path, no Playwright noise. That is the original defect's diagnosis, printed.

And the yield is genuinely restored. Same synthetic early failure I injected at `e10842d`, same script:

| | report keys |
|---|---|
| at `e10842d` | `readingLength, onboardedOnServer, floor` |
| at `f7e70ac` | `readingLength, save, hide, states, onboardedOnServer, floor` |

`save`, `hide` and the four-viewport sweep all run again, and the run still exits 1 with `floor.passed: false`. Both directions hold: it stops what it should and no longer stops what it should not.

The `since` mark is applied consistently — six `bailIfBroken` call sites across three scripts, each with a `sectionStart(floor)` at the top of its own section. No unscoped call remains.

## The latent-edge comment is a claim, and it is true

`verify-saved-settings.mjs:250-254` says moving the bail above its push *"would silently turn it into a no-op — the check would still be here and would stop checking."* Checked: `mark` is taken at 241, the push is at 248, the bail at 254. Move the bail above the push and `floor.length > mark` is false, so it returns without throwing. The comment describes the mechanism correctly, and it is kept adjacent to the thing it protects.

## The inseparability claim, checked against what the file actually says

The remedy is bundling: `markOnboarded(context, base)` writes both halves, so a caller cannot do half of it by forgetting a line inside that function. That much is true and it is what the file claims.

**The bad state is not inexpressible, though, and I want to be precise because "unable to express" is a strong standard.** `markOnboardedOnServer` remains exported, and any script can still write its own `addInitScript` — which is exactly how the original gap arose. What has changed is that the inline path now has one legitimate user instead of being the default.

`seed.mjs` does not overclaim this. Its own comment scopes the claim to the call site — *"a caller cannot do half of it by forgetting a line"* — and then names the real guard outright:

> *Neither half is required to succeed… The landing-path floors in each script are what catch the case where NEITHER applied — **that is the guard, not this function**.*

That is the accurate account, and the floors do hold: the `skip-seed` control exits 1 with three named failures, and the audit's own floor reports `statesRecorded: 28, landedElsewhere: 0`. So the residual risk is a future script copying the inline pattern and omitting the onboarding key — caught by the floor rather than prevented by the type. Given the accessibility sweep genuinely needs to write theme, saved and marks in one init script, an escape hatch with one documented user and a floor behind it is the right trade. I am recording where the safety actually comes from, not asking for a change.

## The mechanical sweep — I checked its predicate, not its contents

A sweep is only as good as what it greps for, so I ran a wider one: every file under `scripts/`, matching navigation by **any** means — `.goto(`, `.reload(`, `waitForNavigation`, `.click(` — against how each seeds.

```
a11y-audit.mjs           goto:1  nav-any:1   serverOnly:1  inlineInit:1
lib/fixture-ids.mjs      goto:1  nav-any:1   markOnboarded:1
verify-saved-settings    goto:9  nav-any:23  markOnboarded:1  inlineInit:2
verify-shell.mjs         goto:6  nav-any:7   markOnboarded:3
verify-today.mjs         goto:6  nav-any:10  markOnboarded:4  inlineInit:1
lib/browser.mjs          —       —           (does not navigate)
lib/floor.mjs            —       —           (does not navigate)
```

`verify-saved-settings` navigates 23 times by the wide predicate against 9 `.goto`s, so a `.goto`-only sweep could in principle have missed a script that only ever clicks — it did not miss one here, but the predicate was narrower than the risk. Every navigating file seeds both sides, and I checked the one exception directly: the audit's inline script does write `ai-radar-fixture-preferences` with `onboardedAt` alongside the theme and saved keys, so its split usage is complete rather than half.

I also checked for browser-driving code **outside** `scripts/`, which the sweep's predicate would not have covered at all. There is none.

`fixture-ids.mjs` is the real find here, and it is the first instance of this habit caught before it shipped rather than after. Its old comment argued the unguarded write was harmless *"because this context does one goto and is closed in a finally"* — true about reloads, and silent about live mode, where the navigation would have landed on `/welcome` and the error would then have blamed Today for offering no stories. A helper that misattributes its own failure is worse than one that fails.

## Verified myself

- All four scripts exit 0 against a clean fixture build: `verify-shell`, `verify-today`, `verify-saved-settings`, `a11y-audit`.
- The audit reports `onboardedOnServer: true` at line 364 — the flag I filed as captured-and-discarded is now in the report.
- **Lint is back to 2 problems**, both `bloomberg` in `fixtures.ts` and `ScoreComponents` in `rank-all.ts`, both confirmed pre-existing at `1c0e5ac`. The `onboardedOnServer` warning is gone.
- `npm test` **344 passed / 29 files / 0 skipped, exit 0** · `tsc --noEmit` 0 errors · `format:check` 0 · `build` 0.
- The audit's floor: 28 states audited, 28 recorded, 0 landed elsewhere, 0 serious-or-critical, 0 horizontal overflow, 14 bars.
- Both `bailIfBroken` controls above, each with the mutation confirmed to have landed before the result was read.

## Not verified — and this should not be recorded as verified by anyone

`verify-saved-settings` end-to-end against a **live** build. The dev database has no ingested stories, so `collectStoryIds` fails first for a data reason and the run would produce a red that says nothing about the gate. That is the same wall as last round and the reasoning is the same: a measurement that cannot discriminate is worth declining rather than performing. The gate half of that script is covered by the three-way experiment; the rest of it is fixture-mode evidence only.

I did not re-run the three-way seeding experiment on `fixture-ids`, since I ran the equivalent last round and the mechanism has not changed.

Issue #79, opened 2026-09-17
"Show HN:" must be a launch verb — #57 and #69 are correct apart and harmful together


zarghammir:
**#57 and #69 are individually correct and harmful together.** Found by the Intelligence lane by reading #57's classifier against Show HN title shapes — *after* `git merge-tree` came back clean with zero shared files and exit 0. PM verified each claim against the diff before ruling.

### The interaction, and why it is asymmetric

After #57, every Show HN item goes through `classifyContentType`, because `isDeclaredFact` exempts only `PAPER` and `DISCUSSION` — and #69 files launches under `RELEASE`, which is not exempt.

**It can never be promoted.** `TOOL` needs a launch verb, and `"Show HN"` is not one. `MODEL` needs either a launch verb or a model family leading the headline — and the `"Show HN: "` prefix guarantees the family is never first. So `Show HN: I built an open-source CLI` cannot reach TOOL, and `Show HN: Gemma 4 fine-tuning toolkit` cannot reach MODEL. Not rarely — never.

**It can be demoted.** The single-word rules still fire. `"ai act"` is in `REGULATION_WORDS`; `"raises"` and `"seed round"` are in `BUSINESS_WORDS`. So:

> `Show HN: an EU AI Act compliance checker` → **Policy**

A tool *about* regulation, labelled *as* regulation, with the one signal that would have corrected it — the prefix announcing a launch — unread.

**Net effect: merging both makes Show HN typing strictly worse than #69 alone.** The half that works is the harmful half.

### The fix

Add `"Show HN:"` to `LAUNCH_VERBS` in `src/pipeline/normalize/content-type.ts`, with its own test.

This cannot go in #69: that file does not exist on `main`.

### The fix that was considered and correctly rejected

The adapter could declare `RELEASE` explicitly rather than leaving the type undefined. `decideContentType` returns an adapter declaration immediately, so the classifier would never run and nothing could be demoted. **It works, and it is wrong.**

#57 established that *adapter-declared* means structured data read from the source, not a policy we chose. The HN API does not return a content type; the adapter would be asserting one. It would also write `adapter` into the provenance column for rows no adapter ever declared — **the exact confusion that column was added to end.** Dodging a classifier by lying about where a value came from is not a fix, and the lane rejected it on those grounds unprompted.

### PM ruling on ordering

[...]

Issue #78, opened 2026-09-17
Does a launch deserve its own kind? — carried from #69, answered by the Signal model


zarghammir:
Carries the deferred decision from #69, so the deferral has an owner rather than sitting in a comment.

### The decision that was made, and its expiry

#69 files Show HN launches under `RELEASE`. That was ruled deliberately and for stated reasons: a new content type costs a migration, a label, a badge colour, a ranking weight, a place in every exhaustiveness switch and a row on #66, and the Signal model may replace the Story content-type axis entirely. **`RELEASE` is imprecise and known to be** — it connotes a version of a known product, and a Show HN post is a debut.

### What would change the answer

The pivot introduces a generalized `Signal` rather than forcing everything into an article shape. A GitHub repository trending, a Hugging Face model appearing, a Show HN launch and a newsroom article are not four flavours of the same object, and the current content type is a single axis doing the work of at least two: **what kind of thing is this** and **what happened to it**.

If that axis splits, a launch is not a content type at all — it is an event kind, and `RELEASE` stops being a compromise and starts being a category error.

### Acceptance

- The Signal model's design says explicitly whether a launch is its own kind, an event on a kind, or stays a content type — **stated, not left to fall out of the implementation**.
- If it becomes its own kind, `src/sources/hackernews` is the migration point and the comment recording the imprecision is deleted in the same change, so the reasoning does not outlive the decision.
- If it stays `RELEASE`, that is written down as a decision rather than left as an unrevisited default. **A deferral nobody ever returns to is indistinguishable from a choice nobody made.**
- Whatever the answer, no new content type is added without a row on #66 — somebody looks at the badge rendered, not merely proves a label exists.

Blocked on the Signal model, which is the pivot's first architectural piece and not yet ticketed. **This ticket is where it gets remembered.**

Related: #69 (the assignment), #66 (badges nobody has looked at), #77 (a category nothing can produce), #57 (badge describes the item, not its source).

Issue #76, opened 2026-09-17
Classifier recall is about a third — the misses are model names it has never heard of


zarghammir:
Split out of #57, which measured this rather than assuming it.

**Measured on #57's head:** of 515 live items, 40 were reclassified and 36 were right — about 90% precision. **Recall is about a third.** On a deterministic sample of items the classifier left alone, roughly one in six should have moved.

The misses are structural, not sloppy. They are model launches whose names are not on the hand-maintained family list:

- "We're launching Lyria 3.5"
- "TimesFM-3: A zero-shot foundation model"
- "Aurora 1.5: Extending open foundation models"

**The obvious fix has already been tried and reverted, with numbers.** A generic "capitalised name + version number" rule found 15 more items and got 8 of them wrong, taking MODEL precision from 90% to 71% — a date ("State of Open Models: Summer 2026"), newsletter numbering ("Import AI 454"), software versions ("Apple releases iOS 27", "Datasette 1.0a39"), an incidental figure ("Training Text-to-Image Models 3.6x Faster"). That measurement lives in the source next to the rules it governs, so anyone reaching for it finds the result first. **Do not re-run that experiment expecting a different answer.**

Precision is worth roughly an order of magnitude more than recall here, because #57 measured what a false positive costs: about **seventy places** of promotion. A wrong MODEL badge puts a feature announcement in the top ten.

### What would actually move it

Two candidates, neither a cleverer regex:

1. **A maintained model-name list** — cheap, deterministic, needs an update mechanism or it rots (the same failure mode as #48's screenshots).
2. **A summariser that reads the item** rather than its headline — stronger, costs money per item, and lands inside the cap question in #35.

### Acceptance

- Recall re-measured against a live ingest of all eighteen sources, with the sample and the method stated, not just the figure.
- **MODEL precision does not fall below the 90% measured on #57** — that floor is the point of the ticket, and a change that raises recall by spending precision is a regression here.
- Whatever list or source is introduced has a stated way of being kept current, or the ticket says out loud that it will go stale and who notices.
- TREND stays deliberately unreachable unless this ticket proves otherwise; #57 records why.

Related: #56 (revisit the MODEL weight against real traffic), #66 (the badges are not rendered yet).

zarghammir:
**A generalisation surfaced by #79, which is worth recording here rather than losing in that ticket's specifics.**

#79 is about Show HN, but the defect it exposes is not about Show HN:

[...]

Issue #73, opened 2026-09-17
"AI-powered X" is never recognised as AI — hyphens defeat the short-keyword branch


zarghammir:
**"AI-powered X" is not recognised as an AI story. It never enters the brief at all.**

Found by the reviewer on #57, in `matchesKeyword` — pre-existing on main, untouched by that PR.

Short keywords take the whole-word branch, and the haystack keeps its hyphens. `DEFAULT_AI_KEYWORDS` contains **ai, gpt, llm, rag, gpu, xai** — all four characters or fewer. So:

| headline | haystack | ` ai ` present? |
|---|---|---|
| AI-powered X | `ai-powered x` | **no** |
| GPT-5 does Y | `gpt-5 does y` | **no** |
| LLM-based Z | `llm-based z` | **no** |
| RAG-based … | `rag-based …` | **no** |
| GPU-accelerated … | `gpu-accelerated …` | **no** |

These are not exotic headlines; they are among the most common shapes on Hacker News. And the consequence is not a wrong badge — **the story is dropped at the gate and never reaches the reader.**

### The new test cannot catch it

`finds every keyword in a title that contains it` builds `Show HN: a ${kw} project I built` — which places every keyword between two spaces, the one context where the whole-word branch cannot fail. The test's **name** claims "a title that contains the keyword"; its **body** tests only the space-delimited case. The hyphenated case is both the realistic one and the broken one.

### The fix is not to strip punctuation

That was the obvious move and it is proven wrong: stripping destroys `a.i.`, `text-to-video` and `text-to-image` — three of the forty keywords, and the last two are exactly the material the product's new direction is about.

**Treat a hyphen as a word BOUNDARY for the short-token branch, while leaving dotted keywords intact.** Boundary and character-class are different questions, and `keywords.ts` currently answers both with the same `replace`. That is the root cause rather than the symptom.

### Acceptance needs both directions

- `AI-powered X`, `GPT-5 …`, `LLM-based …` match.
- `a.i.`, `text-to-video`, `text-to-image` still match.
- `Mistake` still matches `stake` for topic tagging, which depends on substring behaviour — the three callers have three different requirements and this fix must not collapse them.
- A control: revert the boundary change and confirm the hyphenated tests go red **by name**.

### Why it matters more this week than last

[...]

PR #80, opened 2026-09-17
Read Show HN as its own source (#69)

### LINKED ISSUE #69: Read Show HN, not just the Hacker News front page
**The Hacker News connection exists and reads the wrong list.**

`src/sources/hackernews/adapter.ts` fetches `topstories` — the front page. **Show HN, where developers launch what they built, is never read.** It is a different endpoint on the same API the adapter already speaks.

This is the cheapest change with the largest effect on how the brief *feels*, and it is the owner's stated direction: startups, tools, things people built, rather than corporate news.

### What to add

- `https://hacker-news.firebaseio.com/v0/showstories.json`, then the existing per-item fetch.
- Recognise the launch shapes in titles: `Show HN: I built…`, `Show HN: Open-source…`, `Show HN: AI…`, `Show HN: Agent…`.
- Keep the existing top-stories source as it is. This is an addition, not a replacement.

### Collect what is already collected, plus

score, comment count, author, HN item id, creation time — most of which the adapter already reads for items.

### What this cannot do yet, and should not pretend to

Velocity — "gaining stars/points fast" — needs history, and nothing in this repo records counts over time. **Do not fake it by comparing against nothing.** A Show HN item arrives with a score; whether that score is rising is a later ticket and belongs with the Signal model.

### Interaction to check before building

Show HN items are a different *kind* of thing from an article, and today everything is a Story with a content type. `DISCUSSION` already exists and the HN adapter already uses it for text posts. Decide explicitly whether a Show HN launch is DISCUSSION, RELEASE, or wants its own type — and if it wants its own, say so rather than forcing it into the nearest one. #41 is the cautionary case: five content types exist that nothing can emit.

### Acceptance

- A real Show HN launch from today appears in the brief, named, with its link to the project and its link to the discussion.
- The existing front-page source is unaffected — same count, same items.
- A control: point the adapter at an empty list and confirm the run reports zero rather than silently succeeding.

zarghammir:
Closes #69.

## In plain words

Hacker News has a second list. The front page is news and argument; **Show HN** is where people post things they have built — a tool, a side project, a model, a game. The app has been reading only the front page, so none of it has ever reached the brief. This adds Show HN as its own source.

## It is a catalogue row, not adapter code

`src/sources/hackernews/adapter.ts` builds its URL as `` `${list}stories.json` `` and already reads `list` from `sources.config`. `list: "show"` reads `showstories.json` today. The ticket's estimate of the work was high in one place and low in two others; this is the place it was high.

## Content type: RELEASE, and the imprecision is written down

Ruled on the ticket, recorded beside the assignment in `seed-data.ts` rather than left to be re-derived.

A Show HN post is a **debut** — a person putting a thing in front of people for the first time. "Release" connotes a new version of a product that already exists. That mismatch is real and it is smaller than either alternative:

- **Not DISCUSSION.** It is what the adapter assigns to front-page self-posts, it carries ranking weight **0**, and it would tell a reader that somebody shipping a thing is a conversation.
- **Not a new content type.** An enum member is a migration, a label, a badge colour, a ranking weight, a place in every exhaustiveness switch and a row on #66 — with the Signal model possibly replacing the Story content-type axis within weeks, making the second build the one that has to migrate rows.

#78 carries the question forward, so the deferral has an owner rather than living in a comment.

## minPoints 3, measured rather than inherited

The front page's threshold is 20. Against the live show list — 120 ids, 120 usable stories:

```
scores   min 2   p25 3   MEDIAN 4   p75 17   max 2100
```

*Measured at 20:05 on 2026-09-16, inside an authorised machine block, and reported at the time. A measurement should carry the circumstances it was taken under.*

| minPoints | kept of 120 |
|---|---|
| 1 | 120 (100%) |
| **3** | **93 (78%)** |
| 5 | 57 (48%) |
| 10 | 39 (33%) |
| 20 | 27 (23%) |

**The median Show HN item has four points.** 20 keeps 27 of 120 — not empty, but a quarter of a list whose whole point is breadth.

3 is derived rather than picked: the observed minimum is 2, so thresholds of 1 and 2 filter nothing at all, and 3 is the smallest value that excludes anything. A threshold calibrated for one list does not transfer to another.

## One adapter change

A self-post on the front page *is* a discussion: someone asking or arguing. A self-post on Show HN is a person launching a thing and describing it rather than linking to it — a debut, not a conversation — so it falls through to the source default instead of being typed `DISCUSSION`.

Confined to the show list, with a **positive control**: a front-page self-post must still type as `DISCUSSION`. Without it the change reads as correct while silently retyping every front-page self-post in the product.

To describe the control accurately rather than by its intent — it drives a *front-page self-post* and asserts nothing about the words "Ask HN". That is the real mechanism: the adapter branches on `it.url` and never reads the title.

## The interaction with the classifier, which is now fixed on main

This branch carried a merge condition here. **It is satisfied and removed** — #79's fix merged as #88, and this branch is rebased onto it.

The finding is worth keeping, because the shape recurs. `git merge-tree` against #57 reported **zero shared files and no conflict**, and the two changes still combined into something worse than either alone.

Once #57 landed, every Show HN item ran through `classifyContentType`: `isDeclaredFact` exempts only `PAPER` and `DISCUSSION`, and this PR files launches under `RELEASE`. **Six branches can type a title and five ignore the `Show HN: ` prefix entirely**, so a launch could be pulled out of RELEASE by vocabulary measured against news headlines.

> `Show HN: an EU AI Act compliance checker` → **Policy**
> `Show HN: I fine-tuned a 7B model for SQL generation` → **Model**

The Model badge is the expensive one. #57 measured it at about **ninety rank places**, so a stranger's weekend project lands above things that matter.

#88 fixed it by leaving a self-announced launch alone. **A clean merge is not a correct one** — file lists agreeing tells you nothing about whether behaviours do.

## Controls

**The zero case.** An empty list must report zero rather than succeeding quietly. The assertion runs **before anything else that can throw** — a diagnostic that only executes when the rest survives is not a diagnostic for the case it exists to catch. It carries its own floor: the run must have requested the list exactly once, because an empty result from a request never made would pass the first assertion while proving nothing.

**The non-zero case**, in the same run, is inherited rather than added. `scripts/sources-check.ts` already computes `ok: items.length > 0 && bad.length === 0`, with the comment *"treat a run that produced zero usable items as a failure, not a quiet success."* Adding the catalogue row puts Show HN under that floor automatically, and `npm run sources:check` is the live run that exercises it.

**No velocity.** An item arrives with a score; whether it is *rising* needs history this repo does not keep. A test asserts no metadata key matches `/velocity|rate|rising|trend|delta|perHour/`, rather than the commit message asserting it.

## A coverage claim that outran its body

An existing seed test was named *"marks Hacker News so that a linked article is not labelled a discussion"* and used `find` — asserting about one row what its name claimed about the catalogue. A second Hacker News source made that gap real. It filters and loops now, with a floor of two.

## Expect a thin live sample, and not because of this branch

Two known defects upstream will suppress results, and neither is this PR's to fix:

- **#73** — `matchesKeyword`'s whole-word branch is defeated by hyphens, so `AI-powered X`, `GPT-5 …` and `LLM-based …` are dropped at the gate. Those are among the most common Show HN title shapes.
- **#76** — the classifier reaches about a third of what it could.

A real launch arriving with no Release badge is more likely to be one of those than this adapter.

## One acceptance criterion is deferred and the other half is absent

The ticket asks that **"a real Show HN launch from today appears in the brief, named, with its link to the project and its link to the discussion."** That is not verified by anything in this PR, and saying so is cheaper than letting it look satisfied.

- **What appears in the brief** is produced by the scheduled Ingest workflow (`npm run worker:once`) against the live catalogue. It is verified on the first run after this merges, not before.
- **The discussion link is not deferred — it is absent**, and calling it deferred undersold it in an earlier version of this section. `StoryCard` has a single `url` and no field for a second, and the adapter's `url: it.url ?? hnUrl` discards the thread **in precisely the case where both links exist**: a launch that links to its project. So the case where both matter is the case where one is dropped. **No post-merge Ingest run will ever satisfy this half**, because nothing is waiting to happen — the field does not exist. It needs a contract change, it is not Show HN specific (the front-page source has been dropping that link since it was built), and it is tracked as **#84**, deliberately out of this PR's scope.

The honest pre-merge substitute is the one below: the source fetches the real list through the real adapter and returns a non-zero count.

## Live check

`npm run sources:check` against all nineteen catalogue sources:

```
OK   hackernews-ai              200      items= 27  (hackernews API)
OK   hackernews-show            200      items= 33  (hackernews API)

19 OK, 0 FAIL
```

**33 items**, clearing both `minPoints 3` and the AI keyword gate, and the existing eighteen sources are unaffected.

I expected this might fail and said so before running it: with only 33% of Show HN matching the AI vocabulary, and #73 dropping hyphenated titles at the gate, zero was a plausible outcome. It was not — and the margin is explained by the earlier measurement rather than luck. 40 of 120 matched the vocabulary and `minPoints 3` keeps 78% of the list; 40 × 0.78 ≈ 31, against 33 observed. The threshold was chosen from that curve and the curve predicted the result.

## Rebased onto main, and checked by meaning

Fifty-three commits landed under this branch: the classifier (#57), the launch guard (#88), source health (#86) and the catalogue change (#89). The rebase produced **no conflict**, which is the condition under which a semantic clash hides — so it was checked by reading rather than by git's verdict.

| Checked | Result |
|---|---|
| Show HN row survived, in the COMMUNITY block beside the front page | yes |
| adapter's show-list branch intact | yes |
| `import-ai` removal (#89) adjacent to this row? | no — ANALYST block, no clash |
| catalogue size | 17 on main, **18** with this row |
| anything asserting an exact catalogue count | none; every floor is `>=` |
| a Show HN title after #88's guard | stays `RELEASE`, which is what this row declares |

**The 33-item measurement was not re-taken, deliberately.** It is `hackernews-show`'s own count against `minPoints 3` and the keyword gate, and does not depend on how many other sources exist. Removing `import-ai` changes the `sources:check` roster and leaves that number untouched, and nothing in the rebase touches that path.

**One thing this makes worse and deliberately does not fix.** The catalogue floor in `content-type.test.ts` moved to 17 when `import-ai` was removed; adding this row makes 17 an understatement again. The second floor, in `seed-data.test.ts`, is still 15. Both pass. **#90 carries the design problem** — a floor hand-edited every time the catalogue changes is the defect, and correcting the number here would hide it for another release.

## Verification status

CI executed lint, typecheck, `npm test`, build and format:check against `d16e90a` — run `35178896453`, all five checks SUCCESS including Vercel. Those are the evidence rather than my reading of the code; nothing in that list was re-run locally, because re-running what CI has already executed buys nothing and costs a shared machine.

`sources:check` is the one thing CI cannot do — it hits nineteen live feeds and would be flaky and impolite on every push — and it is reported above.

One instrument note, since this PR argues elsewhere that a check's result is its exit status: I captured that run's output but not its exit code, having written `${PIPESTATUS[0]}` in a zsh shell where the variable is `$pipestatus[1]`. The script sets `process.exitCode = 1` only when `failed.length` is non-zero and it printed `0 FAIL`, so the status is derivable — but it was derived rather than read, and I am not re-running nineteen third-party fetches to capture a number I can get from the source.


zarghammir:

Reviewed at head `d16e90a`, four files, 0 commits behind `origin/main`. **This is a reading-only review and I want that stated before anything else.** The machine was held by another lane throughout, the author has executed nothing, and CI at this SHA was still `in_progress` with empty conclusions when I looked — so I have verified nothing about whether this code runs. I deliberately did not hunt for anything a compiler or a test run answers.

Everything below comes from reading this branch against #57 at `329e2c5`. The change requested is a reading-level finding, and it does not depend on a run.

The disclosure in "Verification status" is the right way to ship an unexecuted branch, and it is worth saying so: *"the code is reasoned, typechecked and unexecuted, and the check results below are the evidence, not my reading of it."* That is the sentence I would want from any author in that position.

---

## BLOCKING — the #79 analysis is incomplete, and the gap inverts its conclusion

The body's asymmetry argument is that a Show HN item **cannot be promoted** and **can be demoted**, so there is "no upside to weigh against it". The second half is right. The first half is not, and the reason is that the MODEL rule is described by two of its three branches.

From #57's `content-type.ts`:

```js
test: (t) => {
  const family = MODEL_FAMILY.exec(t);
  const launched = matchesAnyPhrase(t, LAUNCH_VERBS);
  const noun = matchesAnyPhrase(t, MODEL_NOUNS);
  if (family && (launched || family.index === 0)) return true;   // described
  if (PARAMETER_COUNT.test(t) && noun) return true;              // NOT described
  return launched && noun;                                       // NOT described
}
```

The body states the rule as *"MODEL requires a launch verb or a model family at index 0"*. That is branch one only. Branch two is **prefix-independent**: `PARAMETER_COUNT` is `/\b\d+(?:\.\d+)?b\b|\b\d+x\d+b\b/i` tested against the raw title, and `MODEL_NOUNS` includes `model`, `llm`, `weights`, `checkpoint`. Neither cares where `"Show HN: "` sits.

> **"Show HN: I fine-tuned a 7B model for SQL generation"** → `PARAMETER_COUNT` matches `7B`, `matchesAnyPhrase` finds `model` → **MODEL**.

That is not a contrived title; it is one of the most common shapes on that list. And `TOOL` is reachable on the same principle — the rule needs a launch verb *somewhere in the title*, and `"Show HN: Launching an open-source SDK"` supplies one. *"Show HN" is not a launch verb* is true and is not the same claim as *no Show HN title contains one*. The two worked examples in the body are both correctly analysed; the general claim above them is not.

**Why this matters more than a wording fix.** #57's own measurement is that MODEL carries the joint-heaviest weight and that a single false MODEL badge is worth about ninety rank places. So the promotion path is not a lesser mirror of the demotion path — it puts a stranger's weekend project in the reader's top ten, which is the failure #57 spent an evening measuring. The sentence "No upside to weigh against it" is doing load-bearing work in the ruling built on top of it, and it rests on a premise that does not hold.

**The consequence for the merge condition, which I was asked to challenge: the condition is right and its fix may be insufficient.** The body describes what lands with #79 as *"the `"Show HN:"` launch-verb fix"*. A fix scoped to launch verbs does not touch branch two. If #79 ships as described, this branch merges into a tree where `"Show HN: a 4B model that runs on a phone"` still classifies as MODEL. I am not asking this PR to fix #79 — I am asking that the analysis it hands to #79 name all three branches, because #79 will be scoped from this text.

## The rejected fix — I attacked it, and the conclusion survives, but not on the stated reasoning

The argument given is that adapter-declared means structured data read from the source, so declaring `RELEASE` would be asserting a policy rather than reading a fact.

**That distinction is weaker than it looks.** The adapter already declares `DISCUSSION` from `it.url ? undefined : "DISCUSSION"` — which is equally a policy mapping ("a self-post is a conversation") applied to a structural fact ("this item has no outbound link"). "This item came from `showstories.json`" is exactly as much a structural fact about the item as "this item has no url". By the stated test, the existing declaration is the same kind of assertion as the rejected one, so the test does not separate them.

**The conclusion is still right, and there is a stronger argument available.** The source default already yields `RELEASE`. Declaring it in the adapter therefore changes no value — it changes only the *provenance*, from `default` to `adapter`. And `adapter` is the one provenance the backfill will never revisit: `item.storedSource === "adapter"` is the skip branch in `reclassify-content-types.ts`. So the rejected fix would buy protection from #79 by making every Show HN row permanently uncorrectable — frozen against #79's own fix, against #76's recall work, and against every future classifier improvement, with no way to tell later which rows were frozen for this reason.

That is the argument I would put in the body. "Dodging a classifier by lying about where a value came from is not a fix" is a good sentence and it is not the reason the fix is wrong.

## `minPoints 3` — the number does follow from the data; the body's sentence does not

Asked whether the table derives the number or decorates it: **the code comment derives it, the PR body overstates it, and the two disagree.**

The body says *"A threshold calibrated for one list silently empties another."* Its own table says 20 keeps 27 of 120. A fifth of the list is not empty, and the strongest claim in the section is contradicted by the row above it.

The comment in `seed-data.ts` is the careful version: *"The front page's threshold of 20 keeps 23% of the list; 3 keeps 78% and still drops the submissions sitting at 2."* That is a real derivation. The minimum score is 2, so 1 and 2 are no-ops — 3 is the smallest threshold that filters anything at all, and the comment says exactly that. The number is measured and defensible.

Since the comment is the artefact that persists and it is the accurate one, this is a body-only problem. Worth fixing because the body is what the next person reads first, and "silently empties" is the kind of sentence that gets quoted later as though it were the measurement.

## The two controls — both hold, and the second is applied rather than cited

**The Ask HN positive control does control.** The change is `it.url || isShowList ? undefined : "DISCUSSION"`. The test drives a url-less story through a `list: "top"` source and asserts `DISCUSSION`. If the show-list condition ever leaked — `isShowList` computed wrongly, the list check dropped, the branch inverted — that assertion returns `undefined` and goes red. And the opposite leak is caught by the test above it. Both directions are covered, and the red would be the named test rather than a side effect.

One note, and it is about the body rather than the test: the body calls this *"asserting front-page Ask HN threads still type as discussions"*, but the test asserts nothing about `Ask HN` — it drives a front-page self-post, which is the actual mechanism, since the adapter branches on `it.url` and never reads the title. The test is **broader** than the body describes and is right; the body describes a narrower thing that isn't what is tested. Safe direction, worth one word.

**The empty-list floor is applied correctly and not merely cited.** `expect(out).toEqual([])` is the first statement after the fetch, with nothing between them that can throw — so the diagnostic genuinely runs in the case it exists for, rather than only when everything else survived. The floor beneath it asserts `calls.filter(u => u.endsWith("/v0/showstories.json")).length === 1`, which closes the hole it names: an empty result produced by a request never made would satisfy the first line while proving nothing. It also pins the *show* URL specifically rather than counting any request, which is the difference between a floor and a gesture.

The honesty about what is inherited is also correct: the "zero is a failure, not a quiet success" behaviour lives in `scripts/sources-check.ts`, the catalogue row puts Show HN under it automatically, and the body says so rather than claiming the test does it.

## Credited

- **The `RELEASE` decision and its cost accounting** — not DISCUSSION because ranking weight 0 would tell a reader that shipping a thing is a conversation; not a new enum member because of the migration, the badge, the weight, the exhaustiveness switches and the possibility that the Signal model replaces the axis. Deferred with an owner on #78 rather than left in a comment, which is the pattern this project has had to learn twice.
- **The velocity test.** Asserting that no metadata key matches `/velocity|rate|rising|trend|delta|perHour/` puts the absence in the suite instead of the commit message. A delta against no history is a number that looks measured and is not, and the test says so.
- **The coverage-claim fix.** A seed test named *"marks Hacker News so that a linked article is not labelled a discussion"* that used `find` was asserting about one row what its name claimed about the catalogue; adding a second Hacker News source made the gap real. Filtering, looping and flooring at two is the right repair, and noticing it at all is the harder part.
- **Naming #73 and #76 as the reason a live sample will look thin**, before anyone runs it and concludes this adapter is broken.

## Not verified, and what I would run

Nothing here is a claim I could settle by reading, and none of it changed the verdict:

- That the suite passes, that `adapter.test.ts` compiles, that the widened transport regex `/\/v0\/[a-z]+stories\.json$/` still matches every URL the existing front-page tests depend on. `npm test` answers all three in under a minute.
- That the live show list still has the shape the table records — 120 ids, median 4. `npm run sources:check` after seeding.
- The MODEL and TOOL reachability above is derived from reading #57's rules, not from executing them. Three `npx tsx -e` calls against `classifyContentType` with the titles I quoted would confirm or refute it in seconds, and I would run exactly that before #79 is scoped from this text.

zarghammir:
**Execution note — the three derivations in my review above are now confirmed by running them.**

My verdict listed these as derived from reading #57's rules rather than from executing them. I was granted a narrow slot for exactly that and nothing else, so here is the result, against #57 at `329e2c5`, calling `decideContentType(undefined, title, "RELEASE")` — `RELEASE` being the Show HN source default this PR adds:

```
MODEL       / classifier   <- Show HN: I fine-tuned a 7B model for SQL generation
TOOL        / classifier   <- Show HN: Launching an open-source SDK
REGULATION  / classifier   <- Show HN: an EU AI Act compliance checker
```

All three fire. The promotion paths are real, not theoretical: `MODEL` via `PARAMETER_COUNT.test(t) && noun`, which never looks at where the `"Show HN: "` prefix sits, and `TOOL` via a launch verb sitting later in the title.

**And the two worked examples in the PR body are correct** — I ran them in the same evaluation as controls, because a finding that contradicted them would have meant my reading was wrong in a different way:

```
RELEASE     / default      <- Show HN: I built an open-source CLI
RELEASE     / default      <- Show HN: Gemma 4 fine-tuning toolkit
```

Both fall through to the source default, exactly as the body says. So the analysis is right about every example it gives and wrong about the general claim above them — which is the finding, stated precisely: **"Show HN is not a launch verb" is true; "no Show HN title contains one" is the claim the conclusion needs, and it is false.**

Nothing else was run. No suite, no typecheck, no build, no browser — those remain unverified exactly as my review states.

zarghammir:
REVIEW VERDICT: APPROVE

Re-review at head **`3bb6339`**, 0 commits behind `origin/main`. Baseline read at this moment: run `35262934126` has `headSha = 3bb6339`, matching the PR head; four check-runs `completed/success`; `MERGEABLE/CLEAN`.

Fifty-three commits landed underneath this and the rebase produced no conflict, which is the configuration that deserves the most attention rather than the least. The six checks done by reading are the right six. I re-derived the ones that carry weight.

## The two routes agree for the canonical form — and they already diverge outside it

This was the one worth doing independently, and the answer is narrower than "they arrive at the same answer by different routes".

- **Route A:** `hackernews-show` declares `defaultContentType: "RELEASE"`, and the adapter returns `contentType: undefined` for every show-list item (`it.url || isShowList ? undefined : "DISCUSSION"`), so the source default is what reaches `decideContentType`.
- **Route B:** #88's guard returns `declared` when `/^\s*show hn\s*:/i` matches.

For a title in the canonical `Show HN: …` form both give `RELEASE`, by different paths, and that is the good property. **But the guard is anchored to a stricter pattern than the list it protects.** A show-list item whose title does not match that regex — `Show HN – …` with a dash rather than a colon, `ShowHN:` without the space, or a title a moderator rewrote — skips the guard entirely, and then `isDeclaredFact` does not cover `RELEASE`, so the six branches run and can retype it.

So the divergence is not waiting on a future change to either route. **It exists now, for a subset of the same list.**

**No test would notice.** Every show-list adapter test uses the `story()` helper's default title, `OpenAI ships a new model ${id}` — which lacks the prefix, and happens not to trip any branch: "ships" is not in `LAUNCH_VERBS`, "OpenAI" is not in `MODEL_FAMILY`, and there is no parameter count. The tests pass either way, by luck rather than by coverage, and nothing combines a show-list item with a non-prefixed title that *would* classify.

**This is a residual rather than a regression, and I am not blocking on it.** Before #88 every show-list item was exposed; now the canonical majority is protected and a non-canonical remainder is not. That is strictly better. It is worth naming because the fix is available and unused: **the fact that an item came from `showstories.json` is strictly more reliable than a regex over its title**, and #88 deliberately declines to use it — *"the classifier never learns where a title came from"*. That is a defensible design choice, but it is now load-bearing for a second PR, and the decision should be made knowingly rather than discovered when a dash-form title is promoted ninety places.

## The merge-condition rewrite still describes the code

*"`merge-tree` reported zero shared files and no conflict, and the two changes still combined into something worse than either alone."* The mechanism is still visible in the tree rather than only in the history: `isDeclaredFact` remains `PAPER || DISCUSSION`, so `RELEASE` is not a declared fact, so without #88's guard this row's `RELEASE` would still be overridable by the rules. The sentence is a claim about the code and the code still supports it.

## The floor: I agree, and the drift is itself the argument for #90

Both floors pass at 18 — `seed-data.test.ts:16` at `>= 15`, `content-type.test.ts:229` at `>= 17`. Nothing is red, and correcting the 17 here would be the third adjustment to that number in a day.

**I agree it should not be fixed here, and would put the case more strongly than the body does.** That number has now moved in *both directions* within one day from two changes that had nothing to do with it: down when `import-ai` left, and into understatement when this row arrives. A floor that tracks the measurement rather than the failure is not a floor, it is a mirror — and the fact that it drifted twice, in opposite directions, from unrelated work **is the evidence #90 needs.** Bumping it to 18 would erase exactly that evidence and leave the number looking maintained.

## The 33 was correctly not re-taken, and I checked the claim underneath it

The decision rests on the item count depending on `hackernews-show`'s own path rather than on the catalogue. I verified the rebase did not touch it:

- no commits on `src/pipeline/normalize/keywords.ts` between the merge-base and `origin/main`;
- no commits on `src/sources/hackernews/`;
- `DEFAULT_AI_KEYWORDS` is 40 entries on both;
- `keywords.ts` is **byte-identical** between `origin/main` and this head.

The AI gate, the adapter and the vocabulary are all unchanged, so `minPoints`, `limit` and the keyword filter all behave as they did when 33 was measured. The number stands and re-measuring would have cost a slot to reproduce a figure nothing could have moved.

## `seed-data.ts` merged correctly

I read it rather than trusting the clean merge. The `import-ai` removal note sits in the **ANALYST** block immediately above `interconnects`; the new row sits in the **COMMUNITY** block after `hackernews-ai`. They are in different sections with no adjacency, which is why a textual merge had nothing to resolve — and why the clean result is trustworthy here rather than merely quiet.

## The process note generalises, and it is worth stating as a rule

The body edit aborted on a bad anchor — a heading belonging to #88's body, in a session editing two — and nothing was written. But `gh pr edit` ran anyway, because it was a separate command rather than chained to the assertion's exit status.

**The guard protected the content and not the sequence**, and that is the general form: an assertion guards only what is downstream of its *exit status*, not what is downstream of it *in time*. A check and the action it authorises have to be joined by `&&`, or the check is advisory. That is the same shape as reading a pipeline's status and getting `tee`'s — the test ran, returned the right answer, and nothing consumed it.

## Verified

Both routes and where they part · the merge-condition mechanism still present in the tree · both floors and their values at 18 · the four independent facts behind the untouched item-count path · the merged `seed-data.ts` region · baseline pinned to the SHA.

## Not verified

The 33 itself — I checked that nothing could have changed it, not that it was taken correctly. And the live behaviour of any of this: the scheduled ingest has still never completed a run, so the show list has never been fetched by the deployed app.

PR #86, opened 2026-09-17
Tell a failing source apart from a quiet one (#38)

### LINKED ISSUE #38: A source that fails on every hosted run is invisible (import-ai returns 403 to GitHub runners)

zarghammir:
Closes #38 — the detection half. **Does not touch the Settings source list**, which is #74's surface.

## In plain words

One of the eighteen sources, `import-ai`, fails on **every single hosted run**. Substack refuses the datacenter IP that GitHub's runners use; the same feed works fine from a laptop. The scheduled job finishes green with `1 failed` in a log nobody reads.

So the app has been quietly running on seventeen sources, and **nothing anywhere could tell you the difference between a source that has been dead for a week and a source that simply had nothing to say.** This makes those two states distinguishable.

## The exit code is unchanged, on purpose

A scheduled run that goes red every time one publisher blocks a datacenter is a notification people mute — and once muted, a real outage goes unseen along with it. This PR makes the failure **visible**, not **louder**. That distinction is the whole design.

## What it adds

`GET /api/sources` now carries two fields:

```json
{ "health": "FAILING", "consecutiveFailures": 7 }
```

Computed from `ingest_runs` history rather than stored — the data already exists, so there is no new plumbing and nothing that can drift out of sync with the runs it describes. The Settings list can render it later without a second change here.

### Three states, not two

`OK` · `FAILING` · `UNKNOWN`

A source that has never completed a run is neither working nor broken. Reporting it as healthy would be **this ticket's own defect moved into the reporting layer** — an absence read as a clean result.

### The threshold is a duration in disguise

Three consecutive failures. The schedule runs every thirty minutes, so:

| N | time to notice | |
|---|---|---|
| 1 | 30 min | any single 500 from any of eighteen feeds flags it |
| **3** | **90 min** | a blip survives; a source that died overnight is flagged by morning |
| 6 | 3 hours | |
| 48 | 24 hours | a source dead all night still reads healthy |

The failure this ticket is about fails *every* run, so any threshold catches it and speed is the only question. What sets the floor is the other direction: at N = 1, eighteen feeds polled every half hour would flag constantly, and an alert that cries wolf gets muted — **the exact failure the unchanged exit code is there to avoid.** Reintroducing it one layer up would defeat the point.

## Two things the query has to get right

Both tested against a real database, because the pure classifier cannot check the SQL that feeds it.

**A source that has never once succeeded.** Failures are counted from a `-infinity` bound rather than a null-guard. With a null-guard, a source with no successful run to count from counts *zero* failures and reads healthy — and that is precisely `import-ai`'s situation, refused since the day it was added. The test is named for it.

**A run that has started but not finished** is neither a success nor a failure. Counting an in-flight row as a success would mark a failing source healthy the moment the next pass began — and the next pass begins every thirty minutes, so the field would read `OK` for most of its life.

## The floor

`summariseSourceHealth([])` returns `{ kind: "nothing-examined" }`, not `{ failing: 0 }`.

Zero failures over an empty catalogue is arithmetically true and reads as a clean bill of health. Rather than warn about that in a comment, the empty case is **a different variant of the type**, so a caller cannot reach a count without first passing through the branch that says nothing was looked at.

## Health is per-source, and the suite could not see that

Found in review. Every health test created exactly **one** source — and with one source, a correlated count over `ingest_runs` and a global count are the same number. All three `source_id = s.id` correlations in the query could be deleted and the suite would stay green.

In production eighteen sources share that table. A lost correlation marks **all eighteen** FAILING the moment one crosses the threshold, or lets one healthy feed clear everyone else's count — this ticket's own defect, restored across the whole catalogue. Per-source is the *defining* property of the feature, and the suite was testing a coincidence of the fixture.

Three tests rather than one fixture, because the three correlations need three different arrangements and no single one exposes them all:

| The correlation | The fixture that catches it | What breaks without it |
|---|---|---|
| on the failure count | a neighbour that succeeded **before** the failures | the neighbour is handed all three failures and reads FAILING |
| on the last-success lookup | a neighbour that succeeds **after** them | the blocked feed's failures fall before a success that is not its own, and it reads OK |
| on the completed-run count | a source with no runs beside one with many | a source nobody has ever fetched inherits its neighbour's runs and reads OK instead of UNKNOWN |

## Deletion controls

Each must redden the **named** test. Run on the whole suite with a database, tree restored between each.

| # | Mutation | Test that must redden |
|---|---|---|
| 1 | remove the UNKNOWN state — treat zero completed runs as OK | `reports a source that has never run as UNKNOWN, not healthy` |
| 2 | replace the negative-infinity bound with a null-guard | `reports a source that has never once succeeded as FAILING` |
| 3 | count an in-flight run as a success | `does not let a run that has only started clear the count` |
| 4 | return a zero count for an empty catalogue | `cannot report zero failures when it examined nothing` |
| 5 | drop the `source_id` correlation on the **failure count** | `counts each source's failures separately, not the table's` |
| 6 | drop the `source_id` correlation on the **completed-run count** | `keeps a source with no runs UNKNOWN while another source has many` |
| 7 | drop the `source_id` correlation on the **last-success lookup** | `does not treat another source's success as this source's` |

**Control 5 exists because of the finding above.** Against the previous revision it would have reddened *nothing* — and a control that reddens nothing reads as reassurance while proving the opposite.

### Results

Measured on the whole suite with a database, 313 tests, 0 skipped, tree restored between each and verified clean afterwards.

| # | Named test reddened | Total red | Also reddened |
|---|---|---|---|
| 1 | ✅ `reports a source that has never run as UNKNOWN, not healthy` | 4 | the three other tests that assert UNKNOWN |
| 2 | ✅ `reports a source that has never once succeeded as FAILING` | 5 | every test whose source never succeeded — the null-guard reports them all OK |
| 3 | ✅ `does not let a run that has only started clear the count` | 1 | — |
| 4 | ✅ `cannot report zero failures when it examined nothing` | 1 | — |
| 5 | ✅ `counts each source's failures separately, not the table's` | 2 | `keeps a source with no runs UNKNOWN while another source has many` |
| 6 | ✅ `keeps a source with no runs UNKNOWN while another source has many` | 1 | — |
| 7 | ✅ `does not treat another source's success as this source's` | 1 | — |

Every control reddened its named test, none reddened nothing, and none failed to compile. Baseline afterwards: **313 passed, 24 files, 0 skipped.**

**Controls 6 and 7 exist because control 5 was not the sweep it looked like.** There are *three* `source_id` correlations, and control 5 drops one. The completed-run test did go red under it — but for a different reason than the rule it was written for, and a test reddening under a mutation aimed at a different rule is evidence that the two are entangled in the fixture, not that its own rule is controlled. Controlled individually, each correlation reddens exactly the test written for it, alone.

Control 2's spread is worth reading rather than skimming: replacing the bound with a null-guard turns *every* source that has never succeeded into a healthy one, which is why five tests notice. That is the defect this feature exists to detect, and the obvious implementation is the one that reproduces it.

## A test whose name outran its body

`source-health.test.ts` carried a test named *"does not let an in-flight run make an unknown source look healthy"*. It passed the same input as the test three lines above it and asserted something weaker.

`classifySourceHealth` has no notion of an in-flight run — that distinction is made entirely in SQL, by counting only runs that finished either way. So the name described a path the body could not reach, and the real case hid behind it. It is removed, with a pointer to where the case is now tested against a database.

## Scope, and what is deliberately not here

- **The Settings source list is untouched.** That is #74, in review now; a second lane in those files would collide. The field is on the data so the screen needs no second change.
- **The `import-ai` decision is not made here.** Options below, for a product call rather than a lane's.

### `import-ai`: three options, not a recommendation

| Option | Cost |
|---|---|
| Accept the loss on hosted runs | One ANALYST source permanently missing from the hosted brief. Now a *chosen* loss rather than an invisible one — which is what this PR changes about it. |
| Find another feed | Unknown until someone looks; the newsletter may publish elsewhere. |
| Fetch by another route | A proxy, or a self-hosted runner without a datacenter IP. Both add operational surface for one source. |

Whichever is chosen, the detection stands on its own: the next source a publisher blocks will be visible without anyone having filed a ticket about it.

## Merge with #57

Rehearsed against `329e2c5`: **zero shared files, `git merge-tree` exit 0, no conflicts.** #57 touches `src/pipeline/normalize/*`, `run.ts` and `schema.ts`; this lands in `src/api/` and `docs/`. I also looked for a semantic interaction and found none — #57 changes how an item's *content type* is decided, and nothing here reads or writes content type.

## Verification

CI runs lint, typecheck, `npm test`, build and format:check, and provisions postgres with `DATABASE_URL` set — so the database tests above execute there. **Nothing in this PR needs a local machine**, and nothing was re-run locally that CI already covers.


zarghammir:
REVIEW VERDICT: APPROVE

Reviewed at head `b7f00f5`, 0 commits behind `origin/main`, `MERGEABLE/CLEAN`, five files. **Read-based review: I did not take the machine, and nothing here needed it** — the two claims that could only be settled by execution are settled by CI, and I read the run rather than its conclusion.

This is careful work and the reasoning in it is the substance rather than decoration. Both of the things the query has to get right, it gets right, and I checked each against the counterfactual rather than against the comment.

## CI executed the database tests — confirmed from the run, not the badge

The failure mode worth ruling out first is a gated suite that ran nothing and agreed with itself. It did not:

```
Unit tests  env DATABASE_URL: ***localhost:5432/ai_radar_test
            Test Files  24 passed (24)
                 Tests  310 passed (310)
occurrences of "DATABASE_URL is not set": 0
```

`310 passed (310)` with no skipped segment, and zero instances of the loud warning `api.test.ts` prints when it is ungated. The `withDb` suites ran.

## The negative-infinity bound holds, and the counterfactual is the proof

```sql
and f.started_at > coalesce(
      (select max(ok.started_at) from ingest_runs ok
         where ok.source_id = s.id and ok.error is null and ok.finished_at is not null),
      '-infinity'::timestamptz)
```

The claim is that a null-guard would read a never-successful source as healthy. It would, and precisely: without the `coalesce`, the scalar subquery returns NULL, `started_at > NULL` evaluates to NULL rather than TRUE, **no rows are counted**, and `consecutiveFailures` is 0. Then `completedRuns` is non-zero — failures count as completed — so `classifySourceHealth` returns **OK**. A source refused on every run since the day it was added reads healthy. The obvious implementation does reproduce the defect the PR exists to detect, exactly as the body says.

The bound is also safe from the other direction: `started_at` is `.notNull().defaultNow()`, so the comparison can never be NULL for a real row. Had it been nullable, failures with a null start would have been silently dropped and the whole count would have been an absence-shaped hole.

## The in-flight run is excluded on both paths, which is the part that could have been half-done

Two places had to get it right and both do:

- `completed_runs` counts `r.error is not null or r.finished_at is not null`, so a row with neither — started, not finished, no error — is not counted. A source whose first pass has only just begun stays UNKNOWN rather than becoming OK.
- The last-success subquery requires `ok.error is null **and ok.finished_at is not null**`. This is the one that matters most and is easiest to omit: without the second condition an in-flight row would become the most recent "success", the bound would jump forward, and the failure count would reset to zero at the top of every thirty-minute pass. Green for a moment every half hour, forever, exactly as the body describes.

The database test `does not let a run that has only started clear the count` exercises the second. Nothing exercises the first at the SQL level — see the gap below.

## The floor is structural, not conventional

`SourceHealthSummary` is a discriminated union whose `nothing-examined` variant has no count fields at all. A caller cannot write `summary.failing` without narrowing on `kind` first — it is a type error, not a convention. This is a floor that cannot be forgotten rather than one that must be remembered, and the distinction is real.

Worth stating for the record: it is a **compile-time** floor. Types erase, so a hypothetical untyped consumer would read `undefined`. Everything in this repo is TypeScript, so it holds where it needs to.

## UNKNOWN survives to the wire

No path collapses it. `classifySourceHealth` returns it as a distinct value, `summariseSourceHealth` counts it in its own field and the test asserts `ok: 0, failing: 0, unknown: 2` rather than folding it either way, and `src/app/api/sources/route.ts` is a straight `json({ sources: await listSources(...) })` with no mapping in between. The string reaches the client as `"UNKNOWN"`.

## Scope held, and the #57 claim checks out

**Nothing under `src/app` or `src/components`** — the five files are `src/api/` and `docs/`. The Settings source list is untouched, so there is no second lane in #74's files.

The semantic claim about #57 is the one worth checking, since that is where a `merge-tree` exit 0 misled us both on #79. It holds: `catalogue.ts` does join `raw_items`, but only `on ri.source_id = s.id` to reach `stories` — it reads no content type, and `source-health.ts` touches neither. #57 adds a column to `raw_items` and changes how that column is decided; neither affects a join that never selects it. Offering it as a checked claim rather than an assumption was right, and the check survives.

---

## Finding — three `source_id` correlations are unexercised, and a deletion control would show it

`beforeEach` runs `TRUNCATE_ALL`, and every health test creates **exactly one source**. So for all six of them the `ingest_runs` table contains rows belonging to that one source only.

That means the three correlation predicates — `r.source_id = s.id`, `f.source_id = s.id`, `ok.source_id = s.id` — are load-bearing in production and invisible to the suite. **Delete any one of them and every test still passes**, because with a single source the correlated count and the global count are the same number.

The production consequence is not small: eighteen sources share `ingest_runs`. A lost correlation on the failure count marks all eighteen FAILING the moment import-ai crosses the threshold; a lost correlation on the success subquery lets one healthy feed clear everyone's count, which is this ticket's own defect restored across the whole catalogue.

The code is correct as written — this is a coverage gap rather than a defect, which is why it does not block. One test closes it: two sources in the same table with different histories, each asserted independently. Something like a failing `import-ai` beside a succeeding `openai-blog`, asserting FAILING/3 and OK/0 in the same call. That single fixture makes all three predicates load-bearing at once.

**On the four deletion controls:** I cannot see the four that were chosen — the body names none, and I am not going to guess at a list I have not read. If they are recorded somewhere, point me at them and I will say. The four I would choose, and which I think are the right four:

1. **Drop the `coalesce(..., '-infinity')`** → the `never once succeeded` test must redden, and nothing else. This is the line the PR turns on.
2. **Drop `ok.finished_at is not null`** from the last-success subquery → `does not let a run that has only started clear the count` must redden, alone.
3. **Drop one `source_id = s.id` correlation** → today this reddens **nothing**, which is the finding above. After the two-source test it should redden that test and only that one.
4. **Change `summariseSourceHealth([])` to return `{ kind: "examined", examined: 0, ok: 0, failing: 0, unknown: 0 }`** → `cannot report zero failures when it examined nothing` must redden. Worth running even though the union makes it a type change, because it proves the floor is the type rather than the test.

Control 3 is the one that currently proves the least and would prove the most.

## Finding — a test named for a path it cannot reach

`source-health.test.ts:17`, *"does not let an in-flight run make an unknown source look healthy"*, calls:

```ts
classifySourceHealth({ completedRuns: 0, consecutiveFailures: 0 })
```

That is the **identical input** to the test three lines above it, with a weaker assertion — `not.toBe("OK")` where the first says `toBe("UNKNOWN")`. The pure classifier has no notion of an in-flight run; whether one is excluded is decided entirely in SQL. So the test's name describes a path its body cannot exercise, and it adds no coverage over its neighbour.

The corresponding real case is also untested: **a source whose only `ingest_runs` row is in-flight**. It should be UNKNOWN, and by construction it is — `completed_runs` excludes it — but nothing asserts it. Renaming the pure test to what it checks, and adding that one-row database case, would close both halves.

## Not verified, and what it would take

Nothing here needed the machine and I did not take it. For completeness: the four controls above are the only thing I would run, at roughly a minute for the suite each. CI already covers the assertions themselves, which is why this review did not need a slot.

I make no judgement on the `import-ai` decision itself — three options with costs are the right output for a product call, and the detection stands regardless of which is chosen.

zarghammir:
REVIEW VERDICT: APPROVE

Re-review at head `940ba44`, 0 commits behind `origin/main`, `MERGEABLE/CLEAN`, four check-runs all `completed/success`.

The finding is closed properly, and the response went past it: it counted the family instead of trusting a summary of it, found a third correlation neither of us had named, and controlled all three individually. Control 5 exists because it would have reddened nothing against the previous revision — and a control that reddens nothing reads as reassurance while proving the opposite.

## The count is three, and I counted it rather than accepting it

The health computation carries exactly three `source_id` correlations:

```
catalogue.ts:72   r.source_id  = s.id     completed-run count
catalogue.ts:78   f.source_id  = s.id     failure count
catalogue.ts:82   ok.source_id = s.id     last-success lookup
```

Three, matching controls 5, 6 and 7. Nothing else in the health path correlates.

## The two-source fixtures express a distinction rather than merely having two rows

That was the failure mode worth checking — a fixture that looks plural and is effectively singular for the query under test. These are not:

- **`does not treat another source's success as this source's`** — B's success is at 11:00, *after* all three of A's failures at 09:00–10:00. The ordering is load-bearing: without `ok.source_id = s.id` the lookup finds B's success, A's failures all fall before it, A counts zero and reads OK. A fixture where B succeeded first would have been two rows and no distinction.
- **`keeps a source with no runs UNKNOWN while another source has many`** — the second source has *no runs at all*, which is the only shape that isolates the completed-run correlation: without `r.source_id = s.id` the untouched source inherits its neighbour's three completed runs and stops being UNKNOWN.

Both assert the neighbour's state as well as the subject's, so a mutation that broke both sources equally could not pass either.

## The controls table read as evidence

Every control reddened its named test, none reddened nothing, none failed to compile. Two rows are worth more than a tick:

**Control 2, five red.** Replacing the negative-infinity bound with a null-guard turns *every* never-succeeded source healthy, not just the one the test is named for. That spread is the finding of the whole PR restated as a measurement: the obvious implementation reproduces the defect, and it reproduces it catalogue-wide rather than in one row.

**Controls 5 → 6 → 7, the entanglement.** Control 5 reddens two tests, including the completed-run one — which is exactly why it looked like a family sweep and was not. A test going red under a mutation aimed at a different rule is evidence that the two are entangled in the fixture, not that its own rule is controlled. Under control 6 that same test goes red for its own reason, alone. **"A control went near this rule" and "this rule is controlled" are different claims**, and the only thing that separated them was counting.

## The removed test

The one whose name outran its body is gone rather than renamed, with a pointer to where the case is now tested against a database — `keeps a source whose only run is still in flight UNKNOWN`. That is the better repair: the pure classifier genuinely cannot express in-flight, so a renamed version would still have been a test in the wrong file.

---

## One observation — the query has a fourth `source_id` correlation, and it is uncontrolled

Not a defect in this PR, and not in the health path. `listSources` also carries:

```sql
left join raw_items ri on ri.source_id = s.id
```

which feeds `story_count`. It is pre-existing — unchanged by this PR — but it is the same shape in the same statement, so the family is four rather than three once you draw the boundary at the query instead of at the feature.

**Measured rather than reasoned.** I replaced it with `left join raw_items ri on true` and ran the whole suite against a database:

```
baseline          313 passed, 24 files, 0 skipped
correlation gone  313 passed, 24 files, 0 skipped   — nothing reddened
tsc               0 errors
```

Dropping it reddens nothing. The consequence if it ever broke is that every source reports the same `storyCount` — the total across the catalogue — which is a plausible-looking number that nobody would query. The only `storyCount` assertions in the suite are on `listTopics`, not `listSources`.

I am not asking for it here: it is pre-existing, and the field's own comment calls it *"a hint for ordering chips, not a statistic"*, which is a fair reason for it never to have been asserted. It is worth naming because this PR has just built the thing that would close it — before today there was no two-source fixture in that file, and now there are two. Adding `storyCount` to either of the new per-source assertions would bring the fourth correlation under the same guard as the other three, at the cost of one line.

## Verified myself

- The three health correlations, counted in the source.
- Baseline at this head: **313 passed / 24 files / 0 skipped**, matching the reported figure.
- The fourth-correlation control above, with the mutation confirmed to have landed before the result was read, and the tree confirmed clean after.
- Both two-source fixtures read for whether their shape isolates the rule, not just for their assertions.
- Four check-runs at the exact SHA, conclusions read individually.

**One instrument note, because it nearly became a false finding.** My first `tsc` under the mutation reported an error — `.next/types/validator.ts: Cannot find module '../../src/app/welcome/page.js'`. That is a phantom: I had switched a worktree from #74's head, `.next/types` is gitignored and not branch-aware, and `/welcome` exists only on the other branch. After `npx next typegen`, tsc is clean at 0 errors. The suite result above is unaffected, since vitest does not read those types.

PR #88, opened 2026-09-17
Let a Show HN launch keep the type it announced (#79)

### LINKED ISSUE #79: "Show HN:" must be a launch verb — #57 and #69 are correct apart and harmful together

zarghammir:
Closes #79.

**Show HN** is the part of Hacker News where people post things they built. The app reads that list now, and every item off it arrives labelled **Release** — somebody shipped a thing.

Then the classifier reads the headline and can change that label. It was taught on news headlines, where "a 7B model" means a lab put out a model and "AI Act" means a government did something. **On a launch list those same words describe what the person built**, not what kind of story it is. So "Show HN: I fine-tuned a 7B model for SQL generation" was being filed as a model release, and "Show HN: an EU AI Act compliance checker" as a policy story.

**This makes the classifier leave those alone.** A headline that opens by announcing a launch has already said what it is.

## Why getting this wrong is not a small thing

A wrong **Model** badge is not just a wrong word. Model is the joint-heaviest weight in the ranking, and #57 measured what that is worth: **about ninety places**. So a stranger's weekend project lands in the top ten of the morning brief, above things that matter, and the reader has no way to tell why.

That is the direction this fix protects. A wrong label is a nuisance; a wrong promotion takes someone's place.

## Six branches can label one of these, and five ignore the prefix

Enumerated before writing anything, because the ticket's first version described one branch and drew a conclusion about all of them.

| Branch | Reaches a Show HN launch? | Example |
|---|---|---|
| MODEL 1 — family beside a launch verb | via a verb anywhere in the title | `Show HN: Announcing Gemma 4 fine-tuning on a laptop` |
| MODEL 2 — parameter count beside a model noun | **prefix-independent** | `Show HN: I fine-tuned a 7B model for SQL generation` |
| MODEL 3 — launch verb beside a model noun | **prefix-independent** | `Show HN: Introducing a small language model` |
| TOOL — launch verb beside a tool word | **prefix-independent** | `Show HN: Launching an open-source SDK` |
| REGULATION — one phrase | **prefix-independent** | `Show HN: an EU AI Act compliance checker` |
| BUSINESS — one phrase | **prefix-independent** | `Show HN: I raised a seed round for my side project` |

Each was decided separately and each is rejected, for one reason written once in the source: **the vocabulary describes subject matter, and on this list subject matter is not kind.**

## The fix the ticket asked for would have made it worse

#79 first said to add `"Show HN:"` to `LAUNCH_VERBS`. That **leaves MODEL 2 untouched** — the parameter-count branch never reads a verb — and **opens MODEL 1, MODEL 3 and TOOL to every title on the list**, because all three need only a launch verb somewhere and every item would now have one.

It would have increased promotion. Promotion is the expensive direction. The ticket's own correction withdrew it.

## The version that works and is still wrong

The adapter could declare `RELEASE` outright. The classifier would never run and nothing could be mislabelled. Two lines.

It writes `adapter` into the provenance column for rows no adapter ever declared, and **`adapter` is the one provenance the backfill never revisits** — so every Show HN row would be frozen against every future improvement to these rules, with nothing in the data recording why. Rejected in #80 on a weaker argument; this is the one that holds.

This returns the *declared* type rather than a literal `RELEASE`, so provenance stays `default` and the rows stay reachable.

## Measured, not claimed

| What | Result |
|---|---|
| `npm test` (CI, postgres provisioned) | see checks below |
| `npx tsc --noEmit` | 0 |
| lint / format:check | 0 / 0 |
| MODEL precision on the measured corpus | unchanged, **by construction** |

**The precision claim is structural, not a sample, and the structure is what carries it.** The guard is anchored to the start of the title and it is the only new code path, so it cannot fire on anything that does not open with the announcement — and nothing arriving from eighteen news feeds does. **A rule that cannot fire cannot move a number.** #57's 90% stands untouched for everything that is not a launch.

A test also checks that no title in this file's corpus trips the guard. That is a 20-title sample rather than #57's 516, so it is a second look and not the argument.

### The controls

Whole suite with a database, tree restored between each, baseline green before and after at **400 passed**.

| Mutation | Named test that reddened | Total red | What else fell |
|---|---|---|---|
| remove the guard | all six `keeps a launch as RELEASE` cases | 8 | both provenance cases |
| unanchor it (`^` dropped) | `is anchored: a headline that merely mentions Show HN is classified normally` | 1 | — |
| return a literal `RELEASE` | `returns whatever was declared, not RELEASE specifically` | 2 | the provenance case, after the fix below |

**The third control found a hole in one of these tests, which is why it is worth running them rather than listing them.** `returns the declared type rather than a literal, so provenance stays default` declared `RELEASE` in its fixture. So the literal and the declared value were **the same string**, and a mutation returning a literal `RELEASE` left it green — while its name claimed to guard exactly that distinction. A test that could not tell apart the two things it was named for.

Declaring `NEWS` makes them distinguishable. The same control now reddens it alongside its sibling, where before it reddened only the sibling.

## How many items this fixes today: none

Measured on the live list, twice over one cached corpus so both passes saw identical titles.

| | Retyped away from RELEASE |
|---|---|
| with the guard | **0 of 34** |
| with the guard removed | **0 of 34** |

**Not one item on today's Show HN list trips any of the six branches.** So this is **protective, not corrective** — it does not repair anything currently wrong on the page.

What it does: it prevents a class of error worth about **ninety rank places** when one lands, on a list where the shapes are common enough to be unit-testable. Six tests fire those branches on bare titles; today's thirty-four items simply did not carry a parameter count beside a model noun, or "AI Act", or a launch verb beside a tool word.

**What the zero does and does not establish**, since a reader will reasonably ask. Zero events in 34 trials puts the 95% upper bound at roughly 3/34 — about **nine percent**, or **around three items a day** at this list's volume. So it shows the rate is **not high**. It does not show it is **low**.

Two explanations fit equally well and this sample cannot separate them: either Show HN vocabulary genuinely overlaps less with news vocabulary than the branch analysis implies, or the rate is simply low enough that thirty-four items happened not to contain one. **#56 will distinguish them.**

Worth stating precisely, because the looser version — *"the vocabulary is narrower than expected"* — reads as an explanation and is really a guess, and it argues against this PR. If the true rate sits near that upper bound, the protective case is **stronger** than the zero makes it sound: several avoided ninety-place errors a week rather than one hypothetical.

One day is one sample. A rate across several days belongs with **#56**, which already exists to revisit the MODEL weight against real traffic — one instrument, one sitting, rather than the same corpus fetched twice by two people. It is not measurable yet regardless: the scheduled ingest has never completed a run.

## A defect in my own analysis, which is why the branch table is above

I wrote the analysis this ticket was built from, and **I got it wrong.** I said a Show HN item "can never be promoted". I had described one branch of a three-branch rule and generalised from it, with two worked examples underneath that were both correct.

**That is the hard shape: the examples corroborate the sentence without testing it.** Real evidence, placed beneath a conclusion it does not reach. The PM verified all four of my claims and none of them was false — the gap was that they were not all the claims there were.

The branch table exists because enumerating is the cheap check that would have caught it, and it costs ten seconds.

| Decision | Who | Why |
|---|---|---|
| A launch is `RELEASE`, not a new content type | PM ruling on #69 | An enum member is a migration, a label, a colour, a weight and every exhaustiveness switch, on the eve of a model that may replace the axis. #78 carries the question. |
| Merge order: #80 does not enter a tree with #57 without this | PM ruling on #79 | No window in which the interaction is live. |

## It applies wherever the title says it, which is deliberate

The guard reads the title and never learns which source sent it. So a **front-page** Hacker News item whose title opens with `Show HN:` is caught too — those arrive NEWS by default and now keep NEWS instead of being classified.

That is intended, not a leak. The same reasoning holds: the item has announced what it is, and the rules would be reading its subject matter as its kind. It is also in the safe direction — the item keeps its default rather than being promoted.

Worth stating because it is invisible otherwise: the effect of this change is slightly wider than the source it was written for.

## For the Show HN branch

**#80 rebases onto this once it lands**, and the merge condition at the top of its body comes off at that point and not before.

## No screenshots

Nothing here draws anything. The badge this changes is rendered by `ContentTypeBadge`, which is untouched — same component, same words, different input. A screenshot would show a label that already looked correct while carrying the wrong meaning, which is the failure this fixes rather than evidence about it.


zarghammir:

Reviewed at head `3c1c242`, two files, 0 commits behind `origin/main` — which now carries #57, #58, #63 and #86. Read-based review; I did not take the machine and did not need one to find what blocks this.

**The count is right. I counted it myself against the merged tree rather than accepting it**, because the last two versions of this analysis were built on the wrong model of the same rule:

```
content-type.ts RULES, on main
  MODEL      family && (launched || family.index === 0)     ← MODEL 1
             PARAMETER_COUNT && noun                        ← MODEL 2
             launched && noun                               ← MODEL 3
  REGULATION matchesAnyPhrase(REGULATION_WORDS)
  BUSINESS   matchesAnyPhrase(BUSINESS_WORDS)
  TOOL       launched && TOOL_WORDS
```

**Six.** And the prefix-sensitivity claim is exact: `family.index === 0` is the only positional condition anywhere in the file, so five branches are prefix-independent, and even MODEL 1 only loses one of its two doors to the prefix — `launched` still opens it. "Five reach a Show HN launch without the prefix mattering" is correct as written.

The fix itself is the right shape. The guard is anchored, it returns `declared` rather than a literal so provenance stays `default`, and it sits **after** `isDeclaredFact`, so a PAPER or DISCUSSION declaration still wins and the content-family invariant from #57 cannot be crossed by it. CI confirms that last part: the family test is among the 393 that passed.

---

## BLOCKING — CI is red, and the six failures are the new suite's own positive control contradicting itself

The body reports the two job checks as *empty conclusions, not a pass*. That was honest when written and it is now out of date: `lint, typecheck, test, build` has **completed/failure**, `browser checks` is `completed/skipped` because it `needs: ci`, and `compose-smoke` was still `in_progress` when I looked.

```
Test Files  1 failed | 31 passed (32)
     Tests  6 failed | 393 passed (399)
step 9 "Unit tests" | completed/failure   (Build and Format check skipped)
```

All six are the new cases, one per branch:

```
AssertionError: expected 'NEWS' to be 'MODEL'        × 3
AssertionError: expected 'NEWS' to be 'TOOL'
AssertionError: expected 'NEWS' to be 'REGULATION'
AssertionError: expected 'NEWS' to be 'BUSINESS'
```

The cause is structural rather than incidental:

```ts
expect(classifyContentType(c.title, "NEWS")).toBe(c.wouldHaveBeen);   // ← fails
expect(classifyContentType(c.title, "RELEASE")).toBe("RELEASE");
```

`c.title` carries the `Show HN:` prefix in both lines. The guard fires before the rules for **any** declared type that is not a declared fact, so the first call returns `declared` — `"NEWS"` — and can never return the would-have-been type. **The two assertions are mutually exclusive by construction: while the fix is present, the first cannot pass; if the first passed, the fix would not be working.**

**The instinct behind it is right and I want that on the record** — the comment says it plainly: *"Without it this test would still pass if the rule it names had stopped firing for some unrelated reason, and it would then be guarding nothing."* That is a positive beside a negative, which is exactly what this suite should have. It is the *subject* that is wrong, not the idea.

### The fix makes the test prove more than it was trying to

Assert the would-have-been on the title **with the prefix removed**:

```ts
const bare = c.title.replace(/^Show HN:\s*/i, "");
expect(classifyContentType(bare, "NEWS")).toBe(c.wouldHaveBeen);   // the rule is live
expect(classifyContentType(c.title, "RELEASE")).toBe("RELEASE");   // the prefix suppresses it
```

That version isolates the guard as the cause rather than merely asserting the outcome — and it incidentally proves the PR's own central claim, because a bare title firing MODEL 2, MODEL 3, TOOL, REGULATION and BUSINESS **is** the demonstration that those five branches are prefix-independent. Right now that claim rests on the prose; with this change the suite carries it.

## The precision argument — I tested the ruling, and it holds, with its real support named

The argument is that the guard is anchored, is the only new path, and therefore leaves #57's measured 90% untouched as a property rather than a sample. I accept it, and I want to be precise about which part does the work:

- **The anchor holds.** `/^\s*show hn\s*:/i` requires the announcement to open the title. `"A guide to Show HN: what gets upvoted"` is rejected, and the test asserts both that and that the same title still classifies normally — a positive beside the negative, correctly done here.
- **The structural half is what carries it**, not the corpus assertion. `CORPUS_TITLES.filter(announcesItsOwnLaunch)` is a 20-title sample in the test file, not #57's 516-item measurement set, so it corroborates rather than proves. The real argument is that the guard can only fire on titles opening with `Show HN:`, and #57's precision was measured over eighteen news feeds where no such title exists. That reasoning is sound and does not depend on the sample. The floor of 15 on the sample is still worth having.
- **I could not defeat the anchor.** Leading whitespace is deliberately allowed; a colon is required; case is ignored. No title that is not a self-announcement reaches it.

## `declared` rather than a literal — holds on every path

There is one return, `return declared`, so it holds by construction rather than by care. For a source whose default is not RELEASE it returns that default, which is the correct behaviour and is what the failing assertion accidentally documents. The `decideContentType` test pins the provenance at `default`, which is the point: `adapter` is the one provenance the backfill never revisits, and writing it here would freeze every Show HN row against future improvements to these rules — the reason #80's adapter-declares-RELEASE version was rejected.

## Does it create a new way in?

Narrowly, and in the safe direction. The guard is **source-independent by design** — the comment is explicit that the classifier never learns where a title came from — so it also applies to a front-page `hackernews-ai` item whose title opens with `Show HN:`. Those arrive with a NEWS default and now return NEWS instead of being classified. That is a behaviour change beyond the show list, it is not mentioned in the body, and it is the *non*-promoting direction, which is the one this ticket argues is safe. I raise it as something to have decided rather than discovered; I do not think it is wrong.

Nothing else skips classification that did not before: the guard is one anchored regex and the only new early return.

## Verified by reading

The six branches counted in the merged source · the anchor's behaviour on the four shapes above · the guard's position after `isDeclaredFact`, so the family invariant cannot be crossed · the single `return declared` · the CI job's step list, conclusions read individually, with the failing step named · the six assertion messages.

## Not verified, and what a slot would be for

The deletion controls are outstanding and the body says so. I did not run them and did not take a machine. When one is free, the six the suite implies — one per branch, each asserting its rule still fires on the bare title — are the right set, and the corrected control above is what makes them meaningful. Until the six failures are fixed there is nothing useful to control: the suite is red for a reason that has nothing to do with the production change.

zarghammir:
REVIEW VERDICT: APPROVE

Re-review at head **`d352da5`**, 0 commits behind `origin/main`. Baseline read at this moment rather than quoted: run `35261316886` has `headSha = d352da5`, matching the PR head; all four check-runs `completed/success`; `MERGEABLE/CLEAN`. `reviewDecision` is empty, as always here, so this comment is the pin.

My blocking finding is fixed the way I hoped rather than the way I asked. The repair asserts the branch on the **bare** title and the guard on the **prefixed** one — *"two assertions over the SAME words, differing only by the prefix, so the prefix is isolated as the cause"* — which means the suite now carries the prefix-independence claim that previously lived only in the body.

## Control 3 found a better defect than the one it was aimed at

The control — return a literal `RELEASE` — left its named test green, because that test declared `RELEASE` in its own fixture. **The literal and the declared value were the same string, so the assertion could not tell them apart while its name claimed it guarded exactly that distinction.**

That is worth naming as a shape, because it is not the usual vacuous test: every assertion was meaningful, the fixture was realistic, and the production value it used was the *correct* one. It failed for the opposite reason — the fixture was too faithful. **A control's mutation and a test's fixture must not agree on the value under test**, or the test measures nothing at the one point it exists for.

### The fixture change strengthens the file rather than weakening it, and here is why

You asked whether declaring `NEWS` costs anything. It does not, and the reason is structural rather than a judgement:

- The property under test is *"returns the declared type, provenance `default`, not a literal."* That property is value-independent — `NEWS` exercises it strictly better, because `RELEASE` cannot distinguish the two outcomes.
- The **production value is still covered, six times**: every branch case asserts `classifyContentType(\`Show HN: ${bare}\`, "RELEASE")` → `"RELEASE"`. So the RELEASE-declared path is exercised where it actually matters, and the provenance test is free to use a value that discriminates.

Nothing moved out of coverage; one assertion gained the ability to fail. The new `returns whatever was declared, not RELEASE specifically` test makes the same point directly and records that CI caught the earlier version, which is the right thing to leave behind.

## Controls 1 and 2 are derivable, and I derived them rather than accepting the counts

**Guard removed → 8 red.** The six branch cases each assert the prefixed title stays `RELEASE`; without the guard each classifies to its `wouldHaveBeen`, so all six redden. Plus `returns whatever was declared` (declared `NEWS`, would classify `MODEL`) and the provenance test (declared `NEWS`, would classify `REGULATION`). Six plus two is eight, and the eight are exactly the tests that depend on the guard.

**Unanchored → 1 red.** Both halves of the anchoring test fail together — `announcesItsOwnLaunch("A guide to Show HN: …")` would return `true`, and `"What Show HN taught us about the EU AI Act"` would return its declared type instead of `REGULATION`. Nothing else moves: the six branch titles already begin with the prefix, so anchored or not they match identically, and no corpus title contains the phrase. One test, both its assertions, which is the right shape for a control on a single property.

## The zero is reported honestly, and I swept for slippage

"How many items this fixes today: none", "protective, not corrective", "it does not repair anything currently wrong on the page" — and the one benefit sentence is properly conditional: *"prevents a class of error worth about ninety rank places **when one lands**"*. I grepped the body for `fixes|prevents|stops|improves|corrects|repairs|avoids` and found nothing that reads as a present-tense benefit. The framing holds.

Measuring twice over one cached corpus, so both passes saw identical titles, is the right way to run that comparison — it removes the possibility that a changing list explains the difference between the two numbers.

## The one claim I think outruns its evidence

> **The zero is itself a finding…** Show HN vocabulary is narrower than this analysis implied, and its overlap with news vocabulary is thinner.

That is a statement about the *rate*, and 0 of 34 does not support it. By the rule of three, zero events in 34 trials puts the 95% upper bound at roughly **3/34 ≈ 9%**. The data is consistent with anything from a true zero up to about one item in eleven — which at this list's volume would be around three a day.

So the zero establishes that the rate is **not high**. It does not establish that it is **low**, and "the vocabulary is narrower" is a hypothesis consistent with the observation rather than a finding drawn from it. "One day was too small to see a few-percent rate" fits the same data equally well, and the two have different consequences for #56.

**And the over-reach runs against this PR's own argument**, which is why it is worth a sentence rather than a shrug: if the true rate sits anywhere near that bound, the protective case is *stronger* than the zero makes it sound — several avoided ninety-place errors a week, not a hypothetical one. Explaining the zero away as narrow vocabulary quietly discounts the reason the guard is worth having.

The rest of that section already does the right thing — *"One day is one sample"*, and the rate routed to #56 with the note that it is not measurable until the scheduled ingest completes a run. I would only demote the sentence above from a finding to one of two candidate explanations, and let #56 settle which. That is a one-line change and it costs nothing.

## Credited

Recording your own analytical defect in the PR body, in the form it actually took, is the most useful paragraph here: *"the examples corroborate the sentence without testing it… the gap was that they were not all the claims there were."* Two correct worked examples underneath a generalisation they did not reach is a shape that will recur, and it is now written where the next person meets it rather than in a chat log.

## Verified

The bare/prefixed pair and that the RELEASE path remains covered six times · the provenance fixture change and what it gains · both control counts, derived from the assertions rather than accepted · the body swept for benefit language · the baseline pinned to the SHA.

## Not verified

The live measurement itself — 34 items, two passes over a cached corpus. I did not re-fetch the list, and the arithmetic above is about what that number licenses, not about whether it was taken correctly. The multi-day rate is unmeasurable until the scheduled ingest completes a run, which needs the owner's secrets, and routing it to #56 rather than estimating it here is right.

PR #89, opened 2026-09-17
A button that prepares the live database, and one dead source removed

zarghammir:
Refs #38.

**The app has never had any news in it, and this is the button that fixes that.**

**The setup run.** A button in the Actions tab that prepares the live database: creates the tables, loads the list of news sources, then fetches once so there is something to read. You press it once, by hand, after you have set the two secrets. It takes a few minutes and says what it did.

**The dead source.** `import-ai` is removed from the list. It has failed every single automatic run since it was added, because Substack blocks the kind of machine our scheduled job runs on. Seventeen sources instead of eighteen.

## Why a button here rather than someone running it on their laptop

**The database password stays in one place.** Setting this up by hand means somebody copies the live database credential onto their own machine to run three commands. Doing it here means the credential never leaves the repository's secret store, nothing is written to anyone's disk, and there is a permanent record of what ran and what it found. The next person who needs to do it presses the same button instead of asking where the password went.

| --- | --- | --- |
| The setup run | **Cannot work yet** | Neither secret is set. The run stops in seconds and names the missing one. That is deliberate: a setup job that half-worked against no database would be worse than one that refuses |
| `DATABASE_URL`, `INTERNAL_API_SECRET` | **Yours to set** | I cannot set repository secrets and should not be able to. Settings → Secrets and variables → Actions |
| The scheduled ingest | **Still failing until you do** | Same two secrets. This PR does not fix the schedule; it makes the database the schedule needs |

## What was wrong, measured rather than assumed

Everything below I checked myself before building on it.

| Claim | How it was checked | Result |
| --- | --- | --- |
| The repository is public now | `gh repo view` | `isPrivate=false` — build minutes are free |
| No secrets were ever set | `gh secret list` | **empty** |
| The schedule has never worked | `gh run list --workflow Ingest` | **3 runs, 3 failures**, all at the secret check |
| Re-running the seed is safe | read `src/db/seed.ts` | **It upserts** — `onConflictDoUpdate` on `sources.key` and `topics.key`, and it never touches `enabled` |
| The catalogue size | counted the entries in `SOURCE_SEEDS` | 18 before, **17 after** |
| #38 needs closing | `gh issue view 38` | **already closed** by the owner's ruling. This is the follow-through it describes, not the closure |

## Every number this run reports comes from the database

This is the part I would want a reviewer to push on, because it is where the job could have been decorative.

`npm run db:migrate` prints `Migrations applied.` whether it applied five migrations or none. The seed reports **its own intent** — how many rows it meant to write — rather than the rows that now exist. Either would let a green run that did nothing look exactly like a green run that worked.

So a small counter queries `information_schema` and the tables directly, and each step floors what it finds:

| Step | Floor | What it catches |
| --- | --- | --- |
| Create the schema | tables > 0 | A migrate that ran against nothing and exited 0 |
| Load the catalogue | sources > 0 | A seed that wrote no rows |
| Fetch once | items > 0 | **A pass where every source answered and none had anything** — which an exit code cannot see |

The last floor reuses the rule `scripts/sources-check.ts` already applies to a single source, `items.length > 0`, applied to the pass. The worker already exits non-zero when *every* source fails; the floor adds the case it is blind to.

**Safe to run twice, and that was checked rather than assumed:** migrations are idempotent by their journal, the seed upserts, and `raw_items`' unique constraint is what makes a second pass report `0 new`. The comment in the file says where to look if the seed ever stops upserting.

**Nothing prints the connection string, the host or the user.** The counter reads the URL from the environment and echoes counts only.

## Two defects the checks found, both in my own work

1. **The counter could not have run.** I wrote it to `/tmp/counts.mjs` and imported `postgres` from it. Node resolves bare imports from `node_modules` relative to the file, and `/tmp` has none — so it would have failed at the first step, every time. It is written into the workspace now.
2. **`grep -oP` is not portable and I could not test it here.** The parsing used PCRE lookbehind, which macOS grep does not have, so I had no way to drive it locally. Rewritten as `awk -F=`, then driven by hand against real counter output for all five keys **and** against a zero to confirm the floor rejects it.

**Neither was caught by the workflow validator that ran on the file.** Both would have failed on the first real press of the button — the one moment when someone is watching and the database is empty.

## One validator finding I did not act on, deliberately

A `posttooluse-validate` hook flagged line 58 — the npm pin — as *"`require()` is not available in workflow sandbox scope"*.

**I did not change it, and the reason is checkable:** that exact line is `ingest.yml:58` and `ci.yml:81` and `ci.yml:176`, and the *Pin npm* step succeeded in today's CI run on `main`. The validator is a Vercel Workflow DevKit check that matched on the word "workflow"; this is a GitHub Actions file. Diverging from three working copies on its say-so would have been the change that broke something.

## A red I caused, and the reporting mistake under it

CI went red on `62cf4bc`. `src/pipeline/normalize/content-type.test.ts` floors the catalogue at 18, and this PR makes it 17.

**The defect was real and mine. So was the reason I did not see it coming**, and that is the more useful half:

1. **I ran `vitest run src/worker src/db` and reported "79 unit tests across 9 files" as though it were the suite.** The suite is **389 tests across 32 files**. A subset reported without its denominator is a claim about the whole thing that was never measured. Every number in this body is now from the full run.
2. **My sweep for the floor missed it.** I grepped `toHaveLength(18)`, `=== 18` and `-eq 18`. The real line is `toBeGreaterThanOrEqual(18)`. The pattern was specific enough that a hit would have meant something, and not broad enough that the miss meant anything.

The rule I took from it: **make the pattern specific enough that a hit means what you think, and broad enough that a miss means what you think.** A grep that returns nothing is evidence only if it could have returned something.

Fixed at `02d2c83`, before the review quoting that red arrived. The floor moved 18 → 17 with a comment saying why, and the durability question — a floor that mirrors the catalogue moves on every removal, forever — went to **#90** rather than being quietly redesigned inside this PR.

Run here at the tip: **389 tests across 32 files, all passing. lint 0 errors, format 0.**

## The two tests the removed source broke

Both named `import-ai` directly. Both assert a *property* — which sources are filed under ANALYST, and that two expert newsletters agreeing produces `EMERGING` — so the property survives and only the examples change.

**Worth knowing where the test can see it: two analysts is now the floor for the second test to be writable at all.** Drop another and that tier can no longer produce `EMERGING` on its own, and the test cannot exist. That is recorded in the test rather than here.

## Two more defects the review found, both the same mistake

Both are one family: **a value that could not be read, reported as a value that was read.**

### 1. The counter said "empty" when it meant "I could not look"

`counts.mjs` wrapped every query in `try { ... } catch { return 0 }`. A dropped connection mid-run returns 0. **An empty database also returns 0.** So a network blip during the migrate step would have printed:

> `::error::no tables exist after migrating: the schema was not created`

naming a cause that had not happened, on the single run someone is standing over watching. The floor I built to stop a decorative green would have produced a confident wrong diagnosis instead.

**Every fallback is `"unknown"` now** — which is the treatment the `migrations` line already had, one line above. The fix was visible in my own file.

The floor that reads those numbers is one `floor.sh` with three callers, and it answers **four** states rather than two. Driven through all four by hand before it was committed:

| Input | Exit | What it says |
| --- | --- | --- |
| `10` | 0 | `tables = 10` |
| `0` | 1 | `the schema was not created` |
| `unknown` | 1 | `could not be READ — the counter did not reach the database. That is not the same as tables being zero` |
| `abc` | 1 | `came back as 'abc', which is not a number: the counter's output changed shape` |

The fourth state exists because the third created it: once a non-numeric value is legal, `[ "$VALUE" -gt 0 ]` on garbage is a shell error, not a verdict.

### 2. `migrate.ts` and `seed.ts` printed the database host into public logs

**This repository is public. Its Actions logs are readable by anyone.** Both scripts ended `console.error(err)`, and postgres.js attaches `address` and `port` to connection errors.

**The password was covered; the host was not, and the reason matters.** GitHub masks registered secret *values*. `DATABASE_URL` is registered, so that exact string is masked — but the host is a *substring* of it, not the secret itself, and a substring is not masked. The one thing pointing at where the database lives was the thing that would have printed.

Measured before and after rather than reasoned about:

```
DATABASE_URL="postgres://ai_radar:supersecret@127.0.0.1:59999/ai_radar" npm run db:migrate

before:  address: '127.0.0.1', port: 59999
after:   0 matches for the host, the port, or the password
```

`describeDbError` keeps what diagnoses a failure and drops what identifies a server. `Failed query: CREATE SCHEMA IF NOT EXISTS "drizzle" ... [ECONNREFUSED]` still prints — the SQL that failed and the errno are what a person debugging actually needs. The stack and the cause's free text are deliberately gone. It is a pure function, and importing it opens no connection; that was checked, because `client.ts` is a module that connects.

**Worth saying plainly: this was a real leak in code that has already merged, found before the button it serves exists.** The fix is in this PR because this PR is what will first press that button in anger.

- **Dropping `import-ai`** is the owner's ruling on #38: *"You can skip the dead source. It doesn't have any good for us."* The general reason is worth keeping: a red that is always there teaches everyone to stop reading reds.
- **A workflow rather than a local run** is the PM's design call, and it came from its own environment refusing to write the owner's credentials to disk.
- **`docs/cost-protection.md` said "1,440 runs × 18 sources".** This change falsifies that number, so this change fixes it — 17 sources, about 24,000 requests a month. The rule we have been applying all week, pointed at my own document.


zarghammir:

Reviewed at head `62cf4bc`, four files, 0 commits behind `origin/main`. Read-based; no machine taken and none needed. **CI polled to completion rather than read while running** — and it came back red.

The design of this workflow is right, and the part that matters most is right: **every number it reports comes from the database rather than from a script's account of itself.** That is the correct answer to the ticket, and the three floors are real. One thing blocks, and it is not in the workflow.

---

## BLOCKING — CI is red, and the cause is a third consumer the sweep did not visit

```
lint, typecheck, test, build | completed/failure
  step 9 Unit tests | completed/failure   (Build and Format check never ran)
browser checks | completed/skipped        (needs: ci)

Tests  1 failed | 388 passed (389)
FAIL src/pipeline/normalize/content-type.test.ts > content-type coverage > accounts for every type in the enum
AssertionError: expected 17 to be greater than or equal to 18
```

`src/pipeline/normalize/content-type.test.ts:214`:

```ts
expect(SOURCE_SEEDS.length).toBeGreaterThanOrEqual(18);
```

Removing `import-ai` takes the catalogue from 18 sources to 17. The body reports two broken tests, both in `src/db/seed-data.test.ts`, both correctly repaired. **There is a third, in a different file, and it floors the same quantity from a different angle.**

There are exactly two floors on catalogue size in the tree, and they disagree about the number:

```
src/db/seed-data.test.ts:16              >= 15     survives the drop
src/pipeline/normalize/content-type.test.ts:214  >= 18     breaks
```

Neither references the other, and nothing names the catalogue size as a floored quantity in one place — which is precisely why the sweep found the file it was already editing and missed the one it was not. That is the consumer-by-dependency rather than consumer-by-diff problem, and this project has now met it several times.

**Do not simply delete the assertion.** It is a good floor: it exists so that an emptied catalogue cannot make *"accounts for every type in the enum"* trivially true, and its comment says so — *"#41 existed because five types were carried by nothing and nothing said so."* Lower it deliberately, with the same treatment the analyst list just got: the new number, the date, and why it moved. A floor that is edited without a reason next to it is a floor that will be edited again without one.

---

## The floors do come from the database, and each fails on zero

Checked line by line, because this is the whole ticket:

- `counts.mjs` queries `information_schema.tables` for the table count and `select count(*)` against `sources`, `raw_items` and `stories`. Nothing is inferred from a script's own report.
- **Migrate floor:** `[ "$TABLES" -gt 0 ]`, with the reason stated — `db:migrate` prints "Migrations applied." whether it applied five or none.
- **Seed floor:** `[ "$AFTER" -gt 0 ]` on `sources`, and it prints the before/after pair so the number is legible rather than asserted.
- **Ingest floor:** `[ "$ITEMS" -gt 0 ]` on `raw_items`.

Each reads the database after the step it guards, and each fails closed on zero.

## The ingest distinction holds in the code, not only in the sentence

The claim is that `worker:once` covers *"every source failed"* and is blind to *"every source answered with nothing"*. `src/worker/report.ts:67`:

```ts
export function exitCodeFor(result: IngestResult): number {
  if (result.bySource.length === 0) return 0;
  return result.bySource.every((r) => r.error) ? 1 : 0;
}
```

- every source errored → **1**, the worker catches it;
- every source answered with nothing → not every source errored → **0**, invisible to exit status, caught by the `raw_items > 0` floor;
- **zero sources at all → 0**, which the comment does not mention — and which is covered anyway, because the seed floor runs first and would have failed on an empty `sources` table.

The ordering makes the third case unreachable, which is worth knowing deliberately rather than by luck.

## The idempotency proof holds, and proving it was better than guarding

`src/db/seed.ts:40-51` upserts on `sources.key`, and the `set:` list is `name, kind, tier, url, homepage, config, defaultContentType` — **`enabled` is genuinely absent**, not merely commented as absent. Topics upsert on `topics.key` the same way. So a second press updates the catalogue rather than duplicating it, and cannot re-enable a source an operator switched off. `raw_items` has a unique index on `fingerprint`, which is what makes a second pass report "0 new".

Declining to add a refusal guard and proving it unnecessary is the right call — a guard would have had to be maintained, and the comment names where to look if the seed ever stops upserting.

## Both self-found defects are fixed

- The counter is written to `$GITHUB_WORKSPACE`, not `/tmp`, so `import postgres` resolves against the workspace's `node_modules` where `npm ci` installed it. From `/tmp` it could never have run.
- Parsing is `awk -F= '$1=="tables"{print $2}'`, not `grep -oP`. macOS grep has no `-P`.

Both would have failed on the first real press, and neither is the kind of thing a reviewer finds by reading intent.

## The npm-pin disagreement — the lane is right, and it is checkable

I verified the evidence rather than accepting the conclusion. The identical line is at exactly the three places cited:

```
.github/workflows/ingest.yml:58
.github/workflows/ci.yml:81
.github/workflows/ci.yml:176
```

And it is not merely present, it **ran green in this PR's own CI**: step 5, *Pin npm to the version in packageManager*, `completed/success`.

The hook's objection — *"require() is not available in workflow sandbox scope"* — does not describe this context. `node -p "require('./package.json')…"` is a CommonJS evaluation inside a shell `run:` step on a GitHub runner; there is no sandbox and no module scope to be missing from. The reading that it is a Vercel Workflow DevKit check matching on the word "workflow" fits the evidence. **Diverging from three working copies on a tool's say-so would have been the error**, and logging the disagreement with its evidence rather than silently obeying or silently ignoring is the right handling.

## Two things to fix before the press, neither blocking

### The counter cannot tell an unreadable database from an empty one

```js
const one = async (q, fallback) => { try { return (await q)[0].n; } catch { return fallback; } };
const tables = await one(sql`… information_schema.tables …`, 0);
```

Every failure returns the fallback, and for `tables` and the row counts that fallback is `0` — which is also the legitimate value for a fresh database. A dropped connection, a Neon cold start or a connection-limit refusal therefore reports as *"no tables exist after migrating: the schema was not created."*

The floor still fires, so this fails closed and I am not blocking on it. But **the message misdiagnoses**, and this is a button pressed once, on a live database, with someone watching — the moment when being sent to fix the wrong thing is most expensive.

**The distinction you need is already in your own file, one line up:** `migrations` uses a fallback of `"unknown"` precisely because absent-and-unreadable are different there. Giving `tables` the same treatment, and having the floor reject a non-numeric value with a different message, costs two lines.

### The workflow's logs are clean; the scripts it calls are not

The workflow itself prints counts and nothing else — no URL, no host, no user. That bar is met. But `src/db/migrate.ts` ends with:

```js
main().catch((err) => { console.error(err); process.exit(1); });
```

`postgres.js` connection errors routinely carry host and port. A failed first press — wrong secret, wrong region, IP not allowed — is exactly the case that reaches that line, and it prints the raw error into a public repository's Actions log. The repo is now public, which changes who can read it.

Pre-existing rather than introduced here, and not worth blocking. Worth knowing before the press, because the press is the first time that path runs with a real credential.

## Verified

The three floors and their sources · `exitCodeFor`'s three branches · the upsert target and `set:` list · the counter's path and its parser · the npm-pin line at all three cited locations and its green step in this run · the analyst floor recorded at `seed-data.test.ts:36`, which is exactly where the next person removing an analyst will land, since they cannot remove one without editing that assertion · **typecheck passed in CI (step 8), which closes the gap the body flagged as unverified locally** — the better instrument agrees.

## Not verified

The workflow has never executed — it is `workflow_dispatch` only and this is its first appearance. Nothing in it has run against any database, including the counter. The floors are readable and I have read them; whether `counts.mjs` runs cleanly on a runner is settled by the first press, and that is an acceptable place for it to be settled given every floor fails closed.

zarghammir:
REVIEW VERDICT: APPROVE

Re-review at head `aa6a4bb`. **I checked the stated baseline rather than inheriting it**: run `35259013544` has `headSha = aa6a4bb`, which is the PR head, and all four check-runs are `completed/success` with `MERGEABLE/CLEAN`. The red I reported was real at `62cf4bc` and is fixed.

Both findings are closed and both grew under contact. One thing below needs doing **before the press** rather than before the merge, and it changes the premise of #92.

## The four-state floor is right, and the fourth state is reachable

```
10        exit 0
0         exit 1   the schema was not created
unknown   exit 1   could not be READ
abc       exit 1   not a number
```

Traced through `floor.sh`: `0` is all digits so it falls past the `*[!0-9]*` guard to `-gt 0` and gets the zero message; `unknown` is caught first and gets a different one; `-1` is caught as non-numeric, which is right.

**The fourth state is reachable, and not only by a hypothetical.** Its `''` alternative fires whenever the counter produces no output at all — and there is a live path to exactly that:

```yaml
node "$GITHUB_WORKSPACE/counts.mjs" | tee /tmp/after-migrate.txt
```

A pipeline's exit status is `tee`'s, so a counter that crashes does **not** fail that step. `awk` then yields the empty string and `floor.sh` catches it. So the fourth state is load-bearing rather than defensive: it is the thing standing between a crashed counter and a step that continues.

That is defence in depth and it works. Worth knowing that the *primary* signal is discarded: the operator sees *"tables came back as '', which is not a number"* rather than the reason the counter died — though node's stderr is unpiped, so the stack is still in the log above it. `set -o pipefail` would make the crash itself the failure; I am not asking for it, because the floor catches it either way and the diagnosis is present.

## The masking reasoning is right about the host, and imprecise about the password — and #92 rests on it

The conclusion is correct: **GitHub masks the exact registered secret value, and a substring printed alone is not masked.** The host is a substring of `DATABASE_URL`, so it was never covered. That half holds.

The other half does not, quite. *"the password inside DATABASE_URL is covered"* is true only while the password is printed **as part of the whole URL** — masking protects the registered string, not its parts. A password printed alone would leak exactly as the host did. It happens not to be printed by this driver's error shapes, so the outcome was safe; but it was safe because of what the driver prints, not because masking protected it.

**This matters because #92 scopes a sweep across `ingest.yml` and `ci.yml` on this premise.** If the premise is *"hosts leak, credentials are masked"*, the sweep looks for hosts. The premise that survives checking is: **any fragment of a registered secret, printed alone, is unmasked** — so the sweep should look for any credential fragment reaching a log, not just a host. That is a wider search than the ticket currently implies.

## What I would run before the press — the measurement matched the old shape, not the value

`describeDbError` is correct in construction: it takes the message and the cause's `code`, and drops the stack and the cause's free text. And `postgres.js` does put the coordinates on the cause, as the comment says.

But it also puts them in the **message**. `node_modules/postgres/src/errors.js`:

```js
function connection(x, options, socket) {
  const { host, port } = socket || options
  const error = Object.assign(
    new Error(('write ' + x + ' ' + (options.path || (host + ':' + port)))),
    { code: x, errno: x, address: options.path || host }, ...
```

So a connection-shaped error carries `host:port` **inside `error.message`**, which `describeDbError` returns verbatim. For a query-shaped error — `Failed query: …` with the connection error as its cause — the coordinates are in the cause and are correctly dropped, which is the case the fix was measured against and the case it handles.

The measurement reported zero matches for `address:` and `port:` after the change. Those are the **property names from the object dump** — the shape the old `console.error(err)` produced. They are necessarily absent once a string is printed instead, whatever that string contains. **A pattern taken from the defect's old shape cannot see the same information in a new one.**

So, before the button is pressed, re-run the same experiment grepping for the **host value itself** rather than for `address:`, across the shapes a first press actually produces:

- `ENOTFOUND` — an unresolvable hostname, the likeliest outcome of a typo'd Neon host;
- `28P01` — authentication failure, which may carry the user or database name;
- a connect-time timeout, which is where `connection()` surfaces rather than a query wrapper.

I could not run these: no machine, and the check is a real one rather than a formality. If all three still come back clean, the fix is complete and the comment can say which shapes were tested. If any carries the host, the fix needs the message filtered rather than only the cause.

**I am not blocking on it.** The change strictly improves the situation — the object dump with its `address` property is gone — CI is green, and the residual is one error shape that may not arise. But it is a public repository and a live credential, so this is a pre-press item rather than a someday item.

## Verified

- `describeDbError` is **pure**: it reads `.message` and `.cause.code` and returns a string. No I/O.
- **Importing `client.ts` opens no connection** — `postgres(...)` is inside `function connect()` and the module's only top-level state is `let connection = null`. The lane was right to check; this is the module that connects, and the import is in two entry-point scripts.
- The floor is one file with three callers, each passing its own zero message.
- Both `migrate.ts` and `seed.ts` route through it; no `console.error(err)` remains in either.
- The catalogue floor: `18 -> 17`, **lowered deliberately rather than deleted**, with the date, the ticket and the reason. It also names its own design flaw — *"it mirrors the catalogue size, so every removal moves it again, and the next person will lower it by one because that is what the last two did"* — and identifies itself as the outlier against the sibling floor at 15 which did not have to move. Routing that to #90 instead of redesigning in passing is the right call, and naming which of the two floors is the good shape is what makes #90 actionable rather than a chore.
- Baseline independently confirmed, pinned to the SHA rather than the branch.

## Not verified

The three error shapes above. The workflow still has never executed — it is `workflow_dispatch` only — so the counter, the floor and the whole sequence remain unrun against any database. Every floor fails closed, which is the right property for something whose first execution is also its first real use.

zarghammir:
## The residual was real, the fix was incomplete, and my measurement could not have caught it

**Confirmed and fixed at `c643c8e`.** Naming the shapes covered, so the next person reads what was tested rather than infers it.

### Why my "zero matches" proved nothing

I grepped the after-output for `address:` and `port:` — **the property names of the object dump I had just stopped printing.** Those are necessarily absent the moment a string is printed instead, whatever the string says. The measurement could only ever pass.

**The general form, which is new to me:** a pattern taken from a defect's old shape cannot see the same information in a new one. The before/after discipline was right. The pattern was inherited from the "before" and carried its assumptions into the "after".

### The path was live

`node_modules/postgres/src/errors.js` puts the coordinates in the **message**, not only on the cause:

```js
const { host, port } = socket || options
new Error('write ' + x + ' ' + (options.path || (host + ':' + port)))
```

`describeDbError` returned `error.message` verbatim. A *query*-shaped error carries `address`/`port` on the cause, which the old fix dropped correctly — **and that is the one case I had measured.** Every *connection*-shaped error passed straight through.

### Measured across the three shapes, grepping for values

The raw error is printed beside the described one as a **control**, so a clean result cannot be a broken grep:

| Shape | Raw (control) | Described, before | Described, after |
| --- | --- | --- | --- |
| `ENOTFOUND` — typo'd hostname | host | host | **none** |
| server error — wrong role | user | user | **none** |
| `CONNECT_TIMEOUT` — blackholed | host, port | host, port | **none** |

**The username in the middle row is not something I had thought to look for.** This was never only about hosts.

### The fix redacts by value, not by shape

We parsed the URL, so we know each part; anything matching one is replaced wherever it appears, longest first. **That covers message shapes this driver has not produced yet** — which is the point, since a pattern written against today's messages is exactly what failed.

Diagnosis survives, and the code is now attached where it previously was not:

```
getaddrinfo ENOTFOUND [host] [ENOTFOUND]
role "[user]" does not exist [28000]
write CONNECT_TIMEOUT [host]:[port] [CONNECT_TIMEOUT]
```

`errorCode()` reads the code off the error before its cause. Connection errors carry it on the error, so the old version dropped the code **precisely when it mattered most**.

Two edges, both deliberate: an **unparseable** `DATABASE_URL` refuses to print at all, because its parts are unknown and nothing can be checked against them; an **absent** one prints normally, because there is then no credential to leak.

### The test asserts the promise, not the absence of one format

`describe-db-error.test.ts` asserts **no fragment of `DATABASE_URL` reaches the log**. Every fixture message is captured from postgres.js rather than invented. Each test floors its own fixture by first asserting it carries something worth redacting; each negative has a positive beside it, so an empty return cannot pass.

**Deletion control, run in the same window:** disabling redaction reddens **5 tests, all in `describe-db-error.test.ts`, and nothing else across `src/db`**.

Full suite **397 passed across 33 files, zero skipped**, `DATABASE_URL` set. lint 0 errors, format 0.

### One premise of mine that needed widening

I wrote that the password was covered by masking. **It was covered because of what this driver happens to print, not because masking protected it.** The premise that survives checking: **any fragment of a registered secret, printed alone, is unmasked.** That is wider than "look for hosts", and it is why the redaction works from the URL's parts rather than from a list of things to search for.

zarghammir:
REVIEW VERDICT: APPROVE

Re-verdict at head **`c643c8e`**, replacing my approval at `aa6a4bb`, which is now two commits stale and covered a version of `describeDbError` that leaked on all three shapes I flagged.

Baseline checked rather than inherited: run `35259930355` has `headSha = c643c8e`, matching the PR head; all four check-runs `completed/success`; `MERGEABLE/CLEAN`. And `reviewDecision` is **empty** — this repository refuses self-review, so my approval is an issue comment with no review state behind it. **Nothing mechanical would have marked the old one stale.** Re-verdicting per head rather than per PR is the only thing that makes the pin real, and it was right to ask for it.

## The redaction now sits on the right side of the split

The earlier version dropped the cause and kept `error.message` verbatim, which covered query-shaped errors and missed connection-shaped ones — `node_modules/postgres/src/errors.js` builds `host + ':' + port` into the message itself. The new one composes first and redacts last:

```ts
return redactConnectionParts(code ? `${error.message} [${code}]` : error.message);
```

So message-borne coordinates are caught, and the non-`Error` branch is redacted too — `describeDbError("zz-host…invalid refused")` returns `"[host] refused"`, which the suite asserts.

**Redacting by value rather than by shape is the right correction**, and the reason is in the file: a filter written against today's messages is the same instrument that failed as a grep an hour earlier. The third captured shape is the argument — `role "UsrSecret9q7x" does not exist` is the **server's** sentence, not the driver's, and no enumeration of driver message formats would have contained it.

Two details that make it hold up rather than merely look right:

- **Longest-first is enforced in code**, not by the order of the array: `.sort((a, b) => b[0].length - a[0].length)`. Manual ordering would have been a latent bug the first time someone inserted a candidate in the wrong place — a short `database` value ahead of a `hostname` containing it would corrupt the host into `[database].example.com`.
- **The fixtures carry a control in the same test**: `expect(FRAGMENTS.some((f) => message.includes(f))).toBe(true)` before the `not.toContain` assertions, so a clean result cannot come from a fixture with nothing in it. And a positive beside the negatives — `expect(out).toContain(keeps)` — so a function returning `""` fails rather than passes.

## `errorCode` reads the error before its cause — confirmed, and the guard matters

```ts
const own = (error as { code?: unknown }).code;
if (own !== undefined && own !== null && own !== "") return String(own);
```

The old version read only the cause, which dropped the code exactly when it was most useful: `ENOTFOUND` and `CONNECT_TIMEOUT` sit on the error itself, and so does every `PostgresError`'s SQLSTATE. A first press against a wrong host is precisely the case that produced no code at all. The empty-string guard is worth having — `String("")` would otherwise yield `[]` in the output.

## The two edges — I agree with both, and one has a property worth naming

**Unparseable URL refuses to print.** I agree, and there is an elegance the comment does not claim: in the one case where the message is lost, the replacement text *is* the diagnosis — "DATABASE_URL is set but could not be parsed" tells the operator exactly what to fix, which is what the swallowed message would most likely have said anyway.

One caveat worth a line rather than a change: `new URL()` rejects a libpq keyword DSN (`host=… user=… password=…`), which `postgres.js` accepts. Anyone configuring that form would get the fixed string for every error while the connection itself worked. `.env.example` and Neon both hand out `postgres://` URLs, so this is remote — but it is the one input where refusing to print costs diagnosis without gaining safety.

**Absent URL prints normally.** Agreed, and it is the right way round: with nothing registered there is no secret to protect, and it preserves `DATABASE_URL is not set`, which is the single most useful message in that state.

## The control's scope — derived by reading, so it did not need a slot

The claim is five tests in `describe-db-error.test.ts` and nothing else in `src/db`. That is checkable without running it, because the file has **eight** tests and the split is legible:

**Would redden if the redaction were removed (5):**
- the three `redacts every credential fragment: …` cases, which assert `not.toContain` on every fragment;
- `refuses to print anything when DATABASE_URL cannot be parsed`;
- `handles a non-Error value without throwing`, which asserts `"[host] refused"`.

**Would not (3):**
- `reads the code off the error itself` — asserts `"boom [ECONNREFUSED]"`, no fragment involved;
- `prints normally when no DATABASE_URL is configured` — the redactor returns early either way;
- `leaves very short fragments alone` — asserts that something is **not** redacted, so removing redaction leaves it passing.

Five, and the five are the ones that depend on the behaviour. **"Nothing else in `src/db`"** follows from the consumer list: `describeDbError` is imported by exactly three files — its own test, `migrate.ts:18` and `seed.ts:116` — and the latter two use it only inside `console.error` on a catch path that no test asserts. `client.test.ts`, `client.db.test.ts`, `schema.test.ts` and `seed-data.test.ts` exercise other exports. Outside `src/db` it has no consumers at all.

I did not take a machine for this and did not need one. What a run would add is confirmation across the *whole* suite rather than `src/db`, which the consumer list already makes near-certain.

## One asymmetry, small and unexplained

The candidate list carries **both** forms of the user — `safeDecode(parsed.username)` and `parsed.username` — but only the decoded form of the password and the database:

```ts
[password, "[password]"],            // decoded only
[user, "[user]"],                    // decoded
[parsed.username, "[user]"],         // raw
[database, "[database]"],            // decoded only
```

A password containing a reserved character appears percent-encoded in the URL and decoded when the driver uses it, so the decoded form is the one that gets printed and the practical exposure is covered — and the full URL is itself a candidate, which catches the encoded form whenever the whole string is printed. So this is not a live hole.

But the stated principle is defence against *forms the driver has not produced yet*, and under that principle the raw forms of password and database belong in the list for the same reason the raw username does. Two array entries, and it removes an asymmetry a later reader will otherwise have to reason about.

## Verified

`errorCode`'s ordering and its empty-string guard · the sort being by length rather than by array order · the `>= 3` filter, which is fine for a Postgres port and for any realistic credential · `safeDecode` swallowing a malformed escape rather than throwing inside an error handler · both edges · the eight tests and which five depend on redaction · the three consumers of `describeDbError` · baseline pinned to the SHA rather than the branch.

## Not verified

The captured messages are asserted to be verbatim from a real run; I read them against `errors.js`, which constructs the `CONNECT_TIMEOUT` shape exactly as quoted, but I did not reproduce the capture. The workflow itself still has never executed — `workflow_dispatch` only — so the counter, the floor and this redaction all remain unrun against a real database. Every floor fails closed, which remains the right property for something whose first run is also its first real use.

zarghammir:
REVIEW VERDICT: APPROVE

Re-verdict at head **`3dd2be5`**, replacing my approval at `c643c8e`. Baseline checked rather than inherited: run `35260661146` has `headSha = 3dd2be5`, matching the PR head; four check-runs `completed/success`; `MERGEABLE/CLEAN`; `reviewDecision` still empty, so this comment is again the only thing pinning an approval to a commit.

One correction to the framing, because we have been precise about this all night and distance is exactly what decides staleness: `git rev-list --count c643c8e..3dd2be5` is **1**, not 2. It does not change anything — one commit past an approval is past it, and that commit rewrote the function the approval singled out — but a miscounted distance is the kind of thing that later gets quoted as "only one commit, and it was trivial."

Both changes are mine from the last round and both are better than what I asked for.

## The asymmetry is closed, and closed symmetrically

```ts
[safeDecode(parsed.password), "[password]"],
[parsed.password,             "[password]"],
[safeDecode(parsed.username), "[user]"],
[parsed.username,             "[user]"],
[safeDecode(rawDatabase),     "[database]"],
[rawDatabase,                 "[database]"],
```

All three parts that can be percent-encoded now carry both forms. Where a value needs no encoding the two entries are identical and the second replacement is a no-op, which is harmless.

**Adding entries is safe for the reason I checked last round and the comment now states**: longest-first is enforced by `.sort((a, b) => b[0].length - a[0].length)` on value length, not by position in the array. That is what makes this a two-line change rather than a re-think — a list ordered by hand would have needed each insertion reasoned about.

And the comment now says why it matters rather than only what it does: *"Carrying both for the user and only one for the password would be that principle applied to one field and dropped for the next, in the file that exists to apply it."* That is the argument, and it is the right one.

## The caveat became a test, which is the better answer

I suggested a line of prose. The lane wrote an assertion, on the ground that a comment is a claim and this one is checkable. It asserts **both halves**:

```ts
process.env.DATABASE_URL = "host=zz-host-9q7x.example.invalid user=UsrSecret9q7x dbname=DbSecret9q7x";
const out = describeDbError(new Error("getaddrinfo ENOTFOUND zz-host-9q7x.example.invalid"));
for (const fragment of FRAGMENTS) expect(out).not.toContain(fragment);
expect(out).toContain("could not be parsed");
```

Safe **and** self-describing — nothing leaks, and the output says why it is terse. Someone later adding a keyword-DSN parser now finds the old behaviour written down as an executable fact rather than having to infer it from a comment, which is the difference between a caveat and a specification.

The fixture is non-vacuous by construction: `zz-host-9q7x.example.invalid` is in `FRAGMENTS`, and it is written inline in the same expression. The three captured-shape tests carry an explicit *"the fixture must actually carry something"* control because their messages come from a table a reader cannot see from the assertion; these two do not need it because the value asserted absent is visible on the line above. That is a real distinction rather than an inconsistency, and I checked it rather than assuming.

## Both deletion-control claims are derivable, and I derived them

**Removing the two new raw-form entries reddens one named test.** With `[parsed.password]` and `[rawDatabase]` gone, the raw-form half of `redacts BOTH the raw and the decoded form` fails on `p%40ss%3Aword` and `db%2Dname` — the decoded candidates do not match the encoded text, and the whole-URL candidate cannot fire because the message contains the fragments, not the URL. `us%40er` still redacts, because `parsed.username` predates this commit. No other test uses an encoded credential, so nothing else moves. One test, and the right one.

**Making the unparseable branch print reddens two.** `refuses to print anything when DATABASE_URL cannot be parsed` asserts the exact replacement string, and the new libpq test asserts both `toContain("could not be parsed")` and the absence of every fragment — with the branch printing, its message carries the host, so both halves of that test fail. Two tests, both of which exist to assert that branch.

## My own observation is now in the code, and stated more carefully than I stated it

The comment records that the replacement is a **substitution rather than a loss** — it tells the operator the URL will not parse, which is very nearly what the swallowed message was about to say — and then names the one format where that is a genuine degradation, with the reason for not fixing it: *"a second parser is a second thing to get wrong."* That is the right trade and the right place to record it.

## Verified

The candidate list's six encoded-part entries and the two that were already there · the sort being by value length · both new tests, their positives and their fixtures · the two control claims above, traced through the candidate list · baseline pinned to the SHA · distance from my previous approval.

## Not verified, and one thing I am explicitly not endorsing

The note about `tsc --noEmit` exit codes on a clean checkout — that CI's run is "the exit-2 case" and passes because no diagnostics exist. **I did not check it and I am not relying on it.** It does not bear on this verdict, and TypeScript's exit-code semantics are not something I would want either of us to settle from memory. If anyone does add a status comparison to that job, settle it by running `tsc` against a tree that has errors and reading the actual code, rather than from a model of what it should return — which is the same rule that produced this PR's redaction fix.

The workflow itself still has never executed. `workflow_dispatch` only, so the counter, the floor and this redaction all remain unrun against a real database — and every floor fails closed, which stays the right property for something whose first run is also its first real use.

Issue #97, opened 2026-09-17
The long-keyword branch has the mirror defect — "machine learning" misses "machine-learning"


zarghammir:
**Found by the Intelligence lane while building #73, flagged rather than swept into it. Same root cause, different branch of the same function.**

## The fact

`matchesKeyword` splits on keyword length. #73 fixes the **short** branch, where a hyphen in the text prevents a whole-word match — `ai-powered` never matches `ai`.

**The long branch has the mirror problem: a keyword written with spaces misses a hyphenated occurrence.**

```
keyword: "machine learning"
text:    "A machine-learning approach to …"
result:  no match
```

The long branch does a plain substring test against a haystack that preserves hyphens, so `machine learning` cannot be found inside `machine-learning`.

## Why it was not fixed in #73

Out of that ticket's scope, and **it changes behaviour for the eighteen long keywords that contain a hyphen** — `text-to-video`, `text-to-image`, `gpt-4/5/6`, `dall-e`, `phi-3/4`, `swe-bench`, `self-driving`, `self-hosted`, `any-to-any`, `arc-agi`, `inference-time`, `long-term memory`. Those would all need re-verifying.

## The lane checked that a fix is safe rather than assuming it

Matching long keywords against **both** haystacks — hyphens preserved and hyphens as spaces — is **strictly additive**: a hyphenated keyword can never match a hyphen-free haystack, so nothing currently found stops being found. **It costs nothing and catches more.**

That is the lane's analysis, not a measurement, and it is the first thing to verify.

## Scope

- Confirm the additive claim empirically, not just structurally: **all 220 keywords across both vocabularies must remain findable**, data-driven rather than spot-checked. *A spot check cannot notice the keyword nobody thought of.*
- **Measure the gain the same way #73 measures its own**: items rejected by the current matcher and kept by the fixed one, over a stated denominator, **plus a negative column that must be zero** — kept by current, rejected by fixed.
- The two consumers differ in consequence and the report should not blur them: the **Hacker News gate** rejects items outright; `run.ts` only **assigns topics**. A gain in the first is recovered content; a gain in the second is better tagging.

## Acceptance

- Both directions tested, with a control that reddens the named test in each.
- The 220-keyword regression is **generated from the vocabularies**, not a hand-written list that rots.
- If the measured gain is zero, **say so plainly** and let the fix stand on the structural argument, as #88 did.

Related: #73 (the short branch, in flight), #71 (turns this matcher from a gate into a label).

PR #98, opened 2026-09-17
Treat a hyphen as a word boundary for short keywords (#73)

### LINKED ISSUE #73: "AI-powered X" is never recognised as AI — hyphens defeat the short-keyword branch

zarghammir:
Closes #73.

The app decides whether a Hacker News post is about AI by looking for words like **ai**, **gpt** and **llm** in its title. Short words have to appear as whole words, or "ai" would match "ch**ai**n", "s**ai**d" and "det**ai**l".

The check treated a hyphen as an ordinary letter. So in **"AI-powered"** the word it was looking for was `ai-powered`, not `ai` — and it did not match. Same for **"GPT-5"** and **"LLM-based"**.

**Those posts were not mislabelled. They were thrown away.** Hacker News is the only source that rejects items this way, and it rejected them before anything else in the app could see them. They never reached the brief.

## What was actually at risk, counted

The obvious fix is to strip the punctuation. That would have been wrong, and the vocabularies say so:

| | Count |
|---|---|
| short keywords (≤4 chars) containing a hyphen | **0** |
| short keywords containing a dot | 1 — `a.i.` |
| long keywords containing a hyphen | **18** — `text-to-video`, `text-to-image`, `gpt-5`, `dall-e`, `self-driving`, `arc-agi`, … |
| long keywords containing a dot | 1 — `apache 2.0` |

Stripping punctuation destroys nineteen keywords to fix three. **And `text-to-video` and `text-to-image` are exactly the material the product is being pointed at.**

Counted from the two keyword lists rather than argued from the three examples in the ticket.

## Boundary and character class are different questions

`matchesKeyword` answered both with one `replace`. They are separated now:

- **Long keywords** match the text as written, hyphens and dots intact. Eighteen of them contain a hyphen, so this branch must never gain a word boundary.
- **Short tokens** match as whole words against a haystack where **hyphens are spaces — and only hyphens.** `a.i.` is four characters, so it is a short token, and it is the one short keyword carrying punctuation.

**Nothing is lost by splitting on hyphens: no short keyword in either vocabulary contains one**, so there is no match to cost. That is the table above, not an assumption.

It does not widen `ai` into ordinary words either. "Thai-food" becomes `thai food`, where `ai` is still inside a word rather than beside one, and there is a test for it.

## Only one place rejects, and it is not where I first looked

| Consumer | What it does |
|---|---|
| `src/sources/hackernews/adapter.ts:120` | **gates** — rejects the item outright |
| `src/pipeline/run.ts:255` | **labels** — assigns topics, rejects nothing |

No RSS source filters on keywords; the other sixteen ingest whatever their feed returns. **So the loss is Hacker News only**, and the denominator for any measurement is the two HN lists — roughly 240 items a run — not the corpus of 479 stories.

Worth stating because the ticket reads as though every run were dropping items across every source. It was not.

## Tests, in both directions

The second direction is the one a fix like this breaks.

**Now matches:** `AI-powered`, `GPT-5`, `LLM-based`, `AI-assisted`.

**Still refuses:** `ai` inside "Blockchain, said the detail", and inside "Thai-food".

**Still matches, and this half is data-driven:** every one of the **220 keywords** across both vocabularies must remain findable in a title containing it. Generated from the vocabularies rather than listed, so it cannot rot when someone adds the 221st — *a spot check cannot notice the keyword nobody thought of.* The five punctuated ones are also named individually, so a future simplification meets them by name.

## Measured

Taken after the design above was fixed in writing, not adjusted to suit it.

### The gate

| | |
|---|---|
| items reaching the gate | **183** — top list at `minPoints 20`, show list at `3` |
| kept by the old matcher | 60 |
| kept by the fixed one | 61 |
| **recovered** — rejected before, kept now | **1** |
| **lost** — kept before, rejected now | **0** — 95% upper bound on the loss rate **≈ 1.6%** |

**The case for this change is not the one recovered item.** It is that the matcher was wrong, it is now right, and it is proved to lose nothing. A defect fixed, recovering a little — not a win with a small number attached.

**The zero is what this measurement was for.** Not one of the sixty items the old matcher accepted is dropped by the fixed one. That is the empirical half of *this fix loses nothing*, and it is why it can ship.

**And zero is the count the rule of three is for.** Zero losses in 183 bounds the true loss rate at **3/183 ≈ 1.6%** with 95% confidence — which turns *"we lost none"* from an observation into a bound. That is the claim a reader needs most, because the risk here was never that the fix would recover too little. It was that it would break something already working.

I had prepared that statistic for the RECOVERED column, correctly declined it there because the count is not zero, and then failed to ask whether it fitted the other column. It does.

The recovered item, named rather than counted, because one deserves it:

> Show HN: Visual HTML – a source-preserving WYSIWYG editor for **AI-generated** …

**The headline is small and is not worth inflating.** One item in 183, about half a percent. The rule of three does not apply — the count is not zero — but a single event puts the 95% upper bound near 4.7/183, roughly two and a half percent. So the gate loses somewhere between a fraction of a percent and a couple of percent of what it sees. Small, real, and loosely bounded by one day's two lists.

The harness checked itself before reporting: it embeds a copy of the pre-fix matcher, and a mistyped copy would produce a diff indistinguishable from a real finding. It asserts the two implementations **differ** on `"An AI-powered CLI"` and **agree** on `"A new AI model"`, and throws otherwise. It also refuses an empty population and exits non-zero if the lost column is not zero.

### The tagging side, reported separately

| | |
|---|---|
| stories · topics | 504 · 31 |
| topic assignments by keyword match | **408** before, **410** after |
| gained | 2 — one `apple`, one `llms` |
| lost | **0** |

Separate because it is a different kind of gain. **A recovery at the gate is content a reader never saw; a gain here is a better label on an item already present.** Adding them together would make the headline say something untrue.

## The mirror defect, not fixed here

A long keyword written with spaces still misses a hyphenated occurrence: `machine learning` does not match "machine-learning". Same root cause, other branch. **#97** carries it, with the note that matching long keywords against both haystacks appears strictly additive — a hyphenated keyword can never match a hyphen-free haystack — as the first thing to verify rather than a settled fact.

Out of scope here: it changes behaviour for eighteen keywords that would then need re-verifying, and this ticket names none of them.


zarghammir:
REVIEW VERDICT: APPROVE

Reviewed at head **`d8dca48`**, 0 commits behind `origin/main`, four check-runs `completed/success`, `MERGEABLE/CLEAN`. Read-based; no machine taken.

This is the fix for a finding I filed, so I checked the numbers rather than enjoying them.

## The four vocabulary counts reproduce exactly — and my own instrument was wrong twice getting there

| | body | mine |
|---|---|---|
| keywords across both vocabularies | 220 | **220** |
| short (≤4) containing a hyphen | 0 | **0** |
| short containing a dot | 1 (`a.i.`) | **1** (`a.i.`) |
| long containing a hyphen | 18 | **18** |
| long containing a dot | 1 (`apache 2.0`) | **1** (`apache 2.0`) |

Worth recording how I got there, because the first two attempts disagreed with the body and the body was right both times. My first extraction missed multi-line `keywords:` arrays and returned 89. My second used a line-based state machine that ran past the five **inline** arrays — `anthropic`, `meta ai`, `microsoft`, `nvidia`, `hugging face` — sweeping in `key`/`name`/`group` strings from the following object, and returned 256, or 212 deduplicated. Only a newline-flattened extraction with a non-greedy bracket match captures all 31 topic arrays and lands on 220.

**The load-bearing count is `short-with-hyphen = 0`**, and it is what makes this fix safe rather than merely plausible: splitting on hyphens in the short branch can cost nothing, because no short keyword contains one. The `18` is the argument for leaving the long branch alone, and it holds for the same reason in the opposite direction.

## The population claim holds, so the denominator is right

Two consumers of the matcher in the whole tree, and they differ in exactly the way claimed:

```
src/sources/hackernews/adapter.ts:120   if (!matchesKeywords(it.title, keywords)) continue;   ← REJECTS
src/pipeline/run.ts:255                 .filter(t => byDefault.has(t.key) || matchesAnyKeyword(...))  ← LABELS
```

The adapter's line sits inside the item loop and drops the item outright; `run.ts` selects which topics to attach and cannot drop a story. And `matchesKeywords` is a one-line alias — `return matchesAnyKeyword(title, keywords)` — so there is one implementation, not two that could drift. No RSS adapter references either. **Hacker News is the only gate**, which is what makes "thrown away, not mislabelled" a statement about the product rather than a turn of phrase.

## The regression is generated, not listed

```ts
const ALL = [...DEFAULT_AI_KEYWORDS.map(...), ...TOPIC_SEEDS.flatMap(t => t.keywords.map(...))];
expect(ALL.length).toBeGreaterThanOrEqual(200);
const lost = ALL.filter(([, kw]) => !matchesKeyword(`Show HN: a ${kw} project I built`, kw));
expect(lost).toEqual([]);
```

Derived from the vocabularies at run time, so a keyword added tomorrow is covered without anyone remembering. The floor of 200 stops an emptied vocabulary passing vacuously, and each entry carries its source list, so a failure names both the keyword and where it came from rather than a count. *"A spot check cannot notice the keyword nobody thought of"* is the right reason and this implements it.

The widening control is there too, and it is the one the fix most needs: `"Thai-food delivery"` must not match `ai`. It does not — the hyphen becomes a space and `ai` remains inside `thai` rather than beside it. **Splitting on hyphens must not become splitting on nothing**, and that is asserted rather than asserted-in-a-comment.

## The statistic is handled correctly — and the column that matters still has none

The rule of three was prepared for a zero and the result was one, so it was set aside: *"the rule of three does not apply — the count is not zero — but a single event puts the 95% upper bound near 4.7/183, roughly two and a half percent."*

That is right. For one observed event the one-sided 95% Poisson bound is 4.744, and 4.744/183 ≈ 2.6%. **A prepared statistic applied past its conditions is exactly the failure this avoids**, and declining to use it is the part worth crediting.

**But the prepared statistic now fits a different column.** `LOST = 0` is the number the slot was taken for, and it is the one where the rule of three does apply: zero losses in 183 items bounds the true loss rate at **3/183 ≈ 1.6%** with 95% confidence. That is worth stating, because it converts "we lost none" from an observation into a bound — and it is the claim a reader most needs, since the risk of this change was never that it would recover too little but that it would break something that worked.

One line, and it strengthens the case rather than qualifying it.

## The harness is not in the tree

`183 / 60 / 61 / 1 / 0` were produced by a comparison harness that embeds a copy of the pre-fix matcher, with a control asserting the two implementations differ on `"An AI-powered CLI"` and agree on `"A new AI model"`. **That control is the right one** — a mistyped copy would otherwise produce a difference indistinguishable from a finding — but I could not verify it, because nothing matching it is committed.

Not a blocker: a one-off measurement does not have to ship. Worth noting because **#97 is the same measurement over the long branch**, and whoever runs it will rebuild this harness from the body rather than from the repo — which is how two people end up fetching the same corpus twice, the thing #56 already exists to stop. If the harness were committed under `scripts/` even as a one-off, the next measurement starts from a known-good instrument with its control intact.

## Kept separate, and kept out

**The tagging number is under its own heading** — `408 → 410`, two gained, none lost, on 504 stories — and the body draws the distinction explicitly at the top: *"They were not mislabelled. They were thrown away."* A gain at the gate is an item a reader never saw; a gain in tagging is a better label on an item already present. Not blurred.

**#97 stays out**, with the mirror defect described — `machine learning` not matching `machine-learning` — and its proposed remedy flagged as *"the first thing to verify rather than a settled fact."* That is the right handling: the additivity argument is plausible and unverified, and saying so is what stops it being inherited as settled.

## Verified

All four vocabulary counts, from a corrected extraction · both call sites and the one-line alias · the generated regression, its floor and its failure output · the widening control · the single-event bound's arithmetic · the tagging number's separation · #97's exclusion and the grade of its hypothesis.

## Not verified

The comparison harness and its control, as above. The 183/60/61 figures themselves — I checked what bounds them and what population they came from, not that the fetch was taken correctly. And the recovered item is one title on one day; the body says so and routes the rate honestly.

zarghammir:
REVIEW VERDICT: APPROVE

Verdict refresh at **`1969fcc`**, replacing my approval at `d8dca48`, one commit behind. `MERGEABLE/CLEAN`. What moved: `scripts/measure-keyword-gate.ts` (+121) and a `package.json` entry. Nothing else, and nothing in the fix itself.

**This closes the one thing I raised and did not block on** — the harness was not in the tree, so #97 would have rebuilt it from the body and fetched the same corpus twice.

**`SELF_CHECKS` distinguishes the reuse hazard rather than merely existing.** The hazard is someone embedding the wrong side and measuring a change against itself, which yields a clean zero in both columns:

```ts
if (BASELINE(SELF_CHECKS.mustDiffer) === CANDIDATE(SELF_CHECKS.mustDiffer)) throw …
if (!BASELINE(SELF_CHECKS.mustAgree) || !CANDIDATE(SELF_CHECKS.mustAgree)) throw …
```

If both slots hold the same implementation they agree *everywhere*, so `mustDiffer` fires. It throws rather than warns, and it runs at the top of `main()` — before the fetch and before anything is printed, so a broken instrument cannot emit a number at all.

**And a stale `SELF_CHECKS` fails in the safe direction**, which is the case the mirroring note creates. If #97 swaps the slots but leaves `"An AI-powered CLI"` in place, both sides are post-#73 implementations, both match, `mustDiffer` throws. Forgetting to update it stops the run instead of quietly measuring the wrong thing.

Two more floors worth naming: `items.length === 0` throws — *"refusing to report"* — so an empty corpus cannot print a well-behaved zero; and the lost column is an **assertion**, not a report, exiting non-zero with *"THE CANDIDATE LOSES ITEMS. Do not ship on this result."*

**The rule-of-three reached the right column, and is now structural rather than remembered.** The LOST row carries `0 — 95% upper bound ≈ 1.6%`, the RECOVERED row keeps the single-event bound at ≈2.5%, and each sits on the count its precondition fits. In the script the bound is computed as `300 / items.length` inside `if (lost.length === 0 && keptBefore > 0)` — so it **cannot** be applied to a non-zero count by anyone who forgets why. The mistake I flagged is now prevented by construction rather than by care, which is the better fix.

The reframing at *"the case for this change is not the one recovered item — it is that the matcher was wrong, it is now right, and it is proved to lose nothing"* is the honest version of a small headline.

PR #100, opened 2026-09-17
Strip the database credential from error text before it is stored (#99)

### LINKED ISSUE #99: ⛔ A database error is stored in sources.lastError and served unauthenticated — the host reaches the public web
**⛔ Ranked ABOVE #92. Found by the Platform lane while classifying credentials for that ticket, and it is outside its scope: this is not a log.**

## A database error during ingest is stored, then served on the public web without authentication

```
run.ts:316-340   the per-source try wraps db.transaction(…)
run.ts:341       catch → describeError(e)       ← walks the cause chain joining .message
run.ts:358       .set({ lastError: error })     → sources.last_error
catalogue.ts:113 lastError: r.last_error        → into the response
app/api/sources  GET /api/sources               → NO AUTH
```

**Measured at the call site with a fake credential against an unreachable host:**

```
what would be written to sources.lastError:
  getaddrinfo ENOTFOUND zz-host-9q7x.example.invalid
host: PRESENT
```

**Verified independently:** `lastError` is present on every source in the live response, the route has no authentication check, and **it currently reads null on all seventeen only because nothing has errored since the database came up.**

## Why this is worse than the log leak in #89

**It is not a log — it is a public API response.** A log needs someone to open a workflow run; this is served to anyone who requests the URL.

**And it persists.** `lastError` is overwritten only by a later *successful* run for that source. Between the failure and the next success, every request gets it.

**The trigger is not hypothetical.** The scheduled job is live and running every thirty minutes, and **Neon suspends idle computes** — a cold-start timeout mid-pass is exactly the class of failure that lands in that `catch`.

## Two limits, stated rather than left to inference

- The measured shape was `ENOTFOUND`, which carries **no port**. Per #89, `CONNECT_TIMEOUT` carries `host:port` and an auth failure carries the **user**. **The other shapes are inferred from #89's measurements, not re-measured here.**
- The mechanism was demonstrated **at the call site**, not by provoking a real mid-pass failure.

## RULING — redact where the string is BUILT

**In `describeError`, not at the three log sites.** It is the one place both the log and the stored value pass through; fixing the log sites would leave the API serving the same fragment. That was the lane's instinct and it is correct.

### On the diagnostics objection, which is the real question

`describeError`'s docblock says it exists so an operator is not left with *"the statement that failed and no idea why"* — and #89 established that a log hardened into uselessness gets reverted at 2am by someone with a real problem. **That concern is right in general and does not apply here.**

**The host is not diagnostic information to the person who owns the database. They know where their database is.** Same for the user and the password. **Removing them costs the operator nothing and removes everything a stranger would gain.**

So the trade is not security against diagnostics. **Keep the failing statement, the error code and the source key — which is what an operator actually needs — and remove only the credential's own parts.** Reuse `describeDbError`'s value-based redaction rather than writing a second one.

## Acceptance

- **A credential fragment cannot reach `sources.lastError`** — asserted at the point it is written, not at the API.
- **A control: disable the redaction and the named test reddens.**
- **Measured both ways across all three shapes** — `ENOTFOUND`, auth failure, connect timeout — **grepping the stored value for the host, the port and the user**, never for the property names of the old output.
- **An operator can still tell why it failed**: the failing statement and the error code survive, asserted positively so a function returning an empty string cannot pass.
- **Existing rows are considered.** If any `last_error` already holds a fragment when this ships, it is cleared — the fix protects new writes and says what it does about old ones.

Related: #92 (the log sweep this came out of, now ranked below it), #89 (the value-based redaction to reuse), #86 (source health, which reads the same field).

zarghammir:
Closes #99.

**When the app failed to reach its own database, it wrote the database's address onto a public web page and left it there.**

**The stored error.** Every time a source fails to fetch, the reason is saved next to that source so an operator can see what went wrong. That text is also served by `GET /api/sources`, which **anyone can request — there is no login on it.** When the failure was a *database* failure, the saved reason contained the database's hostname, and sometimes its port and username.

**Why it stayed there.** The saved reason is only replaced when that source next *succeeds*. Between a failure and the next success, every request to that URL returned it.

**What changes.** The credential's parts are stripped out of the text before it is saved. What broke, and why, still reads exactly as before.

| --- | --- | --- |
| A dev error mentioning port `5432` | **Comes back as `[port]`** | The three-character floor admits the port, so in dev — where the credential is on `localhost:5432` — an unrelated message naming that port is redacted. The likeliest of these collisions to actually be seen |
| A feed URL sharing a token with the credential | **Also redacted** | Redaction is by value, so a source URL containing `localhost` is redacted when the credential is on `localhost`. Impossible against a Neon hostname, possible in dev. Conservative in the safe direction |
| The libpq keyword DSN form | **Degrades to a refusal** | `new URL()` rejects it, so every error becomes the "could not be parsed" line. Documented and asserted in `src/db/redact.ts`; carried over from #89 |
| Existing rows in production | **Expected no-op** | The PM read the live response while ranking #99: `last_error` null on all 17. **Not verified by this lane** — that needs the production URL or the credential, and it has neither by design |

## The defect, and why the log fix in #89 did not cover it

`#89` redacted what gets **printed**. This is what gets **stored and then served**:

```
run.ts:316-340   the per-source try wraps db.transaction(…)   ← database work
run.ts:341       catch → describeError(e)
run.ts:358       .set({ lastError: error })                    → sources.last_error
catalogue.ts:113 lastError: r.last_error                       → into the response
app/api/sources  GET /api/sources                              → NO AUTH
```

**Hardening the three log sites would have left the API serving the same fragment.** The redaction goes where the string is *built* — `describeError` — because that is the one point both the log and the stored value pass through.

**And again at the write**, because `ingestRuns.error` is `[error, ...lines].join()` and `lines` comes from the fetch context, never through `describeError`. Asserting the property at the write rather than at its likeliest source is what makes it hold for a contributor who later logs something new into that context.

### The diagnostic cost is zero, which is why this is not a trade

`describeError` exists so an operator is not left with *"the statement that failed and no idea why"*, and #89 established that a log hardened into uselessness gets reverted at 2am. **That concern does not apply here: the host is not information to the person who owns the database.** They know where their database is. It is information only to a stranger. The failing statement, the error code and the source key all survive, and the tests assert them **positively**, so a function returning `""` cannot pass.

## Measured, not claimed

Three shapes, driven through a real `db.transaction` in the same position as `run.ts:316-340`, with the **unredacted string printed beside the redacted one as a control** — so a clean result cannot be a broken grep:

| Shape | Before | After |
| --- | --- | --- |
| `ENOTFOUND` — typo'd or wrong-region host | host | **none** — `getaddrinfo ENOTFOUND [host]` |
| Postgres' own sentence | user | **none** — `role "[user]" does not exist` |
| `CONNECT_TIMEOUT` — suspended compute | host, port | **none** — `write CONNECT_TIMEOUT [host]:[port]` |

The middle row is why the remedy is value-based: **that sentence is written by the database server**, not by this codebase or its driver. There is no pattern that could have been written for it in advance.

### The migration's claim, measured on both halves

The migration rests on one claim: a **placeholder** leaves source health arithmetically identical where **NULL** would not. Equality alone would not prove that — *"placeholder equals baseline"* is also what you would see if nothing read the column, which is the false premise I started from. So three states were measured on the same rows, with the health expressions copied from `catalogue.ts` (re-indented only):

| Source | Baseline | After placeholder | After NULL |
| --- | --- | --- | --- |
| `always-failing` — 3 failures, never succeeded | completed=3, **failures=3** | completed=3, **failures=3** | completed=3, **failures=0** |
| `recovered` — failed, succeeded, failed | completed=4, failures=1 | completed=4, failures=1 | completed=4, **failures=0** |
| `clean` — 2 successes | completed=2, failures=0 | completed=2, failures=0 | completed=2, failures=0 |

**The two expressions, copied from `src/api/catalogue.ts` so this claim outlives the script that used them** — re-indented to fit this block and otherwise unchanged, checked programmatically against the source rather than by eye. The deleted artefact can attest that the measurement *ran*; it can no longer show that these were copied rather than re-derived — and a re-derived expression agreeing with itself is exactly what a three-state control cannot catch. It would still show equality-then-difference while measuring a private reimplementation.

```sql
(select count(*) from ingest_runs r
   where r.source_id = s.id
     and (r.error is not null or r.finished_at is not null))::int as completed_runs,

(select count(*) from ingest_runs f
   where f.source_id = s.id
     and f.error is not null
     and f.started_at > coalesce(
       (select max(ok.started_at) from ingest_runs ok
          where ok.source_id = s.id and ok.error is null and ok.finished_at is not null),
       '-infinity'::timestamptz))::int as consecutive_failures
```

**Read the first row.** A source that has never once succeeded, reported with zero consecutive failures — **healthy** — by a migration whose comment said it changed no health verdict.

**The third column is a control on the FINDING, not on the fix.** If NULL had not moved the numbers, the defect reported above was imaginary and the rest of this PR would deserve no trust either. The fixtures are built to expose it: a source that never succeeded is what exercises the `'-infinity'` bound, which is where NULL does its damage. **Fixtures that could not show the defect would have made all three states agree and proved the opposite.**

**Provenance, because output alone cannot be audited.** This came from `scripts/tmp-health-equality.ts`, a throwaway that built the three fixtures in its own scratch database (`ai_radar_health_check`, created and dropped in the same run, never the pipeline test database), printed the three states, and exited non-zero unless equality *and* difference both held. It ran, it passed, **and it has been deleted** — a verification script nobody will run again rots into a lie. Its output is above; its absence from the diff is deliberate, not an oversight.

### The same column has the opposite failure mode, and one predicate guards both

Raised in review, and the code was already right — **the reasoning beside it was not.** Both statements are scoped `WHERE ... IS NOT NULL`. Drop that predicate and the placeholder lands on every row, successes included; because health reads this column's **presence** rather than its content, the watermark subquery then finds no run with `error IS NULL`, coalesces to `'-infinity'`, and counts **every** run as a consecutive failure.

| Mistake | Result |
| --- | --- |
| Writing `NULL` too widely | every failure reads as a success — **all HEALTHY** |
| Writing the placeholder too widely | every success reads as a failure — **all FAILING** |

**I had documented one direction.** The next person to touch that predicate would have read the comment rather than re-derived both, and the comment covered the half I happened to hit. Both are now named in the migration file.

### The controls and the suite

| Check | Result |
| --- | --- |
| Targeted | **70 passed across 7 files** |
| **Deletion control** — redaction disabled | **13 red across 3 files**, including by name `writes no credential fragment into lastError when a pass fails` |
| Migration health equality | **PASS** — identical on the placeholder, different on NULL |
| Full suite | **418 passed across 34 files, zero skipped** |

### Why the control reddened thirteen tests and not six

**A control reddening wider than predicted is usually a smell. Here it is the evidence, and the next reader's instinct will be the wrong one.**

I expected six — the #99 tests. It also reddened #89's `describe-db-error` tests, because **both callers now go through the one implementation in `src/db/redact.ts`**, and the proof that a module is genuinely shared is that breaking it breaks everything that shares it.

**Two copies would have produced a narrower, tidier-looking red and a worse codebase**, and nothing in the cleaner number would have said so.

### 1. My migration would have reported every broken source as healthy

I wrote a migration to clear the old rows and justified it in the comment: *"source health reads `ingest_runs`' own columns, not this text."* **Then I checked my own sentence, because a comment is a claim. It was false.**

```
completed_runs        error IS NOT NULL OR finished_at IS NOT NULL
consecutive_failures  error IS NOT NULL AND started_at > newest run with error IS NULL
```

`SET error = NULL` does not erase text — **it converts every recorded failure into a success**, moves the last-succeeded watermark forward, and drops `consecutive_failures` to zero. The signal whose job is to say *"this feed has been refused every run for a week"* would have said the opposite, silently and permanently.

**The column's PRESENCE was the signal, not its content.** The fix is a placeholder: the column stays `NOT NULL`, every health count is arithmetically identical, and the text says what happened to it instead of pretending there was no failure.

**Worth naming where this would have shipped: inside a security fix.** A hardening PR is the least-inspected change in a repository, so a data migration hidden in one is the least-inspected change of all.

### 2. My own measuring tape had a vacuous assertion

The first run of the three-shape probe used **one fragment map across all three shapes**, so for `CONNECT_TIMEOUT` the host check was pointed at a different hostname and **could never have matched**. It reported "port only" and I nearly wrote that down. Re-run with per-shape fragments, which is the table above.

**Same defect as grepping for the shape of the old output — in the instrument, one day after recording that rule.**

- **Redacting in `describeError` rather than at the three log sites** is the PM's ruling on #99, and it was my recommendation before it was their ruling.
- **Moving the redaction to `src/db/redact.ts`** rather than exporting it from `client.ts`: raised before working around it, as instructed. A pipeline module should not import a connection pool to redact a string, and a second copy would drift. The new file has **no imports and opens nothing**.
- **The placeholder rather than NULL** follows from the health finding above.

## Also in here

- `src/db/redact.ts` — the #89 redaction, moved so two callers share one implementation.
- `drizzle/0003_clear_stored_error_text.sql` — the placeholder migration, with its reasoning in the file.


zarghammir:
REVIEW VERDICT: APPROVE

Reviewed at head **`0ab3081`**, 0 commits behind `origin/main`, four check-runs `completed/success`, `MERGEABLE/CLEAN`. Read-based; no machine taken.

The migration is the part worth the attention and it survives it.

## The migration — right, and guarded in a direction the comment does not claim

The finding it corrects is real and I verified the mechanism against `catalogue.ts` rather than against the summary: `completed_runs` counts `error is not null OR finished_at is not null`, and the last-success subquery keys on `error is null AND finished_at is not null`. So `SET error = NULL` would indeed convert every recorded failure into a success, move the watermark forward and zero `consecutive_failures`. **A migration that hid an outage while claiming to remove a credential** is the right description, and the placeholder keeps the column `NOT NULL` so every count is arithmetically identical.

**And there is a second falsification the `WHERE` clause prevents, which the comment does not mention.** Both statements are scoped:

```sql
UPDATE "ingest_runs" SET "error" = '[cleared …]' WHERE "error" IS NOT NULL;
```

Without that predicate the placeholder would be written to **every** row including the successes, and since health reads the column's presence rather than its content, every success would become a failure and every source would report FAILING. That is the exact mirror of the defect that was found — same column, same mechanism, opposite direction. The code is correct; the reasoning recorded beside it covers only one of the two ways it could have gone wrong, and the other is worth a clause in that comment while someone is still looking at it.

The honesty about what was not checked is also right: *"NOT VERIFIED HERE — that check needs the production URL or the credential, and this lane has neither, by design."* A lane that cannot verify something saying so beats a lane that can and does not.

## The three-state control is a control on the finding, and the fixtures can show it

The framing is the part that makes it work: *"'placeholder equals baseline' is also what you would see if nothing read the column, which is the false premise I started from."* Equality alone is consistent with the column being ignored, which is precisely the belief that produced the original defect — so the NULL column is what distinguishes "the placeholder is safe" from "nothing was ever at stake."

And the fixture-discrimination question is answered rather than assumed: **a source that never succeeded is what exercises the `'-infinity'` bound**, which is where NULL does its damage. *"Fixtures that could not show the defect would have made all three states agree and proved the opposite."* That is the right test to apply to one's own fixtures, and it is the same property that had to hold for #86's two-source fixtures — a fixture that looks adequate and cannot express the failure produces a green that means nothing.

### The deletion preserves provenance and destroys fidelity, and the cost falls on the checkable claim

The provenance is complete: the script is named, its scratch database is named and was created and dropped in the same run, its exit condition is stated — non-zero unless equality *and* difference both held — and its deletion is declared deliberate. *"A verification script nobody will run again rots into a lie"* is a fair reason, and this is the right way to retire one.

But two different claims were attached to that script and the deletion treats them differently:

- *"It ran and it passed"* survives as an **attested description**. That is what provenance buys, and it is bought.
- *"The health expressions were copied verbatim from `catalogue.ts` rather than re-derived"* does **not** survive. The only artefact that could have shown it is gone.

That matters because a re-derived expression agreeing with itself is exactly the failure mode a three-state control cannot catch — the control would still show equality-then-difference while measuring a private reimplementation. I checked what I could: the expressions **as recorded** in the migration comment and in the PR body match `catalogue.ts`. Whether the deleted script used those same expressions is attested rather than checkable.

Not a blocker, and I would not keep the script to fix it. Quoting the two SQL fragments verbatim in the body next to the table would have cost four lines and made the fidelity claim survive its own artefact.

## `redact.ts` is pure, structurally rather than by inspection

**Zero imports.** The only environment access is `process.env.DATABASE_URL` inside the function. There is no top-level `const`/`let`, no `postgres(`, nothing to execute on import. A module with no imports cannot open a connection, which is a stronger guarantee than "I read it and it looks fine."

And the consumer side holds: `run.ts` takes `import type { Db } from "@/db/client"` — a type-only import, erased at compile time — and the value import is `redactConnectionParts` from `@/db/redact`. So the pipeline does not pull the module that owns a pool in order to redact a string. Two importers, one implementation, no duplication.

## The 13-versus-6 argument is right, and the import graph proves it more strongly than the count

Deleting the redaction reddened #89's `describe-db-error` tests as well as the new pipeline ones, and the argument is that this is evidence the module is genuinely shared. It is — and the structural version does not depend on counting: `client.ts:3` and `run.ts:3` import **the same file**, so any mutation inside it necessarily reaches both consumers' suites. A blast radius that stopped at one of them would have meant the sharing was nominal.

*"Two copies would have reddened half as much and looked tidier"* is the useful inversion: a small, neat blast radius from a shared-module mutation is a symptom, not a virtue.

## The per-shape probe shipped, and the instrument defect is the interesting half

The table carries per-shape fragments, and the caught error is recorded: the first run used **one fragment map across all three shapes**, so the `CONNECT_TIMEOUT` host check was pointed at a different hostname and could never have matched — reporting "port only" for a shape where the host was present. *"I nearly wrote that down."*

That is a check aimed at the wrong target returning a confident answer, which is the same family as a sweep whose predicate is narrower than its claim. Catching it inside the measurement rather than after publishing it is the difference, and recording it is what stops the next person re-deriving the same map.

## The write-site test has two floors and reads back out of the database

`writes no credential fragment into lastError when a pass fails` builds its fragments from the live `process.env.DATABASE_URL`, floors on `fragments.length > 0` so an empty map cannot pass vacuously, reads the row back with `db.select().from(sources)` rather than trusting what was passed in, and floors again on `expect(row.lastError).toBeTruthy()` — *"something was stored, so this is not vacuous."* A negative assertion with a positive beside it and a floor under the input. That is the right shape for a test whose whole content is an absence.

## The dev-collision caveat — agreed, and worth one more line than it has

Redaction by value means a feed URL sharing a token with the credential has that token redacted too. Against a Neon hostname this cannot happen; against `localhost` in development it can. I agree it is conservative in the safe direction — an over-redacted dev log costs a reread, an under-redacted production log cannot be taken back.

Worth noting that the `>= 3` filter admits the **port** as a candidate, so in a development environment a feed error mentioning `5432` would come back as `[port]`. Same direction, same cost, and it is the likeliest of these collisions to actually be seen.

## Verified

The migration's two statements and both `WHERE` predicates · the health expressions against `catalogue.ts` · `redact.ts`'s zero imports and absent top-level state · `run.ts`'s type-only import of `client` · both importers of `redact.ts` · the write-site test's two floors and its read-back · the per-shape table · the provenance paragraph's completeness.

## Not verified

That the deleted script's SQL was verbatim rather than re-derived, as above. That `last_error` is null on all seventeen sources in production — that is the PM's reading and the migration correctly does not depend on it. And the exposure itself: I did not fetch `/api/sources`.

zarghammir:
REVIEW VERDICT: APPROVE

Verdict refresh at **`079654d`**, replacing my approval at `0ab3081`, one commit behind. CI polled to completion: four check-runs all `completed/success`, `MERGEABLE/CLEAN` — it was `UNSTABLE` with two jobs still running when I started, and I read the conclusions rather than the colour.

**The explicit question, answered explicitly: no. There is no added line that is not a comment.** Checked two ways rather than by eye:

- every added line in the diff begins with `--`;
- stripping comments and blanks from both versions of `drizzle/0003_clear_stored_error_text.sql` and diffing produces **no output** — the executable SQL is byte-identical.

Thirteen lines, all comment, no statement touched.

**And the comment is correct, including the step that makes it specific.** It does not stop at "the placeholder would be written everywhere" — it carries the mechanism through: with no row left where `error IS NULL`, the watermark subquery returns nothing, coalesces to `'-infinity'`, and every run counts as a consecutive failure. That is exactly how `catalogue.ts` behaves, and naming the `-infinity` step is what makes the paragraph checkable instead of plausible.

Pairing the two directions in one place is the right shape:

```
writing NULL too widely             -> every failure reads as a success (all HEALTHY)
writing the placeholder too widely  -> every success reads as a failure (all FAILING)
```

**On the wording correction — it is right, and it landed in one of the two places it appears.**

The corrected sentence is accurate: *"re-indented to fit this block and otherwise unchanged, checked programmatically against the source rather than by eye."* That is what the evidence supports, and narrowing it after finding the check had normalised whitespace is the correction I would want.

But the sentence introducing the same table still reads *"with the health expressions copied **verbatim** from `catalogue.ts`"*. The old word survives in the more prominent of the two places — the one a reader meets first — while the precise version sits below the table. Worth one edit, and worth it specifically because this is the claim that cannot be checked against the deleted artefact: the looser word is the one doing the load-bearing work.

Issue #102, opened 2026-09-17
One feed, one filter: the app opens on what people built, one control widens it to everything


zarghammir:
**Owner's ruling, 2026-09-17:** *"I don't want to have another page… Let's just not add multiple sections. Honestly, these are all in the same category… It's just a filter that we're adding. Let's just set the fun part as a default, and if they want to get more AI information or news, they can add the filter or change the filter based on that."*

## This is NOT #71, and I conflated them

I presented this to the owner as if it were #71 and it is not. **They are two different axes and both are real:**

- **#71 (ruled 16 Sept)** — *what gets collected and kept.* AI-matched vs adjacent tech. The keyword check stops being a fetch-time gate and becomes a stored label.
- **This ticket (ruled 17 Sept)** — *what leads when you open the app.* Things somebody **built** vs the reporting layer around them. Everything here is already collected and already AI-matched.

One is about the contents of the database; the other is about the front door. **My roadmap page described this one under #71's number, which was wrong and is corrected.**

## The design: ONE control, and #71 does not add a second

The owner was explicit that he does not want sections or pages — *"it's just a filter."* So:

- **Default view: built things.** `MODEL`, `TOOL`, `RELEASE`, `SHOW_HN`, `PAPER`.
- **One control widens it to everything stored:** adds `NEWS`, `FUNDING`, `POLICY`, `ANALYSIS`.
- **The reader's choice is remembered on the device** — device-held, per the saves ruling; it is a view preference, not shared state.

**#71 does not introduce a third stop. It widens what "Everything" contains.** Today "Everything" means every stored story and they are all AI-matched, because the gate is still discarding the rest. After #71 lands, "Everything" quietly includes adjacent tech. **The label stays true at both stages** — which is why this control never needs to grow.

A middle "AI only" stop is *meaningless today*: with the gate still on, AI-only and Everything are the identical set. If #71 lands and readers want AI news without adjacent tech, that stop can be added then, with something real to exclude. **Do not build it now** — a control whose two positions produce identical output is a lie about what it does.

## Why it matters

Twelve of seventeen sources are newsrooms, so **the brief fills with announcements by arithmetic**, not because anyone chose it. The owner has said twice, in different words, that this is not the app he wanted: *"I don't want to just make a generic AI news setter… focus more on the fun part, on the startups, on cool things."* This is the smallest change that makes the app open on what he asked for.

## Scope

1. Default view filters to the built content types. **No navigation changes beyond point 4.**
[...]

Issue #67, opened 2026-09-17
Nothing ever deletes old articles — plan retention before the database fills


zarghammir:
**Nothing in this codebase ever deletes anything.** Confirmed rather than assumed: no `delete` call and no retention logic anywhere outside tests. Every article ever fetched stays forever.

Raised by the owner, and he is right to raise it now rather than when the database is full.

### The measured numbers, from the dev database

516 raw items = **768 kB**; 504 stories = **416 kB**. So roughly **2.3 kB per article** all-in.

The second fetch of the same sources added **zero** new rows — the fingerprint dedupe works — so growth is only genuinely new articles: call it 150–400 a day across 18 sources. That is **roughly 125–340 MB a year**.

Neon's free tier is 0.5 GB. So this is **not urgent and not ignorable**: nothing breaks this month, and it fills in somewhere between one and three years — sooner if sources are added, and Phase 3 adds sources.

### The rule, in the owner's words

Delete old articles that nobody saved. Keep everything he kept.

### What makes this dangerous to get wrong

1. **Saved means saved, forever.** A story he saved must never be deleted, and neither may the raw items it is built from — a story with no items is a card with no sources, which is worse than no card.
2. **Read is not "wanted", and unread is not "unwanted".** Deleting on read state would delete the things he never got to.
3. **The ranking window is seven days.** Anything inside it is live data, not history.
4. **Absence is not permission.** A story with no read row and no saved row means nobody has looked at it yet — the same trap this project has found a dozen times. The rule must key on AGE, not on the absence of a signal.
5. **Deletion cascades.** `raw_items` cascades from sources; check what else cascades before writing a delete, and whether removing items orphans a story or silently changes its verification.

### Suggested shape, for the roadmap page when this is dispatched

- A retention pass in the worker, after ingest and ranking, not a separate schedule.
- Default: delete stories and their items older than **90 days** that are neither saved nor referenced by anything saved. Configurable, and **dry-run by default** like the backfill — print what it would delete, delete only with a flag or an explicit setting.
- Report what it removed in the same per-source line the ingest already prints, so it is visible rather than silent.
- Never touch `saved_items`, `read_state`, `user_preferences`, `sources`, `topics`.

### Acceptance needs a control

Seed a saved story older than the cutoff, run the pass, and confirm it survives **with its items**. Then unsave it, run again, and confirm it goes. If the first half passes and the second does nothing, the pass is not deleting at all and the test proves nothing.

[...]

PR #105, opened 2026-09-17
Keep what the AI gate used to discard, and label it (#71)

### LINKED ISSUE #71: AI by default, adjacent tech kept and filterable — the keyword check becomes a label, not a gate
**Owner's ruling, 2026-09-16:** "Mainly focus on AI, but if there is, for example, a tech startup that is growing, we can also put that in the list. It could have filters — AI only, everything — but mainly the app should be focusing on AI."

### The decision that matters is where the filter lives

Today the AI keyword check is a **gate at fetch time**: an item whose title never says AI, model, agent or similar is discarded and never stored. That makes the owner's ruling impossible to implement — you cannot offer an "everything" view over things you threw away.

**So the check moves from a gate to a label.** Items are stored either way, carrying whether they matched. The filter becomes a view preference rather than a fetch decision.

- **Default: AI only.** The app is an AI radar; that is the front door.
- **Adjacent tech is kept, not shown by default** — a fast-growing developer tool or startup that never says "AI" is exactly what the owner wants findable.
- **The reader chooses**: AI only, or everything.

### Why not simply loosen the keywords

Because "interesting to a founder or engineer" is not a keyword problem, and widening the list until it catches those items would also catch everything else. Labelling is honest about what it knows: *this matched our AI vocabulary; this did not.*

### Scope

1. **Storage** — a flag on the item or story recording whether it matched the AI vocabulary. Name it for what it is, not for what it gates.
2. **Ingest** — stop discarding non-matching items from **discovery sources** (Show HN, and GitHub when it lands). News sources keep their current behaviour; nobody wants an unfiltered newspaper.
3. **Read** — the brief and Radar filter on it, defaulting to AI only.
4. **UI** — a control. Radar's chips are the natural home; Settings is the natural home for the default.

### Consequences to decide rather than discover

- **Retention gets more urgent (#67).** Keeping what we used to discard means the database grows faster than the 125–340 MB a year measured there.
- **Ranking**: an adjacent-tech item that clears the bar should not outrank AI items merely for being fast-moving. Whether AI-matching is a ranking input or only a filter is a real question — answer it explicitly rather than by default.
- **Do not let the flag become a second verification signal.** It says what vocabulary a title used, nothing about whether the thing is true or good.

### Acceptance needs a control

Ingest a real item that does not mention AI from a discovery source, and confirm it is **stored** and **absent from the default view** — then switch the filter and confirm it appears. If the second half passes while the first is untested, the item was probably never stored.

Owner: Intelligence lane for 1–3, UI lane for 4. Blocks nothing in #69 except the filter behaviour itself.

zarghammir:
Closes #71 — points 1 to 3. The reader-facing control is #102, UI lane.

> ## ⛔ BEFORE MERGE — re-stamp the migration
>
> **The migration's `when` must be raised above main's maximum at the moment this merges, with its `idx` moved to follow.** Not now: the right number depends on what has landed by then.
>
> Drizzle selects migrations with `Number(lastDbMigration.created_at) < migration.folderMillis` — **strictly greater** (`pg-core/dialect.js:62`). This branch holds `0004` at `when=1789695986195`. #65 holds `0005` and `0006` at `…987195` and `…988195`. **If #65 merges first, this migration sits below main's maximum and is never selected.**
>
> **CI cannot catch this and its green is about the wrong scenario.** `compose-smoke` starts from an empty database, so `lastDbMigration` is `undefined`, the `!lastDbMigration ||` short-circuits, and every migration applies whatever its timestamp. The at-risk state is a database that has *already applied* the preceding migration, which CI never reaches because CI always starts clean. #109 covers it with a two-step migrate.
>
> **What a skip would look like here:** this migration *adds* columns that the code then queries — `stories.adjacent_tech` is in the brief and Radar `WHERE` clauses. So a skip is not a quiet no-op, it is a SQL error on every brief request. Loud rather than silent, and neither is acceptable.

The app decides whether a Hacker News post is about AI by looking for words like **ai**, **model** and **agent** in its title. If none appear, the post was **thrown away and never stored**.

The owner asked for something that cannot be built on top of that: *mainly AI, but a growing tech startup can be in the list too, with filters — AI only, or everything.* **You cannot offer someone "everything" over things you discarded.**

So the check stops being a decision about what to keep and becomes a **label**. The item is stored either way, carrying whether it matched. Which of them a reader sees is then a choice rather than a fact fixed at fetch time.

## Nothing changes in the default view, today or on merge

This is the part to be clear about, because the ticket's urgency reads as *we are losing items every thirty minutes and this stops it.*

**No seeded source opts in.** `keywordPolicy` defaults to `"gate"`, which is exactly today's behaviour. The only source that gates at all is the Hacker News front page, and the ticket excludes it by name — a general technology firehose ungated is the unfiltered newspaper nobody wants.

**So this ships the mechanism, and the saving starts when a discovery source exists.** That is Show HN, in #80, which is rebased and waiting. Until then the columns fill in and nothing is kept that was not kept before.

I would rather say that plainly than let the PR imply a recovery it does not make.

## Two columns, each named for what it is

| Column | Means | Default |
|---|---|---|
| `raw_items.matched_ai_vocabulary` | did this title use AI words | **NULL** for rows written before this |
| `stories.adjacent_tech` | kept deliberately, not shown by default | `false` |

**The nullable one is deliberate.** A row written before this was never evaluated, and *not evaluated* is a different answer from *did not match*. Reading one as the other would hide stories that have been visible all along. It cannot be filled in SQL either — the vocabulary is forty keywords with a whole-word rule for short tokens, a substring rule for long ones, and #73's hyphen handling on top — so `npm run backfill:ai-vocabulary` does it through the real matcher.

**`adjacent_tech` defaults false, and that is what makes this change inert.** Every story that exists stays in the front door.

A story is adjacent tech only when **both** hold: no item used AI vocabulary, **and** every item came from a source that carries more than AI. One matching item makes a story AI. One item from an AI-curated feed makes it AI by provenance, even if that headline never says the word. A null match is not a miss.

## The ranking question, answered rather than defaulted

**AI-matching is a filter input and never a ranking input.** Written into the source beside the filter, with the reason, so it is not re-derived from the code's general tone.

Inside the default view every story is AI, so a boost would be **constant across the whole set** — a ranking input that never varies within the view it applies to is a no-op with maintenance cost. And in the widened view it would systematically bury the adjacent tech the widening exists to surface, which inverts the ruling that created the ticket.

## The acceptance control, both halves

The ticket warns that the second half passing while the first is untested means the item was probably never stored.

**First half — `run.test.ts`, through a real ingest.** A label-policy source fetches a title that genuinely matches nothing, and the test asserts the item **is stored**, that its flag is `false`, and that its story is adjacent tech. The fixture asserts its own premise first: if someone later adds a keyword that matches it, the test says so rather than quietly changing meaning.

**The control beside it** — the same title through a gating source is still discarded. Without that, the first test would pass equally well if the gate had simply been deleted, which would empty the front door rather than widen the back one.

**Second half — `api.test.ts`.** The adjacent story is absent from the brief and from Radar by default, and present with `?view=everything`. That assertion checks the AI story is **still there** too: a widened view that swapped the set rather than extending it would pass a test that only looked for the new row.

**And an unrecognised `?view=` is the default, not the wider set.** An unknown value must not silently open the front door.

## Where the vocabulary lives now

It was inside the Hacker News adapter while that adapter was its only user. It has two now — the gate, and the labelling in `normalizeItem` — so it moves to `pipeline/normalize/ai-vocabulary.ts`. **A second copy would let the gate and the label disagree about the same title.**

## Noted, not fixed

**Retention (#67)** gets more urgent the moment a discovery source starts keeping what it used to discard. Nothing here addresses it, and the growth it measured assumed the old volume.


zarghammir:

Reviewed at `1023b6b`, read-only, in my own worktree. Base is `d8ec412` = `origin/main` exactly. All seven checks polled to conclusion, every one `COMPLETED/SUCCESS`; `MERGEABLE/CLEAN`. No conflict markers survive anywhere in the tree — I checked, given what the head commit message says.

**The failure mode you asked me to confirm is confirmed, and it is broader than stated.** Two findings below, both small fixes, both in the same family: **an unrecognised value being read as a definite answer.** This PR gets that right twice and wrong once, and the once it gets wrong opens the front door.

---

# ① The adapter's policy check fails OPEN, and the two halves disagree

`src/sources/hackernews/adapter.ts:97`

```js
const keywordPolicy = String(source.config.keywordPolicy ?? "gate");   // :54
...
if (keywordPolicy === "gate" && !matched) continue;                     // :97
```

**The default is narrow and correct. Anything unrecognised is wide.** `"gate"` gates; `"label"` keeps; and so does `"labl"`, `"Label"`, `"gated"`, `"gate "` or any other typo — because the only string that gates is the exact literal `"gate"`. A misspelling in a source's config **silently turns the AI gate off** for that source.

**The other half of the same key gets it right, which is what makes this a defect rather than a preference.** `src/pipeline/run.ts:230`:

```js
const fromLabelSource = (i) =>
  String((i.sourceConfig)?.keywordPolicy ?? "gate") === "label";
```

Here the *widening* branch requires the exact literal. So with `keywordPolicy: "labl"` the two sides disagree, and the disagreement composes into the worst available outcome:

| | `keywordPolicy: "labl"` |
| --- | --- |
| adapter `:97` | `"labl" !== "gate"` → **item kept**, `matchedAiVocabulary: false` |
| run.ts `:230` | `"labl" !== "label"` → **not a label source** → `fromLabelSource` false |
| ⇒ `refreshStory` | `adjacentTech = false` |
| ⇒ `notAdjacentTech()` | `adjacent_tech = false` matches → **the story is in the default view** |

**So one typo in a config JSON puts non-AI stories in the front door**, with the gate off, nothing raised, and nothing on screen to notice. That is the precise inverse of what this PR promises in three separate comments — *"this change adds nothing to the default view"*, *"every story that existed before it stays visible"*, *"`gate` keeps today's behaviour for every source that does not ask otherwise."* The promise holds for the default and for correct spellings, and fails for the one input nobody tests.

**Nothing catches it.** `sources.config` is untyped `jsonb` at `schema.ts:108` with no validation anywhere in the tree; `keywordPolicy` appears in exactly three places (the adapter's doc comment, `:54`/`:97`, and `run.ts:230`); and **every test uses the exact string `"label"`** — `run.test.ts:963`, `:1007`, `:1015`. There is no test with an unrecognised value, so the asymmetry is unpinned in both directions.

**This PR already states the governing rule, two files away.** `src/api/params.ts`, in the new `parseView`:

> *"Refusing is the strict form of 'an unknown value must never widen': nothing is opened at all. An earlier version of this narrowed silently, which was safe and was still the odd one out … **a new convention two lines from an existing one is how the next reader learns the wrong rule**."*

That reasoning is right, and it is exactly as true across two files in one PR as across two lines in one function. `parseView` is the strict form; `run.ts:230` is the safe form; `adapter.ts:97` is neither.

**The fix is one comparison, inverted so that widening requires the match:**

```js
if (keywordPolicy !== "label" && !matched) continue;
```

Then an unrecognised value **gates** — fails closed — and the adapter and `run.ts` agree on what a label source is, because both now key on the same literal. A test with `keywordPolicy: "labl"` asserting the item is discarded pins it. If you would rather refuse loudly than fail closed quietly, throwing on an unrecognised policy is the `parseView` form and I would accept either; what I would not accept is the current shape, where the safe default hides an unsafe edge.

---

# ② The null case is defended correctly in code and pinned by nothing

`src/pipeline/run.ts:228` is right, and the comment above it is the clearest statement of the hazard anywhere in this PR:

> *"A null match is NOT a miss. Rows written before #71 were never evaluated, and reading 'unknown' as 'did not match' would quietly hide stories that have been visible all along."*

```js
const knownMiss = (i) => i.matchedAiVocabulary === false;
```

The strict `=== false` is the whole defence, and `knownMiss` is the right name for it.

**But all three new tests use booleans only.** `run.test.ts:973` asserts `false`, `:1030` asserts `true`, `:1034` asserts the resulting `adjacentTech`. **No test has an item whose `matchedAiVocabulary` is NULL.** So:

```js
const knownMiss = (i) => !i.matchedAiVocabulary;      // mutation
```

**passes all three.** And that mutation is precisely the defect the comment warns against. Its effect on a real upgrade: between running the migration and running `npm run backfill:ai-vocabulary` — a manual command, by design — **every pre-existing row has NULL**. Under the mutation, any story whose items are all from a `label` source flips to `adjacentTech = true` and **disappears from the default view**. Silently, because `notAdjacentTech()` simply stops selecting it.

A comment stating a hazard, one operator defending it, and no test holding the operator in place is the shape where the next refactor removes the defence and nothing goes red. **The fix is one test**: an item with `matchedAiVocabulary` NULL beside a known miss, asserting `adjacentTech` stays `false`. That is also the deletion control for the line — flip `=== false` to `!` and it must redden by name.

---

# The skipped-migration failure mode: confirmed, and wider than "every story query"

You asked me to confirm the shape is a hard SQL error rather than a silent no-op. **It is, and I traced every path rather than reasoning from the schema.** `0004` adds `raw_items.matched_ai_vocabulary` and `stories.adjacent_tech`, and both are declared in `schema.ts`, so drizzle names them:

| path | how the column is named | on a skipped migration |
| --- | --- | --- |
| Radar list | `radar.ts:119` `.select({ story: stories, trend })` — the whole table object expands to every declared column | **error** |
| Radar default filter | `radar.ts:78` → `notAdjacentTech()` → `sql\`${stories.adjacentTech} = false\`` in the WHERE | **error** |
| Brief | `brief.ts` `db.select().from(stories)` — select-all | **error** |
| **Story detail** | `stories.ts:223` `db.select().from(stories).where(eq(stories.slug, slug))` — select-all | **error** |
| Histogram | `radar.ts:173` with `filterConditions` | **error** |
| Ingest | `run.ts:203` selects `rawItems.matchedAiVocabulary` by name | **error** |

So it is not merely "every story query" — **story detail is included**, which I checked because a by-slug lookup is the one that could plausibly have escaped. `column "adjacent_tech" does not exist`, on every reader-facing page and on ingest.

**And the opposite shape from #65 is worth naming, because it is counter-intuitive.** #65's migrations *drop* columns; a skip there is silent, and leaves an email column sitting in the database with its data in it. #105's *add* columns; a skip here is total and immediate. **The loud one is the safer failure** — it is noticed in minutes and recovered by running the SQL, with no data lost — whereas a silent one is never noticed at all. That does not lower the priority of merging this first; it does mean that if the order ever slips, the consequence is an outage rather than a quiet harm, and outages get fixed.

---

# What is right, and worth saying so

**The vocabulary extraction closes a real drift.** It was inside the HN adapter while the adapter was its only consumer; now the gate and the label read the same list, with the adapter re-exporting rather than keeping a copy. *"A second copy would drift — the gate and the label would then disagree about the same title"* is the correct reason, and it is the same reason `refuseUnlessInternal` was shared in #110.

**`parseView` handles absent versus unrecognised correctly**, and for the stated reason: `?view=` absent is no preference and gets the front door; `?view=evrything` is a typo and is told so rather than silently served something it did not ask for. Consistency with the other enum params in the same file is the right tiebreak.

**The first test names the half that usually gets skipped.** *"Testing only that the filter hides it cannot tell a working filter from an item that was never stored at all — which is the bug this ticket exists to fix."* Asserting the row exists **and** is labelled, rather than asserting it is absent from a view, is the difference between testing the feature and testing its symptom. The gating control beside it (`"still discards the same item when the source gates"`) is what stops the first test passing equally well if the gate had simply been deleted — and deleting the gate would empty the front door rather than widen the back one, so the control is load-bearing.

**The two-sources detail in the third test.** Using two sources rather than two items from one, because a fingerprint is scoped to source plus canonical URL, and *"an earlier version of this test proved only that"* — a fixture that collapses to one row cannot demonstrate a rule about several. Recording the earlier version's failure beside the fix is what makes it unrepeatable.

**The backfill is a command, not a deploy step**, dry-run by default, and it defers `adjacent_tech` to `refreshStory` rather than writing it — *"so the backfill cannot disagree with the pipeline about what makes a story adjacent."* One derivation, one owner. It also closes the sql handle in a `finally`, so a failed backfill exits instead of hanging.

**`matchedAiVocabulary` is documented as a fact about words and not a quality signal**, in both the schema and the module. Given how easily a stored boolean becomes a ranking input, saying it twice is not redundant — and `notAdjacentTech()`'s own comment refuses the score boost with a reason rather than an assertion: constant inside the default view, and actively harmful in the widened one.

---

## To clear this

1. `adapter.ts:97` — invert so widening requires the literal (`!== "label"`), or throw on an unrecognised policy. Add a test with a misspelled `keywordPolicy` asserting the item is **discarded**.
2. `run.test.ts` — one test with `matchedAiVocabulary` NULL asserting `adjacentTech` stays `false`, and confirm it reddens when `=== false` becomes `!`.

Neither needs the machine beyond the suite you already run. I will re-read on a new commit.

**Merge order still stands: this one goes in before #65.** My #65 approval is conditional on it, for the reasons on that PR.

zarghammir:
REVIEW VERDICT: APPROVE

Re-read at `71a29a9`, read-only, in my own worktree. Three files, +79/−2 since `1023b6b`. Both findings fixed, and fixed at the root rather than at the symptom. Nothing in flight; `MERGEABLE/UNSTABLE`, and the `UNSTABLE` is the Vercel check, which I treated as a claim to verify rather than an instruction to ignore.

**The Vercel red is an account build-rate limit and not this PR.** I checked it independently rather than accepting it: the target is `vercel.com/zarghammirs-projects?upgradeToPro=build-rate-limit` — an account settings page, not a deployment — and it is **byte-identical across #105, #108 and #110**, while **#65 shows `SUCCESS` with a real deployment URL**. A failure that lands on three PRs with the same account-level link and spares the one that has not pushed since the window rolled is a quota, not a build. **Every real check is green**: lint/typecheck/test/build, compose-smoke, browser checks, both README jobs. Agreed on not re-pushing to clear it — that spends more of the quota to prove something the URL already says.

---

# The two comparisons now key on the same literal, in the same direction

```js
// adapter.ts:113
if (keywordPolicy !== "label" && !matched) continue;

// run.ts:232
... ?? "gate") === "label";
```

Both ask whether the policy **is `"label"`**, so widening requires the exact match on both sides and an unrecognised value **gates** in the adapter and **is not a label source** in `run.ts`. The two halves agree about `"labl"`, `"Label"`, `"gated"` and every other typo, and they agree in the safe direction. The item is never stored, so `run.ts`'s branch is moot — which is the right shape: the narrow decision is made once, at the earliest point.

**The comment carries the reasoning rather than the conclusion**, which is what I wanted and more than I asked for. It names the counterfactual — *"if this asked whether it is `gate`, the two halves would key on different literals in opposite directions and a typo would fail OPEN here and NARROW there"* — so the next person to touch this line can see what the shape was, not just what it is now.

## The `parseView` distinction is stated as a distinction, not asserted as a difference

> *"parseView rejects a value a READER supplied in a query string: refusing gives that caller feedback and costs one request. This is OPERATOR config read at fetch time, where throwing takes the whole source dark on a typo. Both obey 'an unknown value must never widen'; refusing here would trade a wide failure for a dark one."*

That is the right test and the right answer. **The rule is shared and the blast radius is not**, so the same rule resolves differently — and saying which of the two costs is being chosen is what stops the next reader reading the pair as an inconsistency and "fixing" one of them. The PM's ruling is correct and the file now explains it where it matters rather than in a PR comment nobody will find.

---

# The NULL test is a real control, and I traced it rather than trusting the label

**It reddens under the mutation.** Walked through:

| | `matchedAiVocabulary === false` (shipped) | `!matchedAiVocabulary` (mutation) |
| --- | --- | --- |
| `knownMiss(NULL)` | `null === false` → **false** | `!null` → **true** |
| `items.every(knownMiss)` | false | true |
| `items.every(fromLabelSource)` | true (`keywordPolicy: "label"`) | true |
| `adjacentTech` | **false** | **true** |
| `expect(story.adjacentTech).toBe(false)` | passes | **FAILS** |

And the construction is right in the way that matters: **one item, from a label source**, so the NULL is the only thing standing between this fixture and `adjacentTech = true`. A control whose subject is prevented from exhibiting the property for a second reason cannot redden, and this one has no second reason.

**Setting the column back to NULL after a normal ingest is the correct way to reach that state**, because it is the state the database is actually in — `UPDATE ... SET matched_ai_vocabulary = NULL` reproduces the window between the migration and the manual backfill rather than approximating it with a hand-built row that skipped the pipeline.

## Non-blocking: the control's precondition is computed and not asserted

`run.test.ts:1042` fetches `before` and uses only `before.id`:

```js
const [before] = await db.select().from(schema.stories);
await db.transaction(async (tx) => { await refreshStory(tx, before.id); });
```

**`before.adjacentTech` is `true` at that point, and nothing says so.** It is guaranteed today by the sibling test at the top of the describe, which uses the same source config and the same title and asserts `adjacentTech === true` — so the control is valid. But its validity lives in a *different test*. If this fixture ever changes so the story is not adjacent before the NULL — a different title, a source that stops being a label source — this test keeps passing, **and stops reddening under the mutation**, silently.

One line, on a value already in scope:

```js
expect(before.adjacentTech).toBe(true);   // the fixture exhibits the property before NULL removes it
```

That is the same floor-on-the-input discipline this PR applies elsewhere, and the same one I had to apply to my own comment-strip check on #65 after it compared two empty files and reported "identical". **Establish that the subject can exhibit the property before reading the control's verdict** — here the value is already fetched, so it costs a line.

**The misspelled-policy test is right as written.** `keywordPolicy: "labl"` asserting `rawItems` is empty is the assertion that would have failed before this fix and passes after, in the direction that matters — discarded, not merely unlabelled. Nothing tested that in either direction before.

---

# The pattern you asked me to weigh: yes, it is visible, and it is a third instance inside the fix for the second

I would not have taken this on the lane's own account, so I looked for it in the diff. **It is there.**

This PR extracts `DEFAULT_AI_KEYWORDS` out of the adapter into `src/pipeline/normalize/ai-vocabulary.ts`, with the reason stated explicitly:

> *"A vocabulary is not an adapter detail, and **a second copy would drift — the gate and the label would then disagree about the same title**."*

**That is, word for word, the argument for extracting the policy literal.** And `"label"` is still a bare string in two files:

```
src/pipeline/run.ts:232              ... === "label"
src/sources/hackernews/adapter.ts:113   ... !== "label"
```

coupled by nothing but two prose comments pointing at each other — *"The other half of this rule, in run.ts"* and *"Keyed on the same literal the adapter keys on"*. There is no shared constant and no shared predicate. **Rename the policy to `"labelled"` and edit one file, and the gate and the label disagree again, silently, exactly as the vocabulary would have.**

So the shape is not "a rule written in one file and broken in a sibling" so much as **a rule the author knows how to enforce structurally, enforced structurally in one place and left in prose in another, in the same PR, across the same two files.** The comments are the prose form; the extraction is the enforced form. This PR contains both forms and applies the weaker one to the thing that already broke once.

**That is a finding about durability, not correctness, which is why it is not blocking.** The behaviour is right today and the tests pin it. But the honest version of the lane's self-observation is sharper than the one it offered: it is not that the rule was forgotten two files away — it is that **the same file that demonstrates the enforced form reaches for the prose form when the duplicated thing is a string rather than a list.** A `LABEL_POLICY` constant, or an `isLabelSource(config)` predicate imported by both, is one small change and retires the whole family here.

Worth carrying to #111, which is the right home: `sources.config` being untyped `jsonb` is why a misspelling is silent, and a shared predicate is where validation would naturally live.

---

# What is right

**The fix is at the root.** Inverting the comparison rather than adding a validation branch means there is no new code path to get wrong, and the safe behaviour is the *default* of the expression rather than a case handled beside it.

**The `run.ts` comment is one line and load-bearing** — *"Keyed on the same literal the adapter keys on, so the two halves of this rule cannot disagree about an unrecognised value"* — because that file is where someone would otherwise "tidy" the `?? "gate"` and reintroduce the asymmetry.

**Both new tests name what was invisible before**, rather than describing what they do. *"Every other test in this describe passes the literal `label`, so the asymmetry this pins was invisible"* is the sentence that stops the test being deleted as redundant in six months.

---

**Merge order unchanged and now for a stated reason rather than an assumed one: #105 first, then #65 at `a4d6063`, which I have approved.** A skipped `0004` is a loud SQL error on every story query including detail; a skipped `0005`/`0006` is a silent no-op that leaves the email column with its data in it. Outages get fixed; silence does not.

Nothing merged, nothing pushed, no branch touched; `git status` clean. The box was not taken — reading, and `gh` calls to compare the Vercel target across four PRs.

zarghammir:
REVIEW VERDICT: APPROVE

Confirmation re-read at `52efb51`. **The diff since my approval at `71a29a9` is exactly the one change, and nothing else** — seven lines, all in `src/pipeline/run.test.ts`, no source file touched. All five real checks polled to conclusion, every one `COMPLETED/SUCCESS`, including browser checks, which were still in flight when I started and which I waited out rather than reading a pending as a pass. Vercel remains red on the account quota.

**The assertion is on the value already in scope, before the mutation point:**

```js
const [before] = await db.select().from(schema.stories);
expect(before.adjacentTech).toBe(true);        // <- added
await db.transaction(async (tx) => { await refreshStory(tx, before.id); });
```

So the control now establishes that its subject **can** exhibit the property before the NULL is introduced, rather than inferring it from a sibling. `before` was already being fetched for `before.id`; the assertion costs a line and no extra query.

**The comment states the failure mode rather than the fact**, which is the part that keeps it from being deleted as noise:

> *"…otherwise the final expectation is true for a reason that has nothing to do with NULL handling, and the control stops reddening without failing. It is guaranteed today only by the fixture in the sibling test above, which is one edit away from silently retiring this one."*

Naming the sibling as the thing it currently depends on is precise about what was wrong: not that the control was invalid, but that **its validity lived in a different test** and could be removed from there without anything here noticing.

---

**On overruling my non-blocking: you were right, and the tally is the argument.** I flagged it as non-blocking because the control was valid *today*. But "a control that quietly stops reddening" is the failure we have now hit in four instruments in one evening — including my own comment-strip check on #65, which compared two empty files and reported "identical". When a family recurs that often in that short a span, the right prior flips: a control whose validity rests on something outside itself should be treated as already broken, because the evidence says the outside thing moves. **My "non-blocking" was calibrated to the instance and yours to the pattern, and the pattern was the better instrument.**

That is worth keeping as a rule rather than as an anecdote: **a finding's severity is not only about this instance — it is about how often its family has fired recently.** The same defect deserves a harder line the fourth time than the first, and the count is the thing that should move it, not the argument.

---

**Ready to merge in the stated order: this first, then #65 at `a4d6063`, then #108 once its two items are closed.**

PR #108, opened 2026-09-17
One feed, one filter: the app opens on what people built (#102)

### LINKED ISSUE #102: One feed, one filter: the app opens on what people built, one control widens it to everything

zarghammir:
Closes #102.

**Stacked on #65** (`feat/today-live`), and based there rather than on `main`. A stacked PR reports `MERGEABLE/CLEAN` against its **parent**, so that green says nothing about `main` — do not read it as one. Merging #65 first is what makes this diff honest.

**Restacked 2026-09-17**, most recently after #110, #105 and #112 landed on `main` and #65 was rebased onto them. This branch sits on **`7b229d8`**, the current tip of `feat/today-live`, is **0 behind it**, and the **10 commits** here are the only ones it adds.

> That SHA and that count are **dated, not live**. A stacked PR's parent is rewritten by every rebase, so this sentence was already false once — it named `a4d6063` and "two commits" after the branch had moved past both. `git rev-list --left-right --count feat/today-live...feat/one-feed-filter` is the answer that cannot go stale.

**The app opens on what people built.**

Twelve of seventeen sources are newsrooms, so the brief fills with announcements by arithmetic rather than because anyone chose it. Now the front door shows models, tools, releases and papers — things somebody made — and **one control at the top widens it to everything**: the news, discussion, funding and policy around them.

**Research and Releases are gone from the navigation.** They were never destinations; they are two positions on this one control now. Their pages were 19-line placeholders and are deleted, because leaving them would leave two pages nothing links to.

## Why there are two positions and never three

With the fetch-time AI gate in place, **every stored story is already AI-matched** — so an "AI only" stop and "Everything" would return the identical rows. **A control whose two positions produce the same set is a lie about what it does.**

When #71 turns that gate into a label, it does not add a stop either: it widens what "Everything" already contains, and the label stays true at both stages. That is why this control never has to grow.

## Where the filter lives, and why it matters

| | |
|---|---|
| **In the query**, in `src/api/brief.ts` | So the route and the page pass the same thing and cannot disagree about what a brief is — the property #65's own comment promises |
| **Before the reading-time budget** | Filtering after it returns **three stories for a ten-minute brief**, because the budget was spent on rows the reader never sees. A bug a reader would feel and never be able to describe |
| **An additive parameter, defaulting to everything** | Omitting it preserves today's behaviour exactly |

It takes a **list of types**, not a view name: that module decides what a brief *is*, not what leads on the front door.

## Every content type has a side

A total `Record`, like `CONTENT_TYPE_LABELS` — so a type added to the database without a side is a **compile error** rather than a story that belongs to neither view and appears in nothing.

| Leads | Arrives when you widen |
|---|---|
| `MODEL` `TOOL` `RELEASE` `PAPER` `RESEARCH` | `NEWS` `DISCUSSION` `TREND` `BUSINESS` `REGULATION` |

**`RESEARCH` sits with `PAPER` deliberately.** Splitting them draws a boundary **the reader cannot see** — both render as a published piece of work somebody did — and a filter whose boundary is invisible reads as randomness. "Why is this in and that out" is the question that destroys trust in a default view.

## An unrecognised view never widens

My first version had the route do `parseView(raw) ?? "all"`, so `?view=evrything` returned **everything stored** — a typo opening the front door. `parseBriefLength` was the precedent two lines below: a bad length **throws** rather than guessing.

**Absent and unrecognised are kept apart, because they are different answers.** No `?view=` means the caller expressed no view, and the route answers what it answered before this PR. A `?view=` this build does not know is a caller asking for something specific that does not exist, and it gets told. Collapsing them would make a typo indistinguishable from silence.

The page cannot throw at a reader for a mistyped URL, so there it narrows instead. **On neither path can an unknown value produce more than the reader asked for.**

## The choice is remembered on this device — the choice, not the default

A **cookie**, for the same reason the brief length uses one: Today filters in the query on the server, so the view has to arrive **before this device runs any JavaScript**. It carries one word. URL first, then the device, then `built`.

**It stores what the reader actually asked for, and nothing when they have not asked.** The first version wrote on every render from the view Today had settled on, so somebody who never touched the control still got `ai-radar-view=built` written on their first visit. After that the stored value could not tell **"chose Built"** from **"never said anything, and Built was the default that day"** — so the day the default changes, every silent reader stays pinned to the old one, by a preference they never expressed.

It now writes only what `?view=` actually carried. A shared link still counts as a choice, because it carries the parameter, and every link this control renders sets it, so a click still writes. **Absence stays absent** — which is the same distinction `parseView` draws one file over, and the one this whole project keeps getting wrong in new places.

## Removing two tabs made an existing control weaker — the thing I would not have predicted

The accessibility sweep has a control that squeezes the phone viewport until the bottom bar **must** clip, so the bar gate can be watched going red.

Six tabs at 200px gave each about **33px** and forced every label to clip. **Four tabs at 200px get about 50px — roughly the width of the word "Settings".** The old number now sits on the boundary: it might still redden, it might quietly stop, and those look identical in a green run.

**A control that silently stops going red is worse than no control**, because the gate it guards keeps reporting a pass nobody can distinguish from an untested one. The width drops to **160px** to keep the same pressure per tab.

> ⏳ **The arithmetic says it must clip. Nobody has watched it.** Marked unverified in the file and in the script's own output. The next run with a browser has to see this control go red before its green means anything again.

## Measured, not claimed

**By CI at `1cb1db3`, on a base that already contains #65, rather than by me on my laptop.** Quoted as artefacts rather than as a verdict, because an earlier green on the sibling branch was green only because two checks **had not run at all**.

| what | measured |
|---|---|
| browser checks derived | **5**, and **6** browser-driving scripts accounted for (the sixth is the capture script, declared not-a-check) |
| controls that reddened as required | `a11y`, `verify:isolation`, `verify:saved` — each exit 1 |
| declared no control, named not absorbed | `verify:shell`, `verify:today` (#113) |
| checks that ran and passed | all five |
| **bottom bar** | `tabsPerBar = 4` — **this PR's nav change, measured**. The same job on #65 reported **6**. |
| accessibility sweep | `statesAudited = 20`, **0** violations, **0** serious, **0** horizontal overflow |
| reading-length switch | `all = 11` → `ten = 10` → `five = 5` |
| empty state, pre-seed | explained itself in **224 characters**, with `data-screen-state="quiet"` and no "cannot reach" clause |
| seeded corpus | `briefMinutes = 11` against a shortest budget of 5 |

**`tabsPerBar` 6 → 4 is the one worth pausing on.** It is this PR's removal of two destinations, measured from the DOM by a job that has no idea what this PR is about, on a branch where the only difference is this change.

### The 160px control, now actually watched going red

The body of this PR previously said this was unverified. **It has now been run, and the check that went red is the one it guards** — not a neighbour.

| | normal, 390px | control, `AUDIT_CONTROL=narrow` at 160px |
|---|---|---|
| exit code | **0** | **1** |
| labels clipped | **0** | **10** |
| narrowest tab | 97.5px | **40px** |
| widest clipped label | — | **"Settings" at 36px** |

That is the gate proving it can fail at **four** tabs, which is the whole reason the control's width had to move from 200px when #102 removed two. A gate that has only ever passed is not yet known to be able to fail.

**One thing the control turns up that I am not hiding:** at 160px a `color-contrast` violation appears on `/settings`. It is an artefact of squeezing the viewport to a width no phone has — it does not appear in any of the 20 real states — but the control's output reports it and so does this table.

## A control that had stopped detecting what it exists to detect

Found in review, and it is the worst kind of defect to ship in a PR whose subject is controls that stopped working.

`verify-saved-settings.mjs` proved that a brief length chosen in **Settings** reaches **Today**. It measured one side on a bare `/` and the other on `?length=all&view=all`. Since this PR, a bare `/` also means `view=built` — so the two sides differed on **length and view at once**.

**With four built things among seven fixtures, a Today that ignored the saved length completely would have shown 4 against 7. 4 is less than 7. The check passes.** It had been measuring this PR's filter and reporting on the reading-length preference.

The file's own docblock stated the reasoning it no longer had: *"a Today that ignored the preference would show the same count for both"* — true only while length was the only axis. **A file that documents a discipline it does not implement is worse than one that documents nothing**, because the next person reads the comment and believes the check is safe to edit around.

Both sides now pin `?view=all` and neither supplies a length. I swept the other browser scripts for the same shape rather than fixing the line: `verify-shell`'s bare `/` is the service-worker check and counts no stories, and the accessibility sweep uses the default view deliberately. **That was the only instance** — a measurement, not a hope.

## A CI check that pinned a wording, twice

The step asserting that an empty database renders a *quiet day* rather than a failure greps the page for words. It has now reddened **twice on a page that was behaving perfectly**: first on `quiet morning` when this PR's copy became *quiet day*, then on `quiet (morning|day)`, which is a wider pin and still a pin. Its comment claimed *"in whichever words suit them"* while the pattern said otherwise — the same defect one clause over, and worse for being documented as safe.

It now finds the empty-state block by a data attribute and requires its body to be **long enough to be an explanation**, with no opinion at all about the words. Marker-with-no-prose is the defect; a rewritten sentence is not.

| Case | Wanted | Got |
|---|---|---|
| the current copy | green | green, 154 chars |
| **the copy rewritten completely** | green | **green, 135 chars** |
| the marker present with no body | **red** | red — "no empty-state body at all" |
| a stub body ("Nothing.") | **red** | red, 8 chars |

The middle row is the whole point: it is the exact case that reddened CI twice.

## The acceptance, asserted rather than photographed

**#102's acceptance is that the two positions return different sets. CI now proves that on every run**, against rows the real pipeline wrote:

```
MEASURED: view=built -> 5   [RELEASE, RESEARCH]
MEASURED: view=all   -> 11  [BUSINESS, NEWS, REGULATION, RELEASE, RESEARCH]
RESULT: the two positions return different sets (5 built of 11 total).
```

**Six stories separate the two positions, and the three types that appear only on Everything — `BUSINESS`, `NEWS`, `REGULATION` — are exactly the three `VIEW_OF` sends there.** The control is doing what it says, on real data, not on a fixture.

**Why this exists when there is already a screenshot pair below.** A screenshot proves the sets differed **once**, on whatever corpus that machine happened to hold. Mine held five stories that were every one of them a `RELEASE`, so the capture script refused to write a pair that would have looked like proof and shown nothing. The assertion cannot be satisfied by a lucky corpus, and it fails the day the property stops being true.

**It asks the JSON API rather than counting `<article` in the page, and that is load-bearing.** A Next.js document carries the page twice — the DOM HTML and the RSC flight payload, JSON-escaped in a script tag — so counting elements in HTML counts **both copies**, and two equally doubled numbers compare perfectly well while measuring something else. That hazard cost four CI rounds on the empty-state check in this very PR and is ticketed as #115; this check is written not to need that sweep.

**The floor runs before the assertion and names which thing is broken.** Two empty sets are not "different", and `0 < 0` is false, so an empty corpus would fail the comparison and read exactly like a broken filter. It fails separately with *"this is the SEED, not the filter"*, and the assertion's own error names both possibilities and points at the content types it just printed. That is the repair from this PR's reading-time precondition applied one step earlier — two scripts had confidently reported a preference broken when the preference was fine and the corpus was too short.

## The pictures

**The two positions, at the same moment, on the same phone.**

| Built — what the app opens on | Everything — one tap wider |
|---|---|
| Today on the Built position: 4 stories, 4 minutes, the Built button selected, and a first card badged RELEASE | Today on the Everything position: 7 stories, 7 minutes, 6 unread, the Everything button selected, and a second card badged DISCUSSION |
| **4 stories · 4 minutes** | **7 stories · 7 minutes · 6 unread** |

**The count is not the only thing that changes, and the second row of each picture is why this pair is the evidence.** Card 02 is a **RELEASE** on Built and a **DISCUSSION** on Everything. A reader does not just see a longer list, they see a *different kind of thing* appear — which is the actual promise of the control, and the thing a number alone would not show.

The same pair on a laptop, where the control and the count sit in one frame:

| Built | Everything |
|---|---|
| Today on Built at laptop width, showing the sidebar with four surfaces and 4 stories | Today on Everything at laptop width, showing 7 stories |

**The quiet day — the state the owner will actually hit, and the one most likely to be mistaken for a breakage.**

Today with an empty brief: Nothing built today, explaining that the database answered and suggesting switching to Everything

It says **"The database answered, so this is a quiet day rather than a fault"** and then tells the reader where to go: *switch to Everything above*. The control stays on screen. **It never says it cannot reach the stories** — that is a different state with different words, and keeping those two apart is what this project exists to get right.

## Two different builds, and this section is here so the pictures do not imply otherwise

**The pair is a fixture build. The quiet shot is live data.** Both facts matter and neither is visible in a screenshot, so they are written here rather than left to be assumed.

**The live development database cannot produce the pair at all.** All five of its stories are `RELEASE`, so both positions return the same set, and **the script refused to write the files** — which is the refusal doing its job rather than a failure. The cause is the **catalogue, not the filter**: `seed-e2e-stories` deals its items round-robin across sources that have a usable feed URL, and this database has effectively one, so every seeded story inherits that source's `RELEASE` default. I re-seeded the catalogue and re-ran the pipeline before concluding that.

**The quiet shot is the opposite way round**, because its subject *is* the brief window, and in a fixture build the window comes from the fixture store rather than from preferences. It was produced the way a reader produces it: the brief time moved forward through the real preferences API so the window genuinely opened a minute ago and genuinely had nothing in it, then restored to `07:30` in a `finally`. **Not planted** — a state reached by stubbing the query would only prove the stub works.

## What running the script taught the script

**It died the first time, and the way it died is the point.** The quiet section called `.textContent()` on an element that was not on the page, so Playwright waited thirty seconds and threw. The run ended on a `TimeoutError` naming a locator, **with the floor never printed and four good screenshots already on disk** — a passing result that looked like a crash. That is precisely what `scripts/lib/floor.mjs` exists to prevent, committed hours before, by me. It now counts before it reads and fails as a floor that names the fixture-mode cause.

**The sections also run independently now** (`--only=pair|quiet`), because they need different builds. On live data the run bailed at the pair's floor and never reached the quiet section at all, so the flag is structural rather than a convenience.

## Looked at, not only captured

The summary line shows **unread** on Everything and not on Built, which looked wrong in the picture. It is correct: `brief-summary.ts:36` hides that segment when *every* story is unread, which is true of Built's four and not of Everything's seven. **Checked the rule before filing it** — the finding would have been false.

## Rebased onto main after #110 and #105 merged, and what the merge decided

**The one real collision was `src/api/brief.ts`, and #105's own docblock settled it before I got there.** It says the options object exists because two lanes were adding an axis from different bases, and that *"a third axis is another key here, not another argument."* `types` arrived on this branch as a third **positional** parameter, so it is now a **key on `BriefOptions`** and both callers pass `{ types: typesForView(view) }`.

**The two axes are orthogonal and both apply.** `includeAdjacent` asks whether a story is AI-related at all; `types` asks which *kind* of thing it is. They narrow different columns and neither substitutes for the other, so the query carries both conditions rather than choosing. That is written into the interface rather than left for the next reader to re-derive.

**The migration journal merged to `idx` 0–6 with no gap and `when` strictly increasing.** The gap I had deliberately left at `idx 4` closed by itself when #105's `0004_freezing_the_phantom` arrived — which is what it was reserving. My `0005` and `0006` sit at `…987195` and `…988195`, both above #105's `…986195`, so they remain eligible against a database that has already applied `0004`. **A migration whose `when` falls below the applied maximum does not run late — it never runs at all**, and cannot be made to by any later `migrate`.

**One thing the rebase would have broken silently, fixed deliberately:** my `0005` snapshot and #105's `0004` snapshot both recorded the same `prevId`, because both were generated from the same parent. Left alone, a future `drizzle-kit generate` would diff against the wrong base. `0005` is now chained onto `0004`'s id, and `0006` already chained correctly onto `0005`.

**And the semantic conflict worth naming, because it auto-merged cleanly:** `docs/architecture.md:198` is the line I corrected from *"Three surfaces are still placeholders"*. Git saw no overlap, so nothing would have warned me if the stale version had won. It did not — the corrected line is on this branch. On **#65** the original line is still there and **that is correct**, because #65 does not remove those two routes; the line only becomes false once this PR lands.

**The screenshot URLs are now anchored by a pushed tag, `evidence/108-filter-shots`.** They have been re-pinned three times, once per rebase, and each time the previous commit became unreachable while still returning HTTP 200 — a body that looks perfect and rots later, for someone else. A tag is a ref on the remote, so the commit stays in **every reader's** ref set rather than only in the clone of the person who created the danger. That is the actual fix; re-pinning was treating the symptom.

## Known stale, named rather than left

- **Seven shots in `docs/screenshots/shell/`** show a six-tab bottom bar. Their subjects still exist, so they are *wrong* rather than *dead*, and they need a browser. Regenerated in the same block as the four above. Nothing in the repository references that directory, which is why they are last in the queue and not first.
- **Two shots that photographed deleted pages** — `research-phone-dark.png` and `releases-phone-light.png` — are removed here, along with `src/app/{research,releases}/loading.tsx`. Those two files outlived their own `page.tsx`: a segment holding only a `loading.tsx` is not a route, so both were dead, and one was still announcing **"Loading Research"** to a screen reader for a page that now 404s.
- **Three comments counted surfaces that no longer exist.** `nav.ts` still opened *"The six surfaces… then the two filtered views"* with my explanation of this PR appended **underneath** the sentence it contradicts; `BottomNav` claimed it carries *"ALL SIX"*, where the number is load-bearing because the point is that the bar is complete; `architecture.md` still listed *"Three surfaces are still placeholders. Radar, Research and Releases"*. All corrected. **Both comments rotted because they contained a count**, so `nav.ts` no longer counts and `BottomNav` now claims it **is** `navItems` rather than a filtered copy of it — which stays true the next time that list changes.

## Also in here

- `BriefResponse` carries the `view` that produced it, so a client can tell which set it is looking at.
- Fixtures filter by view too — and **before** the budget, exactly as the server does — or fixture mode would stop being a way to see the product work.
- `EmptyState` gained two data attributes so an assertion can ask whether the reader was told anything without having an opinion about the words.


zarghammir:

Reviewed at `4f5d097`, read-only, in my own worktree. Three real checks green — lint/typecheck/test/build, compose-smoke, **browser checks** — with Vercel red on the account quota, which I verified independently earlier and am not re-litigating.

**Held to its parent, not to main.** `feat/today-live` is `a4d6063`, which is **#65's head** — so this is stacked on #65, not on a separate branch, and the merge chain is **#105 → #65 → #108**. Its `MERGEABLE` says it can merge into `feat/today-live`, a branch that is itself unmerged and 18 commits ahead of `d8ec412`. Its +512/−92 is measured against that parent. Nothing here is a statement about main.

**Two blocking items.** One is a control that stopped detecting what it exists to detect — introduced by the commit that fixes controls that stopped detecting. The other is the thing you asked me to judge rather than wave through.

---

# ① `verify-saved-settings.mjs` pins the view on one side of the comparison and not the other

You asked me to judge whether the `view=all` change is a test regaining its subject or a test being made to pass. **For `verify-today.mjs` it is the former, cleanly. For `verify-saved-settings.mjs` it is neither — it is half-applied, and that is worse than either, because the file now carries a comment asserting the axes are separated while the assertion that depends on it is not.**

`scripts/verify-saved-settings.mjs:282-291`:

```js
await page.goto(base + "/", { waitUntil: "networkidle" });              // <- NO view
const atPreference = await page.locator("article h2 a").count();
await page.goto(base + "/?length=all&view=all", { waitUntil: "networkidle" });
const atAll = await page.locator("article h2 a").count();
...
if (!(atPreference < atAll)) floor.push("... the preference is not reaching the server");
```

`DEFAULT_VIEW` is `"built"` (`brief-length.ts:69`) and nothing in the harness stores a view, so **the bare `/` renders the built view**. The two sides of the comparison now differ on **two axes**: `atPreference` is *(saved 5-minute length × built)* and `atAll` is *(all length × all)*.

**The fixtures make the failure concrete.** Seven stories — PAPER 1, RELEASE 2, RESEARCH 1 on the built side (**4**), plus NEWS 1, DISCUSSION 1, BUSINESS 1 (**7** total). So:

| | stories |
| --- | --- |
| `/` — built view, **length preference ignored entirely** | **4** |
| `/?length=all&view=all` | **7** |
| `atPreference < atAll` | **4 < 7 → PASSES** |

**The assertion's own docblock states the reasoning it no longer has:** *"Five minutes is provably shorter than everything, so a Today that ignored the preference would show the same count for both."* That held when the two sides differed only in length. It does not hold now — a Today that ignored the preference shows **4 and 7**, and the floor stays quiet.

**And this is the file whose new header says it holds the view still:** *"The two are separate axes: this file measures LENGTH, so it holds the view still."* It holds the view still on one side.

**Fix:** `await page.goto(base + "/?view=all", ...)` at line 282 — pins the view without supplying a length, so the saved length preference is still the thing under test and the comparison is single-axis again. The two `?length=all&view=all` navigations below it are already right.

`verify-saved-settings.mjs:107`'s bare `/` is the first-run redirect check and involves no count, so it is correctly left alone. **`verify-today.mjs` is fully pinned** — every one of its navigations carries `view=all`, and there it genuinely is a test regaining its subject: leaving those on the default view would have made them measure length × content-type while claiming to measure length.

---

# ② There are no pictures, and there will be no preview for 24 hours

You asked me to say so if the body's screenshots are not sufficient. **There are none.** Zero images in the body, zero `.png` added by the PR — I checked for markdown images, HTML ``, `user-images` and `githubusercontent` links, and for added image files. Nothing.

This PR **deletes two navigation destinations, removes two tabs from the bottom bar, and introduces the app's single most visible control** — and with Vercel rate-limited there is no preview deployment either. The owner reads this project through pictures, and tonight there is no picture and no preview. **A written description of a two-position filter is not a substitute for seeing which position the app opens on.**

The minimum that would close it: the two positions of the control at phone width, and the built-view empty state showing that it tells the reader where to go. The lane already has the machine and already drove it at 160px and 390px for the floor — these are the same run.

---

# The 160px floor: observed, dated, and honest about what would retire it

**This is the part of the PR I would keep unchanged.** The floor is no longer derived:

> *"OBSERVED FAILING AT FOUR TABS on 2026-09-17, at this head: 160px produced a narrowest tab of 40px — the figure the arithmetic above predicts — and 'Settings' clipped in 10 of 10 bars, with the floor still passing, so the red was the layout rather than an empty instrument."*

Three things make that a measurement rather than a claim: the observed narrowest tab **matches the arithmetic** (4 tabs × ~40px at 160px), the clip count is **10 of 10** rather than one marginal case, and **the instrument's own floor still passed** — so the red was the layout and not an empty sweep. That last clause is the one that distinguishes a control going red from a control going missing, and it is the check I had to add to my own comment-strip instrument on #65 after it compared two empty files and reported "identical".

**And the re-earn condition is stated rather than implied:** *"Add a fifth tab, or a word longer than 'Settings', and this figure is a prediction again — at which point this paragraph is evidence about a tree that no longer exists and should be re-earned rather than trusted."* That is the dated-observation discipline with its expiry condition made explicit, which is strictly better than a date alone.

**`EXPECTED_TABS = 4` is asserted rather than derived, with the right reason:** *"a count read from the same config the bar renders from would agree with it however wrong both were."* That is the registry-test lesson — an instrument that reads its expectation from its subject cannot see the subject being wrong. Changing the population and the expectation in the same commit and then proving they still agree is the correct order.

**The diagnosis of the original red is right, and the important half is that the page was never broken.** The assertion grepped for the literal `"quiet morning"` and #102's copy made the default view say `"quiet day"`. All three clauses are now genuinely asserted with `exit 1`, including the third — I checked it is a real `if grep -qi ...; then exit 1; fi` and not a comment.

---

# Non-blocking

**Clause two still pins a wording, and its comment claims otherwise.** The check greps `quiet (morning|day)` while the comment says *"Both views must say the day was quiet, in whichever words suit them."* The regex permits exactly the two spellings that exist today — `page.tsx:128` and `:129`, inline template literals with no shared constant. A third view, or an edit to "a calm day", breaks it again **on a correct page**, which is the same failure being fixed here. Clauses one (the marker) and three (the negative) are meaning assertions and will survive any copy; clause two is a wording pin in a wider regex. The comment should claim what the pattern enforces, or the strings should be exported from one place and the check derived from them.

**`nav.ts`'s docblock survived the deletion it describes.** Directly above the new comment explaining the removal:

> *"The **six** surfaces. Order is the reading order: the brief first, then everything arriving, **then the two filtered views**, then your own pile, then configuration."*

The list below it has four entries and no filtered views. `EXPECTED_TABS` went 6 → 4, so the instruments know; only the prose still says six. A comment is a claim, and this one is false immediately above a comment added to explain why.

**`ViewMemory` records a default as though it were a decision.** It writes the cookie on every render from `current`, which is the *resolved* view — so a reader who has never touched the control gets `ai-radar-view=built` written on their first visit. The cookie afterwards cannot distinguish *"this reader chose Built"* from *"this reader never expressed a view and Built was the default when they first arrived."* If `DEFAULT_VIEW` ever changes, everyone who merely visited is pinned to the old one. That is the same absent-versus-answered distinction this stack gets right twice elsewhere — `parseViewOrThrow` keeping absent apart from unrecognised, and `matchedAiVocabulary` keeping NULL apart from false. Writing only when a `?view=` was actually present would preserve it, and the shared-link case the comment cares about still carries `?view=`.

**Related and deliberate, but worth the owner knowing:** because the cookie is written on arrival, **opening someone else's shared `?view=all` link permanently changes the recipient's own default.** The comment argues for this and it is a defensible ruling; it is also the kind of thing that is easier to agree to before it happens than after.

---

# What else is right

**My carried-in item is discharged, and pinned the way I wanted.** `views.test.ts:85` asserts `parseViewOrThrow(null, "all") === "all"` **and** `parseViewOrThrow(null, "built") === "built"` — two different fallbacks, so the test cannot pass against a hard-coded constant. `:75` asserts unrecognised throws. The distinction is checked in the tests and not only in the parser.

**`VIEW_OF` is a total `Record<ContentType, BriefView>`**, so a new content type is a compile error rather than a story that belongs to neither view and appears in nothing. Pinned, not asserted.

**`views.test.ts:23` — "RETURNS DIFFERENT SETS — the property, not the instance"** — asserts `built ≠ all` and `all.length > built.length` rather than listing today's membership. A control with two positions that return the same rows is the defect; this tests for it directly instead of enumerating.

**The route deletion is clean.** I swept the tree for `/research` and `/releases`: the only hits are external feed URLs in `seed-data.ts`. No dangling link, and the a11y sweep's `ROUTES` was updated with the reason — a deleted route would 404 and the landing-path floor would report it as landing elsewhere, *"which is true but unhelpful."*

**The built-view empty state tells the reader where to go** — *"switch to Everything above to see the news and discussion around it"* — so a quiet default view is not a dead end. That is the gated-control-stays-visible rule applied to an empty state.

---

## To clear this

1. `verify-saved-settings.mjs:282` — `base + "/?view=all"`, so the length comparison is single-axis again.
2. Screenshots in the body: the two positions at phone width, and the built-view empty state. Same browser run as the floor.

Nothing merged, nothing pushed, no branch touched; `git status` clean. The box was not taken — reading only. **Merge order unchanged: #105, then #65, then this.**

zarghammir:

Reviewed at `33f0805` (the push landed mid-review), read-only, in my own worktree. **Based on `main`, 0 behind** — `MERGEABLE` finally means what it says. All five real checks `COMPLETED/SUCCESS`, polled to conclusion; Vercel is the account quota.

**Everything in the brief checks out, the acceptance is now measured rather than photographed, and I am blocking on one paragraph** — a falsified claim that was corrected in a commit message and left standing in the source comment, which is the copy that outlives it.

---

# ① The blocker: the correction landed in the commit and not in the file

`34814dd` says it plainly, and says it well:

> *"'A bare boolean data attribute never reaches the server-rendered HTML.' **It does.** React renders `data-empty-body` as `data-empty-body="true"` either way, so that commit changed the output not at all. Its reasoning was false and its evidence — screen-state present, empty-body absent — **had a different cause than the one I assigned it**."*

**The PR body is clean** — I searched it for the falsified mechanism and it is not there.

**But `src/components/page-shell.tsx:74-86` still states it as fact:**

> *"Written as `data-empty-body`, JSX passes the boolean `true`, and **the attribute was NOT in the server-rendered HTML**: CI curled a correctly quiet page and found the body in zero characters."*
>
> *"**The evidence is a controlled comparison** inside that one failing run: the same HTML carried `data-screen-state="quiet"` — a STRING value, three lines up in this file — while the bare boolean was absent."*

That is the exact mechanism `34814dd` retracts, **plus the evidence `34814dd` says was misattributed, presented as proof.** It also draws a conclusion from it — *"Every other data attribute in this codebase is written with a value; these two were the only exceptions, which is also why nothing else was affected"* — which rests entirely on the retracted cause.

**Why this is the blocking copy rather than the venial one.** A commit message is read once, by whoever is looking at that commit. A docblock above an exported component is read by everyone who touches it, forever, and this one is framed as an evidenced finding rather than a guess. The next person to meet a bare boolean data attribute in this codebase will read it and believe React drops it — and either "fix" working markup or distrust a correct pattern. **It is the same defect this PR spent three CI rounds and two wrong hypotheses discovering, preserved in the place most likely to propagate it.**

**The fix is already written** — `34814dd`'s body is better prose than anything I would suggest. It needs to move into the file, keeping the real reason for the explicit values, which the follow-up also states: they match `data-screen-state` three lines up, and the literal `="true"` is what disambiguates the two copies. **That reason stands on its own and does not need the false one.**

Nothing else is outstanding.

---

# ② The RSC sweep — I tested the claim rather than repeated it, and it is narrower than stated

**The claim** — *"this applies to EVERY curl-and-grep assertion against this app"* — is true as a warning and **over-broad as a rule**, and the over-broad version is the one that gets ignored, because a reader who checks one, finds it safe, and stops checking is worse off than one who was told what to look for.

**The complete population, swept rather than sampled.** Every raw-HTML assertion in the repository:

| site | reads | exposed? |
| --- | --- | --- |
| `ci.yml:210` | `-o /dev/null -w '%{http_code}'` — status only | no |
| `ci.yml:618`, `:657` (compose-smoke) | status only | no |
| `ci.yml:232` | **the body** — the empty-state step | the only candidate |
| every `scripts/*.mjs` | Playwright **DOM locators** | **structurally immune** |

`grep -rn "page.content()\|innerHTML\|outerHTML" scripts/*.mjs` returns **nothing**. That is the important half and it is not in the comment: **a DOM locator queries the parsed document and can never see the flight payload.** The exposure is confined to curl-and-grep by construction, and there is exactly **one** such assertion in the tree.

**Inside that one step, the four patterns, classified:**

1. `grep -o 'data-screen-state="[a-z]*"' | head -1` — **`head -1` over a loose pattern, the exact dangerous shape** — and **immune**, because `page-shell.tsx:35` renders `data-screen-state={state}` and the payload carries `data-screen-state\":\"quiet\"`. The pattern requires a literal `="`, which the escaped form never contains. Safe *by the same mechanism the fix relies on*.
2. `grep -o 'data-empty-body="true"…'` — the repaired one, same immunity. ✓
3. `grep -o 'data-empty'` in the diagnostics — inside the failure branch, not an assertion.
4. `grep -qi "cannot reach your stories"` — plain prose, **matches both copies** — but it is a **negative** assertion, so matching either copy can only make it **over**-fire. A false red is loud; it cannot produce a silent pass.

**So the rule worth writing down is two conditions, not one:** the hazard is `head -1` over a pattern that **can match the escaped form** (anything containing `="` cannot), **in a positive assertion** (a negative one only over-fires). Both must hold. Stated that way it is checkable in seconds per site; stated as "every curl-and-grep" it is a caution, and #115 will re-derive this on its first case.

---

# ③ `shoot-filter.mjs` as a capture — the reasoning holds, and the escape hatch worked as designed

*"Those assertions guard the PICTURES, not the product"* is the right test and it is correctly applied: a capture script passing means files were written, and nothing in this job reads them, so its green would assert nothing a reader could act on.

**The second half — *"running it here would make CI depend on a corpus that differs by view, which is a property of the seed rather than of the app"* — is true, and it is an argument for a floor rather than against a check.** That is precisely what #65 did for the reading budget: a corpus property its consumers depend on, asserted in the producer, with *"Add items to ITEMS above; do not lower this floor."* The lane has now built the view equivalent — see below — so the objection is answered rather than merely accepted.

**And this is the first use of the escape hatch, working as intended.** #112's floor caught the file driving a browser unclassified; its comment had named *this file and this moment* in advance; and the resolution was **one line and a decision** — `NOT_A_CHECK="scripts/shoot-filter.mjs"` — not a panic edit. `6 found` against `5 derived` in the log is the floor reporting that all six browser-driving scripts are accounted for. **Predicted, caught, and closed by deciding rather than forgetting.**

---

# ④ The view-split: my ruling is moot because it is now measured, and the way it got measured is better than what either of us asked for

**The acceptance, from this run:**

```
MEASURED: view=built -> 5   [RELEASE,RESEARCH]
MEASURED: view=all   -> 11  [BUSINESS,NEWS,REGULATION,RELEASE,RESEARCH]
RESULT: the two positions return different sets (5 built of 11 total).
```

**The types agree with `VIEW_OF` independently.** RELEASE and RESEARCH are built-side; BUSINESS, NEWS and REGULATION are the three that only "all" admits. So the mapping and the real rows now corroborate each other, and printing the type lists is what makes that visible rather than taking `5 < 11` on trust.

**I verified the implementation rather than the intention:**
- **It reads JSON.** `curl /api/brief?view=…` → `JSON.parse` → `b.count`, `b.stories.map(x => x.contentType)`. `json()` in `http.ts` is `JSON.stringify(data)` with no envelope, `count: stories.length` is a real field, and `StoryCard.contentType` exists at `stories.ts:43`. **No `<article` counting, so the double-copy cannot reach it.** Written not to need the sweep, which is #115 applied forward.
- **The floor is genuinely first and cannot be satisfied by the assertion's condition.** `built < 1 || all < 1` runs before `built >= all`, and that ordering is load-bearing: an empty corpus gives `0 >= 0`, which is true, and would read as a broken filter. The floor intercepts it and says *"this is the SEED, not the filter"*. The assertion's own message then names **both** candidate causes rather than blaming the filter — which is the direct repair for the shape that cost two rounds on the reading-time precondition, where two files confidently blamed a preference.
- `length=all` is in both URLs, so the reading-time budget cannot truncate either side and turn a filter question into a budget question.

**My ruling, for the record, since you asked for it:** I agreed with yours — not a blocker for the merge, a blocker for calling the acceptance satisfied. It is now neither, because the property is asserted on every run.

**And on the box being the wrong instrument, the lane's argument is stronger than it stated.** It is not only that CI runs every time. **The borrowed box held the one corpus already observed to be unable to express the property** — five stories, every one a `RELEASE`, the pair refusing to be photographed. Measuring there would have produced either a red that meant nothing about the filter, or a green obtained by hand-fixing the corpus first. **An instrument known to be blind to the property is not a weaker instrument, it is the wrong one**, and arguing your own offer down to that is the harder call.

**If it had landed red: seeder, and I can support that with numbers rather than instinct.** `seed-data.ts` declares **11 NEWS** sources against **7 built-side** (4 RESEARCH, 2 RELEASE, 1 PAPER), so the CI catalogue can express the split — while a dev database with effectively one usable feed inherits that one source's default for every story and cannot. The dev-box failure was diagnosed as *"the catalogue, not the filter"*; this run confirms that diagnosis **in a different environment**, which is what turns it from an explanation into a finding.

---

# The two numbers, checked as arithmetic rather than read

`statesAudited` **28 → 20** and `tabsPerBar` **6 → 4**, with `expectedBarStates` **14 → 10**. Derived from the tree rather than trusted:

- `ROUTES` is now 5 (`/`, `/radar`, `/saved`, `/settings`, `/welcome`) × 2 sizes × 2 themes = **20** ✓ — and the 28 before was 7 × 2 × 2, so the delta is exactly the two removed routes (2 × 2 × 2 = 8).
- `nav.ts` has **4** `href` entries ✓, and `EXPECTED_TABS = 4`.
- Bar states are phone-only: 5 routes × 2 themes = **10** ✓.

**Every number moves by exactly what removing two destinations predicts, and nothing else moves.** That is the check worth doing — a count that fell by the right amount for the wrong reason is the failure this rules out, and an instrument with no stake in the PR's subject is the one that can rule it out.

---

# Also verified

`architecture.md:198` — the *"Three surfaces are still placeholders"* line **is corrected on this branch**, and correctly still stands on #65, because #65 does not remove those routes. Git saw no overlap, so nothing would have warned anyone if the stale version had won; checking the merged file's claim against the tree is what caught it in both directions.

The full artefact set is present and complete: 5 derived, **6 accounted for**, three control RESULTs each reddening under its own control, two NOTEs, five MEASURED digests, plus *"the empty state explained itself in 224 characters"* — the RSC repair working on a real page.

---

## To clear this

`src/components/page-shell.tsx:74-86` — replace the retracted mechanism and its misattributed evidence with what `34814dd` already says, keeping the real reason for the explicit values. One paragraph, and the text exists.

zarghammir:
REVIEW VERDICT: APPROVE

Confirmation re-read at `3a6dd6d`, read-only, in my own worktree. One file, +26/−12 since `33f0805`. **0 behind `main`, 17 ahead.** All five real checks `COMPLETED/SUCCESS`, polled to conclusion; Vercel red on the account quota — target is `?upgradeToPro=build-rate-limit`, not a deployment.

---

# The sweep: I grepped the tree, not the file

Four patterns across every `.ts`, `.tsx`, `.mjs`, `.md` and `.yml` in the repository:

| pattern | occurrences |
| --- | --- |
| `"was NOT in the server-rendered"` | **0** |
| `"not in the server-rendered"` | **0** |
| `"controlled comparison"` | **0, anywhere in the tree** |
| `"never reache…"` | 8, of which **exactly one** is in `page-shell.tsx` — line 90, *"claimed that a bare `data-empty-body` never reaches the server-rendered HTML"*, inside the retraction, naming what it retracts |

The other seven are unrelated — floor messages in `verify-saved-settings.mjs` and `verify-reader-isolation.mjs`, comments in four test files, and `shoot-filter.mjs`'s note about its own bail. **"Swept the tree" holds.** The misattributed evidence is gone outright rather than relocated, which is the stronger outcome: there is no copy left to be quoted.

---

# Over-correction: it introduces new claims and takes them out of the load-bearing position

That was the right thing to ask, and the answer is not "no" — **it is that nothing now depends on them.**

**The two reasons lead, and are explicitly not about React:** *"for two reasons that are both about the reader and neither about React"* — it matches `data-screen-state` and every other data attribute here, and the literal `="true"` distinguishes the attribute from its escaped twin. **Both reasons hold whatever React does.**

**The retraction then asserts a mechanism, because it has to:** *"React renders the bare form as `data-empty-body="true"` exactly as the explicit one."* That is a new standing claim, and it is **true** — React special-cases `data-*` and `aria-*`, stringifying the value instead of applying boolean-attribute rules, so `data-empty-body={true}` and `data-empty-body="true"` serialise identically. But it now sits as a **footnote explaining a withdrawn claim, not as a reason for anything.** If it were wrong tomorrow, no line of this file would change.

**That is exactly the inversion from before.** The old claim was load-bearing — it was offered as *the reason* the attributes carry explicit values. The new one is not. **A retraction that explains a mechanism cannot avoid asserting something; what it can do is make sure nothing rests on it**, and this one does.

**The same test applied to the other new claim, `"the payload copy comes FIRST"`:** also explanatory, also not load-bearing — requiring `="true"` matches only the DOM copy **regardless of order**, so the check survives that ordering claim being wrong. Two new sentences, neither of which can take anything down with it.

---

# The acceptance still measures at this head

```
MEASURED: view=built -> 5   [RELEASE,RESEARCH]
MEASURED: view=all   -> 11  [BUSINESS,NEWS,REGULATION,RELEASE,RESEARCH]
RESULT: the two positions return different sets (5 built of 11 total).
```

Byte-identical to the previous head, which is what a docblock-only change should produce and is worth confirming rather than assuming — a comment edit inside a component that the empty-state assertion reads through is not obviously inert.

**And the full artefact set is present:** 5 derived, **6 accounted for**, three control RESULTs each reddening under its own control, two NOTEs, five MEASURED digests, `statesAudited=20`, `tabsPerBar[0]=4`, and *"the empty state explained itself in 224 characters"*.

---

# Three formulations I am taking rather than improving on

**"A bare wrong sentence invites doubt; a wrong sentence wearing its evidence invites belief."** That is a better statement of my point than mine was. I said a docblock is more *visible* than a commit message; the real distinction is that the sentence had *"the evidence is a controlled comparison inside that one failing run"* sitting on top of a cause already withdrawn. Visibility made it worse; **borrowed credibility made it dangerous**, and those are different failures.

**A rule applied in one direction only.** Using the commit-versus-file distinction hours earlier to decide *not* to amend a commit, and never turning it around to ask what the file still said, is the sharpest self-observation of the night. **A directional rule is half a rule, and the unexercised direction is the one that fails silently** — nothing goes wrong, so nothing prompts the question. The tell is uncomfortable: **having just used a rule is what makes it feel handled.** It is the same shape as tonight's recurring absence-equals-pass, applied to one's own reasoning rather than to an instrument.

**"A class is a caution; two conditions and a population is a check."** Better than my version. I gave the two conditions and the population; that sentence says why they were the thing worth giving, and it generalises past #115 to every "this applies to everything of kind X" warning any of us writes.

---

**Approved. The merge click is the owner's, and nothing here needs waking him for.**

Issue #106, opened 2026-09-17
check-route-cost.ts: absorb #103's guard assertion, and floor the walk's COVERAGE rather than its output


zarghammir:
**Follow-up to #103, owned by the Intelligence lane, because `scripts/check-route-cost.ts` is theirs.** Filed so that neither half depends on anyone remembering — a note in a PR body and a lane's verbal yes are not an owner.

## Part 1 — fold the guard assertion into the instrument that already knows the boundary

#103 establishes the rule: **`internal/` means operator-only and secret-guarded; everything outside it is reader-facing and must not mutate operator state.**

`check-route-cost.ts:13` already encodes that same seam for a different concern — *"The public/internal boundary is the `internal/` path segment, which is a rule rather than a list"*. So the boundary is already walked, already classified and already machine-checked; the guard rule is a second assertion over a traversal that exists.

**#103 ships its own test rather than editing this file**, deliberately — Platform does not edit another lane's script on that lane's invitation alone, and I would rather the rule land WITH its enforcement than land correct-but-unenforced for a day. **That test is explicitly temporary.** It duplicates a traversal this script already performs, and this ticket is the instruction to fold it in and **delete the duplicate**.

**Why the duplication was acceptable for a day and is not acceptable indefinitely:** two instruments for one rule is how they drift, and the drifted one is the one nobody watches. What made it tolerable is that #103's version is a **test** rather than a second implementation — if it drifts it goes **red**, which is not true of a second copy of a check. That property is what buys the day; it does not buy a month.

## Part 2 — the floor counts what the walk FOUND, not what EXISTS

**This is the more valuable half and it is a live defect in the instrument, not a tidying job.**

`MINIMUM_ROUTES` is a lower bound on the search's **output**. It cannot see a route the walk never reached, because a missing route simply makes the loop shorter — **exactly the registry-test-cannot-see-a-missing-entry shape, sitting inside a floor built to prevent it.**

The concrete gap, measured by Platform across all of `src/app`:

- `all = walk(ROUTE_ROOT)`, so a route containing `internal/` **cannot** sit outside the walked directory and be wrongly exempted. That hazard does not exist.
- **But Next allows a handler anywhere under `src/app`.** `src/app/health/route.ts` would be a real, live route that the walk never sees — **not exempted from both rules, UNEXAMINED by both.**
- **Current count: 13 `route.ts` files, all inside `src/app/api`, 2 internal and 11 public, zero outside.** The gap is real and currently empty.

[...]

Issue #104, opened 2026-09-17
A pass over zero ENABLED sources reports success — "not seeded" and "all switched off" are the same exit code


zarghammir:
**Surfaced by #103 and filed separately at the PM's instruction, so it does not disappear because the auth fix landed.**

## An ingestion pass over zero enabled sources reports success

`src/worker/report.ts`:

```ts
export function exitCodeFor(result: IngestResult): number {
  if (result.bySource.length === 0) return 0;
  return result.bySource.every((r) => r.error) ? 1 : 0;
}
```

**This is deliberate, and its docblock says why:** *"No sources at all is a seeding question, not a failed run."*

**That reasoning is correct for the cause it names and blind to the one that now exists.** `bySource` is empty when no source is *enabled* — which happens for two unrelated reasons:

| Cause | What it means | Right response |
| --- | --- | --- |
| The catalogue was never seeded | Not set up yet | Green. Nothing is broken |
| Every source has been disabled | **Ingestion has stopped** | **Red** |

**The code cannot tell them apart, so it picks the reassuring one.** That is the absence-equals-failure shape at the operational level: *"collected nothing"* and *"had nothing to collect"* are the same observation.

## Why it matters beyond the tidiness of an exit code

Until #103 the disabling did not need an operator — `PUT /api/sources/[key]` was unauthenticated and the keys are public. **With every source off, the scheduled worker still runs every thirty minutes, still exits 0, still prints a total line, and nothing reddens.** The brief quietly stops growing and the first signal is a reader noticing the news is stale.

**#86's health does not cover it.** It reports per-source failures. A disabled source does not fail — it is not run at all, so it produces no `ingest_runs` row and no `lastError`. A catalogue that is entirely switched off is, to the health signal, seventeen sources with nothing to report.

## The distinction is one query

The two causes are separable with information the database already has:

```
sources with rows at all   vs   sources with enabled = true
```

- rows = 0 → never seeded → green is right
- rows > 0 **and** enabled = 0 → **every source has been switched off** → this is a failure

## Acceptance

- **An ingestion pass with a seeded-but-fully-disabled catalogue exits non-zero** and says which of the two states it found. `exitCodeFor` currently takes only `IngestResult`, so this needs the enabled/total counts plumbed to it — that is the reason this is a ticket and not a one-line change.
- **A never-seeded database still exits 0**, so the fix does not turn a fresh install red. **The negative needs the positive beside it**, or a function that always fails would pass.
[...]

Issue #101, opened 2026-09-17
GET /api/sources serves arbitrary worker diagnostics unauthenticated — the boundary was drawn and lastError was on the wrong side


zarghammir:
**Raised by the Platform lane while finishing #100, and it is the durable question that #99 and #100 only treated an instance of. Ranked above #92.**

## The boundary was already drawn, reasoned about, and documented — and one field was on the wrong side of it

`src/api/catalogue.ts:55-61`, verbatim from main:

> *The catalogue as a reader sees it.*
>
> *`config` and the feed `url` are deliberately absent: they are operational settings, not display data, and **a feed URL is the one field here that could carry a credential in a query string**.*

**Someone thought about exactly this and drew a line: operational settings do not cross into what a reader sees.** Then `lastError` — operational diagnostics, written by the worker, never read by a reader — was returned anyway.

**And the comment's own claim was false.** *"The one field here that could carry a credential"* was not the only one. `lastError` could, and #99 measured that it did: `getaddrinfo ENOTFOUND <host>` on a connection failure, `role "<user>" does not exist` on an auth failure. **A correct boundary, correctly explained, with an incorrect census of what it had to cover.**

## Why #100 does not close this

**#100 stops a *credential fragment* being written into that field.** It does not change the fact that `GET /api/sources` serves **arbitrary stored error text, to anyone, with no authentication.**

The route's contract is effectively *"whatever the worker stored"*. **The next thing that writes to that field does not have to know it is publishing.** Today the text is a feed's 403 or a parse failure and is harmless. That is a property of today's writers, not of the design.

**This is the smaller, more durable question**: not *does today's text contain a credential*, but **should an unauthenticated endpoint serve operator diagnostics at all**. It is the one that stops this recurring.

## Options

1. **Drop `lastError` from the reader-facing response entirely.** It is diagnostics; `health` and `consecutiveFailures` (from #86) already give a reader everything actionable — *this source is failing* — without the text. **Simplest, and consistent with the boundary already documented three lines above.**
2. **Reduce it to a category** — `network`, `parse`, `auth`, `unknown` — derived at the write site. Keeps the signal, drops the free text, and cannot carry anything a writer did not intend.
3. **Keep it and authenticate the route.** Largest change, and turns a public catalogue into an authenticated one for the sake of one field.

**My inclination is 1 or 2, and the choice belongs with whoever owns what the Settings source list needs to show.**

## Acceptance

[...]

Issue #95, opened 2026-09-17
topicKeys is still one shared row — it feeds the ranker before any request, so it could not move with the rest


zarghammir:
**Split out of #94 by the UI lane, for a structural reason rather than an effort one — and it verified the reason rather than asserting it.**

## Why this one could not follow the others

`briefLength`, `theme` and `onboardedAt` moved to the device in #94 because nothing but the reader consumes them. **`topicKeys` is different: it feeds the RANKER, in the WORKER, ahead of any request.**

Verified in code: `ingest.ts` calls `rankAllStories`, which reads `prefs.topicKeys` and writes `stories.score`. The request path (`brief.ts`) then simply orders by `desc(stories.score)`.

**So the reader's topics are baked into a stored number before anyone opens a page.** Making them per-reader means **ranking at read time** — which is a pipeline change and a contract change to `/api/brief`, not an edit to one lane's files.

## Why it matters that it is still shared

Under #91's ruling — *the catalogue is shared because the news is the same for everyone; the reader's state is not* — `topicKeys` is plainly reader state. It is **what you follow**. Today one shared row decides it for every reader of a public URL, and there is no second person's copy to collide with because there is only ever one.

**It is the least visible of the shared fields and the most consequential**, because it does not merely mislabel a preference — it changes **what reaches the top of everyone's brief**.

## The shape of the work

Three honest options, and this needs deciding rather than defaulting:

1. **Rank at read time for the reader's topics.** Most correct, most expensive: the score can no longer be a stored column computed once per ingest.
2. **Store a topic-neutral base score, apply the reader's topic weighting at read time.** The ranker keeps its pre-computed work; the per-reader part is a cheap re-order. **Probably the right shape**, and it needs the ranking explanation (#15's provenance) to stay truthful about which half came from where.
3. **Accept that topics are instance-level** — a self-hoster's choice for their own deployment, like `briefTime` and `timezone` — and say so where the field is defined. Cheapest, and defensible for a self-hosted product; wrong the moment two people share a URL and expect different briefs.

## Acceptance

- Whichever option, the decision is **recorded where the field is defined**, not only here.
- If topics become per-reader: **two browsers following different topics get different top stories**, tested with two contexts, with a control that shares the topics and reddens the named test.
- **The ranking explanation must not claim a per-reader weighting was applied when it was not**, and vice versa — that surface is exactly where a half-truth becomes a visible lie.
[...]

PR #110, opened 2026-09-18
Move the source write under the guard that already exists (#103)

### LINKED ISSUE #103: Unauthenticated PUT /api/sources/[key] lets anyone disable every source, silently — the operator write surface has no guard
**Found by the Platform lane while scoping #101. I verified every claim below by reading the routes myself rather than relaying it. Ranked ABOVE #101, and above #92.**

## Anyone on the internet can turn this product off

```
PUT /api/sources/[key]    body { enabled: boolean }    NO AUTHENTICATION
```

`src/app/api/sources/[key]/route.ts` parses the body, calls `setSourceEnabled` and returns. **There is no secret check, no session, no cookie, and there is no `middleware.ts` anywhere in the tree.**

**The keys are not secret.** They are in `src/db/seed-data.ts`, in a repository the owner made public yesterday.

**And it fails silently, which is the part that makes it worse than it sounds.** Disable all seventeen sources and the worker still runs every thirty minutes, still exits 0, still reports a pass — because *a pass over zero enabled sources is not a failure*. Nothing reddens. The brief just quietly stops growing, and the first signal is a reader noticing the news is stale.

That is the [absence == failure](../../memory) shape at the operational level: **"collected nothing" and "had nothing to collect" are the same observation**, and the health signal from #86 reports per-source failures, not "the catalogue is empty."

## The survey — six write routes, zero auth references

Verified by reading every `route.ts` under `src/app/api`:

| Route | Verbs | Writes | Guarded |
|---|---|---|---|
| `sources/[key]` | PUT | **source enabled/disabled** | **no** |
| `preferences` | PUT | shared `topicKeys`, which feeds the ranker (#95) | **no** |
| `saved/[storyId]` | POST, DELETE | shared saves | **no** |
| `read/[storyId]` | POST | shared read state | **no** |
| `hide/[storyId]` | POST | shared hidden state | **no** |
| `internal/ingest` | POST | triggers ingestion | **yes** — secret checked inside `trigger.ts` |

**`internal/ingest` is the proof the project already has the concept.** One route is guarded, by a pattern that exists and works. The others were simply never brought under it.

## The ranking, and why this outranks #101

#101 is a stranger **reading** a hostname. This is a stranger **stopping the product**, silently, with a key published in the repo. **Disclosure versus denial of service, and the denial is the one with no detection path.**

`sources/[key]` also returns the full `SourceSummary` on success — **including `lastError`** — so it is a second route serving the text #101 is about. Fixing #101's read alone would leave the same string reachable through the write.

The reader-state rows (`saved`, `read`, `hide`, `preferences`) are a genuine issue but a *different* one, already being addressed: #65 moves reader state onto the device and #95 covers the shared `topicKeys` that remain. **Do not fold them in here** — this ticket is about the operator control surface.

## The ruling

**Move the source write under the guard that already exists rather than inventing authentication.**

1. `PUT /api/sources/[key]` moves to **`/api/internal/sources/[key]`** and checks the same secret `internal/ingest` checks, through the same code path in `trigger.ts` — not a second copy of the check.
2. **One stated rule, written down where the next route author will read it:** *`internal/` means operator-only and is secret-guarded; everything outside it is reader-facing and must not mutate operator state.* A rule beats an enumeration — #101 exists precisely because a list of excluded fields did not survive a new field.
3. Nothing in the app calls this route today — the Platform lane grepped `/api/sources`, `listSources` and `SourceSummary` across `src` and read `settings/page.tsx` and `client.ts` directly, finding no caller. **So this is a move, not a removal of capability**, and no UI has to change.

**Full authentication is still out of scope.** The app is single-user and self-hosted by design (`schema.ts:206`) and accounts are a different product. The secret guard is proportionate and it already exists.

## Acceptance, and the control

- **An unauthenticated `PUT` to the source write returns 401/404 and the row is unchanged.** Assert the row, not just the status — a route that 401s and writes anyway passes a status-only check.
- **The same request WITH the secret succeeds**, so the test proves the guard discriminates rather than that the route is broken. A negative assertion needs a positive beside it.
- **A control:** remove the guard and the named test must redden. If it stays green the test is not reaching the guard.
- **The silent-failure half, which is the reason this matters:** with every source disabled, an ingestion pass must **not** report success. That is a separate defect this ticket surfaces — file it if it is bigger than a line, but do not let it disappear because the auth fix landed.

Related: #101 (the read side, same field, ranked below this) · #86 (health, which does not cover an empty catalogue) · #95 (shared `topicKeys`) · #92 (log sweep, below both).

zarghammir:
Closes #103.

## This route was on the wrong side of a boundary the project already draws

**Not a new convention.** `scripts/check-route-cost.ts:13` already says it, and already enforces it by machine — for a different concern:

> *The public/internal boundary is the `internal/` path segment, which is a rule rather than a list.*

That script uses the `internal/` segment to decide which routes are exempt from the cost rule. **The authorisation question is answered at the same seam, by the same segment.** `PUT /api/sources/[key]` simply sat on the public side of a line the repository had already drawn.

**Anyone on the internet could switch off every news source, and nothing would have told you.**

**The control.** One request — `PUT /api/sources/verge-ai` with `{"enabled": false}` — turned a source off. **There was no password, no key, no check of any kind.** The source names are listed in `src/db/seed-data.ts`, in a repository that has been public since yesterday.

**Why nobody would have noticed.** With every source off, the collector still runs every thirty minutes, still finishes, still reports success — because *a run with no sources to collect from is not a failed run*. The brief just quietly stops growing. The first sign would be a reader noticing the news is stale. That detection gap is **#104**, filed separately.

**What changes.** The control moves to `/api/internal/sources/[key]` and requires the same secret that already protects the ingest trigger.

| --- | --- | --- |
| `PUT /api/sources/[key]` | **Gone, not redirected** | No shim. Nothing in this repository called it, and the app does not need it |
| The search for callers | **Covered this repository only** | `/api/sources`, `listSources` and `SourceSummary` grepped across `src`; `settings/page.tsx` and `lib/api/client.ts` read directly. **It cannot see a consumer calling the deployed URL from outside this repo** — the repo is public, so somebody may have one. Such a caller now gets 401 until it sends `x-internal-secret` |
| The silent failure that hid it | **Not fixed here** | #104. The auth fix stops the *remote* cause; #104 is why nobody would notice *any* other cause |
| Reader-state routes (`saved`, `read`, `hide`, `preferences`) | **Untouched** | A real but different question, already tracked: #65 moves reader state onto the device, #95 covers shared `topicKeys` |
| Full authentication | **Out of scope** | The app is single-user and self-hosted by design (`schema.ts:206`). Accounts are a different product; this is the guard that already exists, applied where it was missing |

## A rule, not a list

Written where the next route author will read it — `src/api/internal-guard.ts` and `docs/architecture.md`:

> **`internal/` means operator-only and is secret-guarded. Everything outside it is reader-facing and must not mutate operator state.**

**The doc paragraph this replaces was itself the defect.** It read *"All but one are the client API; the exception is `internal/ingest` … twelve and one"* — an enumeration, wrong the moment a second internal route exists. **#101 exists because a list of excluded fields did not survive a new field.** A list of guarded routes would not survive a new route.

"Operator state" is defined there too: anything that changes what the product *does* — which sources run, what gets ingested, how ranking behaves.

## One guard, two callers

`refuseUnlessInternal` is shared, and `handleIngestTrigger` now delegates to it rather than keeping its own copy. **Two copies of an authorisation check drift, and the one that drifts is the one nobody is looking at.** The guard runs *before* the body is parsed or the key is read, so an unauthorised caller costs nothing and leaves no trace.

**One deliberate behaviour change:** the 503 body was `"the internal ingest trigger is not configured"` and is now `` `${route} is not configured` ``, so both internal routes answer the same shape and name themselves. **`trigger.test.ts` asserts only the status, so nothing pinned it — checked before changing it, not after.**

## Measured, not described — and the base was refreshed first

**The four conclusions below were measured at `6aca5c5`.** Two commits have landed since, carrying three changes.

They were actually run on a local wip commit, `5413e68`, which was rebased away and **returns 422 from the GitHub API** — a reader on this page could not check it. `5413e68^{tree}` and `6aca5c5^{tree}` are both `c064681aa037775f4f93b22a7ea0daa9d26cdedb`, so the measurement was taken at exactly the tree `6aca5c5` names, and that is the SHA cited. **A date plus an unresolvable SHA is worse than a date alone**, because the SHA advertises checkability it does not have. Rather than re-run them on a shared machine to confirm arithmetic, here is what changed and whether each thing can move a control outcome — dated measurement plus a reasoned delta, which is this repo's new house style applied to its own results.

| Change since `6aca5c5` | Can it move a control outcome? |
| --- | --- |
| The `docs/` sweep | **No.** No test reads `docs/` |
| `describeError` extracted to a leaf | **No.** It changes what `internal-guard.ts` *imports*. Both controls are about reachability **of** the guard module *from* a route, which is the other direction |
| `internal.length >= 1` → `>= 2` | **No — and this is the one worth deriving, because it looks like it should.** |

**The derivation, because the floor lives in the exact file Control A asserts stays green and Control B reddens by name.** `internal` is computed at `internal-guard.test.ts:93` as a **path filter over route files** — `routes.filter(r => r.includes("src/app/api/internal/"))` — not from reachability. Control B removes an *import*; it does not delete a *file*. So the walk still finds two internal routes, `>= 2` still passes, and the only assertion that changes is `reaches(route, GUARD)` for the sources route.

**Control B's red set stays at three.** Had `internal` been derived from reachability, the raise would have added a fourth red and the number above would now be wrong.

**Verified against `d8ec412`**, which is `origin/main` exactly. The branch was cut eight commits earlier; it was rebased before any of this ran. **A green on a stale base is a statement about a tree nobody will merge into.**

| Block | Exit | Result |
| --- | --- | --- |
| Targeted (`src/api`, `src/worker`) | 0 | **102 passed across 7 files** |
| **Control A** — guard **call** removed, import kept | 1 | **2 red**, both behavioural, by name |
| **Control B** — import removed too | 1 | **3 red**, including `reaches the internal guard` |
| Full suite | 0 | **442 passed across 35 files, ZERO SKIPPED** |

### Control A's green is a result, not a caveat

**Removing the guard's CALL while leaving its import reddens the behavioural tests and leaves the reachability test passing.**

```
Control A   × refuses an unauthenticated write AND LEAVES THE ROW UNCHANGED
            × refuses a WRONG secret and leaves the row unchanged
              internal-guard.test.ts — GREEN

Control B   × src/app/api/internal/sources/[key]/route.ts reaches the internal guard
            × refuses an unauthenticated write AND LEAVES THE ROW UNCHANGED
            × refuses a WRONG secret and leaves the row unchanged
```

That green is the import-versus-call limitation **demonstrated in the same run that demonstrates the coverage.** It is in the file's docblock too, but a reader can now check it instead of believing it.

**So the two tests are complementary and it is now provable which covers what:** the reachability check covers *every* internal route automatically and shallowly; the behavioural tests cover *each* route deeply, through the database, asserting the row. Control B reddening both files is the pair working.

**Reproducing Control A:** it removes the guard's *call* and leaves the import, so `tsc --noEmit` would likely reject that tree on an unused import. **It is a vitest-only mutation.** That does not weaken it — the point is which tests redden, not whether the tree typechecks — but it is worth saying so nobody reads the control as broken.

### Which route is covered by what

| Route | Import check | Behavioural, route module | Behavioural, function |
| --- | --- | --- | --- |
| `internal/sources/[key]` | ✅ | ✅ 401 + **row unchanged**, no-secret and wrong-secret, plus the positive | — |
| `internal/ingest` | ✅ | ✅ 401 + **`ingest_runs` unchanged**, and 200 with a correct secret | ✅ `trigger.test.ts`: `runCount()===0` on **all five** refusal paths, and `===1` on success — so the counter is not inert |

**Stated as a composite rather than letting one test take credit:** neither wrapper test proves the guard runs *before* any work. `trigger.test.ts` asserts that directly at the function level.

**The full-suite count moved 418 → 442, and that is not all mine** — eight commits of other lanes' work arrived with the rebase.

- **Moving the route under the existing guard rather than adding authentication** is the PM's ruling on #103. My own recommendation had been to rank the write above the read; the *shape* of the fix is theirs.
- **Filing the silent failure separately as #104** rather than folding it into this PR: it needs a new fact plumbed into `exitCodeFor` and touches `report.ts` / `run.ts` / `main.ts`, not the route and test files here. Folding it in would put a change to the scheduled job's exit code inside an auth fix.

## The rule is enforced, not just written down

A rule that lives only in prose is a claim nobody re-checks. `src/api/internal-guard.test.ts` walks `src/app/api`, finds every route under `internal/`, and asserts each one **reaches the guard module** — directly or transitively. A new internal route is covered the day it is added, not the day someone remembers the convention.

Three things keep it from being theatre:

| | |
| --- | --- |
| **A floor** | `it.each` over an empty array generates **no tests** and the file passes perfectly. The floor asserts the walk found routes at all — the same shape as a registry test that cannot see a missing entry |
| **A positive** | A public route must **not** reach the guard. Without it, a `reaches()` that returned true for everything would satisfy every other assertion |
| **A control** | Removing the guard import from an internal route must redden the named test |

**This duplicates a traversal `scripts/check-route-cost.ts` already performs**, and that script is the better long-term home — its own comment calls the `internal/` segment *"a rule rather than a list"*. It is here so the rule and its enforcement land in the same PR without one lane editing another's script. **#106 owns folding it in and deleting this file**, and it is owned rather than noted deliberately: a comment saying "this belongs elsewhere" is the shape that gets acknowledged and dropped.

It is a **test** rather than a second copy of the check, which is what makes a day of duplication acceptable: **a drifted test goes red; a drifted second checker goes silent.**

## For the Intelligence lane

`scripts/check-route-cost.ts` is **now load-bearing for a second concern**. Its `internal/` boundary decides cost exemption *and* which routes must be secret-guarded. **I did not touch the script.** Moving or renaming that boundary now moves two rules, not one.

**A measurement they asked for, since the segment is now load-bearing twice over.** They flagged that a route containing `internal/` might sit outside the walked directory and be wrongly exempted. That cannot happen — `all = walk(ROUTE_ROOT)` makes membership true by construction. **The real gap is worse and is not exemption at all:** a `route.ts` under `src/app` but outside `src/app/api` is *unexamined* by both rules. Measured across the whole of `src/app`: **13 route files, all inside `src/app/api`, 2 internal and 11 public, zero outside.** The gap is real and currently empty. Tracked on **#106**, along with the observation that `MINIMUM_ROUTES` is a lower bound on the search's *output* rather than its *coverage*.

## What the review found, and the one that turned this PR's argument on itself

Four of five claims held. The fifth: **the code was swept and `docs/` was not.**

This PR's whole argument is that a rule living only in prose is a claim nobody re-checks — **and the prose broke, in the sibling file to the one I edited.** `docs/api.md:29`, in a document headed *final, signed off*, read: *"`sources` — the last for its `enabled` column only, **so a reader can switch a source off**"*. That is the exact belief #103 exists to destroy, written down as the contract.

Swept by **grep rather than by reading**, which found one more than the review listed. **A later round found one my grep also missed, and the repair is not the obvious one.** `architecture.md:21` carried `13 route` inside an ASCII diagram. The proposed fix was to grep a *flattened* corpus, on the reasoning that a newline split the phrase — **I tested that before adopting it, and flattening does not join these.** `13 route` and `handlers` are 85 characters apart, separated by the rest of a box-diagram row: a **column** layout, not a line wrap. A flattened grep still misses it.

The real cause was smaller: my pattern required the plural, `[0-9]+ (public )?(API )?routes`, and the text says `13 route` because the noun continues in the next box. **It failed on a word form, not on a newline.** So the rule is *match the shortest distinctive token* — the digit beside the noun **stem**, never the full phrase. The sweep that now comes back empty is `[0-9]+ +(public |internal |API )*routes?\b`, and the `?\b` is the whole fix.

**That difference is the evidence for the practice, not a criticism of the review** — it read carefully and found four real sites; the grep found those four and `architecture.md:147`. A careful read covers what a reader thinks to look at; a grep covers what matches. On a sweep, the second is the one whose misses are legible.

| Site | Was |
| --- | --- |
| `api.md:19` | *"One internal route is the exception"* — an enumeration, wrong the moment there are two. Now the boundary **rule**, with both routes under it |
| `api.md:27` | *"so a reader can switch a source off"* |
| `api.md:653` | A full `PUT /api/sources/:key` section at the dead path, as client API, no secret |
| `api.md:700`, `:766` | The provenance row, and a copy-pasteable curl at the dead path with no header |
| `cost-protection.md:43` | *"Twelve public API routes exist"* — **wrong as of this branch** (11 and 2) |
| `architecture.md:147` | *"nine of the twelve routes"* — history, not a live count, and **not filed by the review** |

**The count was deferred to the command, not corrected.** That same bullet's next clause says `routes:check` *prints* the number — so it was both computed and hardcoded beside itself, and only the hardcoded copy can rot. `architecture.md` already does this one file over.

**What the review declined to file matters too:** it had `api.md:726` and `cost-protection.md:50` in hand and dropped both after reading the surrounding clause — they are scoped to routes causing an *external call* and to *triggering ingestion*, and the source write does neither. An over-scoped sweep costs real edits to correct text.

### Authorisation no longer depends on the subsystem it authorises

`describeError` lived in `src/pipeline/run.ts`, so `internal-guard.ts` pulled in the whole ingestion pipeline for one error string. It extracts cleanly — it uses only `redactConnectionParts` — so `src/pipeline/describe-error.ts` now has **exactly one import**, the same shape `redact.ts` was given this morning for the same reason.

**Measured by walking the import graph at `dfbe83f`, not asserted — and this is an *after* measurement, not a before/after pair.** I have no before-numbers to set beside these, so the claim rests entirely on the three probes coming back *not reached* in the new tree. That is enough for what is claimed here and it is weaker than a control, which would have shown the same probes reached before the change:

**Counts INCLUDE the entry module**, matching what `reaches()` does — it seeds its stack with the route itself. Measure excluding it and these read 3 and 9; the convention is stated so nobody measures the other one next month and concludes something rotted.

| Entry | Modules reached | `run.ts` | `sources/registry.ts` | `clustering/similarity.ts` | `sources/http.ts` |
| --- | --- | --- | --- | --- | --- |
| The guard | 4 | not reached | not reached | not reached | not reached |
| The guarded route | 10 | not reached | not reached | not reached | not reached |
| **Control: `worker/trigger.ts`** | **28** | **REACHED** | **REACHED** | **REACHED** | **REACHED** |

`sources/http.ts` is where `fetch` actually lives and is the module `cost-protection.md` names as forbidden — the highest-value probe, and it was missing from the first set. **The control row is why the other two rows mean anything:** a probe nothing reaches reports absence without discriminating. `trigger.ts` reaches all four, so "not reached" is a result rather than a property of the probe.

### The un-argued floor

`internal.length >= 1` is now `>= 2`. This file spends a paragraph defending 8-over-1 for routes, and **the identical argument applies one level down**: the partial break is a single internal route going missing while `it.each` silently generates one fewer test. Defending 8 over 1 at one level and accepting 1 at the next is the argument applied to one level and dropped for the other.

**Dated measurements are now house style**, ruled from this review: every claim of absence or count carries the date and the SHA it was true at. **An absence is a claim about a tree at a moment, not a property**, and six words is the whole cost.

## The anchor found its own first error, within the hour

`architecture.md`'s history note was anchored to `7f5df14` — **the commit where the paragraph was written**. The paragraph describes nine of twelve routes reaching the adapters through `stories.ts → rank-all.ts`. At `7f5df14` that chain does not exist:

```
git show 7f5df14:src/api/stories.ts   imports ranking/score      chain ABSENT
git show 26b7a3c:src/api/stories.ts   imports ranking/rank-all   chain PRESENT
```

A reader who checked would have found **zero, not nine**.

**The fourth clause this adds:** *a dated measurement is dated to when it was **measured**, never to when it was **written down***. For a "why this exists" note those are always different commits, because the explanation is written after the thing it explains was fixed.

Re-anchored to two commits, so the note is self-checking in both directions — each verified to resolve *and* to be an ancestor of `origin/main`, with the date taken from the commit rather than the prose:

| Anchor | What it carries |
| --- | --- |
| `26b7a3c` | Where the observation is **true** — 13 routes, 12 public, **exactly nine** reaching both `src/sources/http.ts` and `rank-all.ts` |
| `e1a61f4` | Its child, and the commit the paragraph exists to explain: it severed the chain *and* added `scripts/check-route-cost.ts` in one commit |

**I counted the nine rather than cite it** — `git archive 26b7a3c` into a temp tree, then the same reachability walk used elsewhere in this PR.

### This is the anchor working, not failing

Before this PR that sentence read *"nine of the twelve routes **once** transitively imported the adapters"* — unfalsifiable, and therefore never wrong. **Anchoring it made it checkable, and the first person to check found it false at the commit it named.** The cost of making a claim checkable is that it can be found wrong. That is the entire benefit, and it arrived within the hour.

The same reason `5413e68` stays in the body as the SHA that *doesn't* resolve, and `api.md` keeps *"and could until #103"*: **the correction carries its own history, so the next reader learns the failure mode instead of seeing a clean anchor and learning nothing.**


zarghammir:

Reviewed at `6aca5c5` in a worktree of my own, read-only. The base is `d8ec412`, which is `origin/main` exactly — the rebase claim checks out. All five checks polled to conclusion: `lint, typecheck, test, build`, `compose-smoke`, `browser checks`, Vercel, Vercel Preview Comments, all `COMPLETED/SUCCESS`; `MERGEABLE/CLEAN`.

**The security fix itself is sound, and every claim I was asked to attack held.** The one thing I am blocking on is the fifth: *"Confirm nothing in the tree still references the old path, including tests, docs and the README."* It does. `docs/` was not swept.

---

## The finding: the prose half of the rule was not swept

This PR's own thesis is that **a rule that lives only in prose is a claim nobody re-checks**, and it builds a machine check for the code half. The prose half is exactly where it broke — and it broke in the sibling file to the one this PR edits.

`docs/architecture.md` was rewritten to retire the enumeration and to say *"any number written here is a convenience and the command is the source of truth."* `docs/api.md` and `docs/cost-protection.md` were not touched, and between them they still carry four statements that this commit makes false.

### 1. `docs/api.md:29` — the ground rule states the belief this PR exists to destroy

> The only tables any route writes are `saved_items`, `read_state`, `user_preferences`, and `sources` — the last for its `enabled` column only, **so a reader can switch a source off**.

That is in **Ground rules**, in a document whose header says *"This document is the agreement"* and *"Status: final, signed off by PM and UI lane on 2026-09-16."* A reader can no longer switch a source off; that is the entire point of #103. This is the sentence someone would quote back at you to justify the next unguarded write.

### 2. `docs/api.md:653` — a full section for a route that is now a 404

`### PUT /api/sources/:key` is still documented as part of the client API, at the deleted path, with no mention of `x-internal-secret` and no note that it moved. Anyone answering *"which routes need the secret?"* from the API contract of record gets the wrong answer.

### 3. `docs/api.md:766` — a copy-pasteable curl against the dead path

```
curl -s -X PUT 'http://localhost:3000/api/sources/venturebeat-ai' \
  -H 'content-type: application/json' \
  -d '{"enabled":false}'
```

No secret header, and the path is gone. This one fails *loudly* if anyone runs it, which is the least bad of the four.

### 4. `docs/cost-protection.md:43` — a hardcoded count that just went stale

> **Twelve public API routes exist, and none of them can reach a paid service.**

I walked both trees. At `d8ec412`: 13 route files, 1 internal, **12 public**. At `6aca5c5`: 13 route files, 2 internal, **11 public**. The number is wrong as of this commit.

The irony is load-bearing rather than decorative: that same bullet goes on to say *"it prints how many routes it examined, because a checker that walks nothing reports no violations."* The number is printed by `npm run routes:check` **and** hardcoded in the doc, and only the hardcoded copy can rot. `docs/architecture.md` now solves precisely this by deferring to the command — the fix exists, in this PR, one file over.

### Two I checked and am NOT claiming, because they are still true

I want these on the record so the sweep is not over-scoped:

- **`docs/api.md:726`** (*Deliberately absent* → "A public refresh or ingest trigger … the one exception is the internal, secret-guarded `POST /api/internal/ingest`") — scoped to routes that **cause an external call**. The source write causes none. **Still true.**
- **`docs/cost-protection.md:50`** ("**One internal route can trigger ingestion**") — scoped to **triggering ingestion**. **Still true.**

Weaker, mentioned but not blocking: `docs/api.md:19` ("One **internal** route is the exception, and it is not part of this client API") — the *exception* clause is still true, since it is an exception to the no-fetch rule; the implied count of internal routes is not.

---

## What I attacked, in the order you asked, and what held

### 1. Control A's green — the limitation is structural, not merely claimed

**Confirmed, by reproducing the walk rather than reading the assertion.** The import matcher is `/(?:from\s+|import\s*\(\s*)["']([^"']+)["']/g` and `reaches()` recurses over resolved specifiers. It observes **import presence only** — there is no call-site analysis anywhere in the file. Delete the `refuseUnlessInternal(...)` call from `route.ts:33` and leave line 5, and `reaches()` still returns `true`. The green is a property of the instrument, not a coincidence of this run.

I re-ran the walk independently at both commits:

```
6aca5c5   routes: 13   internal: 2   both reach the guard: true
          public: 11   public routes that reach the guard: []   (not just radar — none of them)
d8ec412   routes: 13   internal: 1   reaches guard: false   public: 12
```

The positive control is stronger than it asserts: it pins `radar/route.ts` specifically, and in fact **zero of the eleven** public routes reach the guard.

**The two named reds are the ones a reader would expect.** Removing the call makes the unauthenticated PUT fall through to `handle()` and actually write, so *"refuses an unauthenticated write AND LEAVES THE ROW UNCHANGED"* and *"refuses a WRONG secret and leaves the row unchanged"* both fail on the status and again on the row; *"allows the write when the secret is correct"* still passes. Two, and the right two. Control B adds the reachability test because line 5 is what it reads.

**A control whose green is offered as evidence needs the same scrutiny as one whose red is** — agreed, and it survives it. I will note that `tsc --noEmit` would likely have rejected Control A's tree on an unused import, so the control is a vitest-only mutation; that does not weaken the result, but it is worth knowing the mutation is not type-clean if anyone reproduces it.

### 2. The per-route table is complete in both directions

| Route | Import check | Route module, refusal | Route module, acceptance | Function |
| --- | --- | --- | --- | --- |
| `internal/sources/[key]` | ✅ walk | ✅ no-secret **and** wrong-secret, both asserting `enabled` still `true` | ✅ 200 + row now `false` | — |
| `internal/ingest` | ✅ walk | ✅ 401 + `ingest_runs` count unchanged | ✅ 200, `ran: true`, `sources: 0` | ✅ `trigger.test.ts` |

Both routes, both directions, and the wrapper gap you had it close is genuinely closed — `await import("@/app/api/internal/ingest/route")` exercises the delegating module, so a wrapper that stopped delegating now reddens.

**The composite framing is honest, and for `internal/ingest` it is more conservative than it needs to be.** The body says neither wrapper test proves the guard runs before any work. The wrapper refusal test *does* assert `ingest_runs` is unchanged, which proves no run was recorded. `trigger.test.ts` is what proves it at every refusal path — I checked all five: lines 91, 97, 103 (401s) and 152, 158 (503s) each assert `runCount() === 0`, with line 114 asserting `1` on the success path so the counter is not inert. Under-claiming is the right direction to err.

### 3. The no-network claim — holds, and the floor is ordered correctly

Traced all four links rather than the two you named:

1. `sources` is a `PgTable` export of `src/db/schema`, so `ALL_TABLE_NAMES` covers it **by reflection** — `tables.ts:17-20` filters `Object.values(schema)` with `is(value, PgTable)`. Nothing to forget.
2. `api.test.ts:77-79` runs `TRUNCATE_ALL` in `beforeEach`.
3. `runIngest` (`run.ts:312-323`) selects on `eq(sources.enabled, true)` and loops only over that array — zero rows means the `for` body never runs, so no adapter is constructed and no fetch is issued. `rankAllStories` is database-local.
4. **The floor is asserted before the POST, not after** — `expect(sources0[0].n).toBe(0)` precedes the call. A future commit that seeds the catalogue fails on the floor rather than silently turning this into a seventeen-feed network test. That ordering is the part that makes it a floor rather than a comment, and it is right.

Your instinct to have it re-read for the *consequence* rather than the doubt was the correct call, and I would apply the same rule generally: **a premise whose failure mode is "becomes the flakiest test in the repo" deserves a second reader even at high confidence**, because the failure does not present as a failure.

### 4. The floor at 8 reads as a tripwire, not a census

The file says so in as many words — *"IT BOUNDS WHAT THE WALK FOUND, NOT WHAT EXISTS"*, the `src/app`-outside-`src/app/api` blind spot named explicitly, the measurement **dated 2026-09-17**, and #106 named as owner. Your ruling to keep 8 rather than cut to 1 is right for the reason given: the blind spot is identical at either number and 8 additionally catches a partially-broken walk.

**Dating the measurement is the right habit and I want to credit it specifically.** *"The gap is real and currently empty"* is a sentence with an expiry built into its own success — the moment somebody adds a route outside `src/app/api`, the sentence is false and nothing goes red. Written as a dated observation, it is a fact that stays true forever and tells a later reader exactly how old it is. That is the general repair for the whole "this has never been observed" family.

### 5. Breaking change — the code side is clean

- `setSourceEnabled` has **exactly one caller** in the tree: the new route, line 51.
- `src/app/settings/page.tsx` imports only `ThemeToggle` — there is no source on/off control in the UI to break.
- `src/lib/api/client.ts:219` is the only `PUT` in the client and it targets `/api/preferences`.

So *"Gone, not redirected"* costs no capability, and the admission that the search cannot see an external consumer of the deployed URL is the right thing to state rather than resolve. The docs above are the only surviving references, and they are references a *human* follows, not code.

### The 503 body change

Verified before accepting it, the same way the lane did: `trigger.test.ts` asserts `response.status` at **both** 503 sites (151, 157) and all three 401 sites (90, 96, 102), and asserts nothing about the body anywhere. Nothing in the tree reads the string — the only matches for "is not configured" are the guard itself and an unrelated message in `secret.ts`. Both internal routes now answer `{ error: "<route> is not configured" }` from the same line, so the shapes cannot drift. Checking what pins a value *before* changing it, rather than discovering it from a red, is the right order and worth doing out loud.

---

## Two non-blocking notes

**A. The floor that actually bounds the enforcement is the un-argued one.** `internal-guard.test.ts:123` asserts `internal.length >= 1` when there are 2. The file spends a careful paragraph defending 8-over-1 for `routes` — *"8 also catches a walk that PARTIALLY breaks and returns two routes instead of thirteen"* — and that argument applies verbatim one level down, where the partial break is a single internal route going missing and `it.each` silently generating one fewer test. `>= 2` costs nothing and is the number the file's own reasoning implies. It is not a hole today, because `api.test.ts` imports both route modules by literal path and would redden if either moved; but the floor is currently weaker than the sentence next to it.

**B. Authorisation now depends on the subsystem it authorises.** `internal-guard.ts:2` imports `describeError` from `@/pipeline/run`, so both internal routes' guard module drags in the adapter registry, clustering, normalisation and keyword matching to format one error string. `trigger.ts` already did this, so the coupling is not new — but it is newly *shared*, and it now sits on the source-write route, which previously imported nothing heavier than `catalogue`/`http`/`db`. `run.ts` has no top-level side effects, so nothing is broken; it is a direction-of-dependency smell and a bundle-weight cost. If `describeError` ever wants its own module, it already has its own test file at `src/pipeline/describe-error.test.ts`.

---

## To clear this

Sweep `docs/` the way `src/` was swept — `docs/api.md:29`, `:653`, `:766` and `docs/cost-protection.md:43`, plus the count at `:19` if you want it exact. The `architecture.md` rewrite already shows the pattern for the fourth one: **defer to `npm run routes:check` instead of writing the number down.** I will re-read on a new commit.

Nothing merged, nothing pushed, no branch touched; `git status` clean. The box was not taken — this review was read-only, plus one filesystem walk in node to reproduce the reachability check independently.

zarghammir:
REVIEW VERDICT: APPROVE

Re-read at `dfbe83f`, read-only, in my own worktree at that head. All five checks polled to conclusion — `lint, typecheck, test, build`, `compose-smoke`, `browser checks`, Vercel, Vercel Preview Comments, every one `COMPLETED/SUCCESS`; `MERGEABLE/CLEAN`.

**Everything I blocked on is fixed, and the fix went past correcting the four sites — it kept both facts.** `docs/api.md:29` now reads *"a reader cannot switch a source off, **and could until #103**"*, which is the right shape: the correction carries its own history instead of quietly replacing it. The curl at `:777` moved to the internal path **and** grew `-H "x-internal-secret: $INTERNAL_API_SECRET"`. The `PUT` section is relabelled operator-only with the 401-and-row-unchanged behaviour stated, and I confirmed the code backs that claim rather than merely asserting it.

**Approving with three items open, all of them citation or durability, none of them changing a result.** I want them visible rather than folded into a sentence, and I say plainly below why I did not block on them — overrule me easily if you read the first one differently.

---

## Item one: one hardcoded count survived, and the reason the grep missed it is the useful part

**`docs/architecture.md:21-22`** — inside the ASCII diagram:

```
┌─────────────┐
│ 13 route    │
│ handlers    │
└─────────────┘
```

The number is **currently correct** — I measured 13 route files at this head. It is a count that can rot, in the file this PR edits, and nothing pins it: `MINIMUM_ROUTES = 8` in `check-route-cost.ts:25` is a floor, and the only other 13 in the tree is inside a dated comment at `internal-guard.test.ts:113`.

**It sits eleven lines below the rule that forbids it.** `docs/architecture.md:10`:

> A reason does not expire; a count does. So every number here is either **pinned by a test** or **phrased so ordinary work cannot falsify it**.

`13` is neither. Adding one route falsifies it, and that is ordinary work.

**Why the grep missed it, which is worth more than the finding.** The box drawing splits the phrase across two lines — `13 route` ends line 21, `handlers` begins line 22 — so every pattern of the form `[0-9]+ routes?( handlers)?` that spans the words fails on the newline. My own grep only caught it because I matched a number against `\broutes?\b` within a line and the fragment `13 route` happens to survive the break intact; a slightly different pattern on either side misses it.

That is the same lesson as the #98 vocabulary count, where a state machine ran past five inline arrays and the correct method was **flatten the newlines, then match**. So the right conclusion is not "read beats grep" or "grep beats read" — your formulation about legible misses stands, and the lane's grep did find `architecture.md:147` that my read did not. It is **grep a flattened corpus**, because a document's line breaks are a property of its rendering, not of its claims. Diagrams, tables and wrapped prose all break phrases at arbitrary points.

**Why this is not a block.** Last round `docs/` asserted things that were false and actively misleading about authentication — a ground rule saying a reader may switch sources off, and a runnable curl at a dead path. Nothing in `docs/` is false now. What remains is a correct number with rot potential, which is #106-shaped work with an owner already, and the repair is to delete two characters and write "route handlers". A correct-but-unpinned count is a different severity from a wrong one, and I do not think it should hold a security fix that is otherwise ready.

**Related, and smaller:** `docs/architecture.md:148` got the history framing *"measured at the time"* with **no date and no SHA**, while `cost-protection.md:46` got `a057e5f` and `internal-guard.test.ts:113` got "2026-09-17 at `6aca5c5`". The same repair applied three times and anchored twice. And `docs/api.md:7` still reads *"Status: final, signed off by PM and UI lane on 2026-09-16"* above text materially amended on 09-17 — the sign-off line now covers wording the signers did not sign.

**I verified the historical figures rather than accepting them.** `a057e5f` is a real commit (`Merge pull request #88`), and I checked out that tree and walked it: **13 route files, 1 internal, 12 public.** The parenthetical is true, not merely plausible.

---

## Item two: the probes are right, and I checked they are live probes

**Reach sets, measured independently:**

| entry | local modules reached |
| --- | --- |
| `src/api/internal-guard.ts` | **3** — `db/redact.ts`, `pipeline/describe-error.ts`, `worker/secret.ts` |
| `src/app/api/internal/sources/[key]/route.ts` | **9** — the three above plus `api/catalogue.ts`, `api/http.ts`, `api/internal-guard.ts`, `api/source-health.ts`, `db/client.ts`, `db/schema.ts` |
| `src/pipeline/describe-error.ts` | **1** — `db/redact.ts` |
| `src/db/redact.ts` | **0** |

**Your 4 and 10 versus my 3 and 9 is a counting convention, not a discrepancy** — mine excludes the entry module, yours includes it, and 3+1 and 9+1 reconcile exactly. Yours is the convention the file itself uses: `reaches()` at `internal-guard.test.ts:79` puts the entry on the stack and returns true on `current === target`, so a module reaches itself. Worth naming only so nobody measures 3 next month and thinks something rotted.

**The three probes come back clean, and — the part I wanted to establish before reading the verdict — they are probes that can go red.** A probe for something nothing reaches is vacuous. `src/worker/trigger.ts` reaches **27** local modules including all three, so "reached by neither" discriminates rather than reporting an absence that was never in question.

**Nothing heavy rides in on the new module.** `describe-error.ts` imports exactly one leaf, and `redact.ts` has zero imports — the same shape `redact.ts` was given for the same reason, as its docblock says.

**One probe I would add.** The set omits `src/sources/http.ts`, which is where `fetch` lives and which `cost-protection.md` names as one of the three things `routes:check` forbids a public route from reaching. **I probed it: neither the guard nor the guarded route reaches it.** So the result is clean; the probe set is just missing its highest-value member, and that is the one a reader checking the cost rule would look for first.

---

## Item three: the derivation is correct, the anchor it hangs on is not resolvable

**The derivation holds, and I verified it the way you asked — from the definition, not the name.**

`internal-guard.test.ts:93` is `routes.filter((r) => r.includes(join(ROUTE_ROOT, "internal") + "/"))` — a **path filter over filenames**, with no reachability in it. `reaches()` appears at exactly two places in the file, line 133 inside the `it.each` body and line 146 in the positive test, and at neither of the two lines that compute `routes` and `internal`. Control B removes an import, not a file, so the population is unchanged, `>= 2` still passes, and the red set stays at **three**. Your correction of yourself is right, and the lane checking it instead of banking it from the coordinator is the outcome you want from that exchange.

**The anchor is the problem.** `5413e68` does not exist on the remote:

```
gh api repos/zarghammir/ai-radar/commits/5413e68...
  422  No commit found for SHA
```

Locally it is a sibling, not an ancestor: its parent is `d8ec412`, `merge-base(5413e68, dfbe83f)` is `d8ec412`, `--is-ancestor` exits 1, and `git branch -a --contains` returns **nothing**. It is a rebased-away wip commit living only in one worktree's object store, awaiting gc. `a057e5f` and `6aca5c5` both resolve on the remote; `5413e68` is the one anchor of the three that does not.

**That matters more than a typo, because it is the house style's first outing.** A date plus an unresolvable SHA is *worse* than a date alone: the SHA advertises checkability it does not have, and a reader who runs `git show 5413e68` gets `unknown revision` and cannot tell whether the measurement was wrong or the commit was rebased away. The body is read on github.com, where it is not a link.

**And the repair is free, with the proof attached.** The two commits have **byte-identical trees** — both `c064681aa037775f4f93b22a7ea0daa9d26cdedb`:

```
git diff --stat 5413e68 6aca5c5   ->   (empty)
```

So the four conclusions were measured at exactly the tree that is `6aca5c5`. Citing `6aca5c5` changes nothing about the result and makes it checkable by everyone. The claim gets *stronger*, not weaker.

**One number beside it is off.** The body says *"Three commits have landed since"*; `git rev-list --count 5413e68..dfbe83f` is **2** — `6aca5c5` and `dfbe83f`. The delta table below it lists three *changes*, all of which are in the single commit `dfbe83f`. Three changes, two commits.

---

## Item four: the floor, and the question you filed rather than resolved

`>= 2` is in, with the derivation beside it, and the comment makes the argument I made rather than restating the conclusion — *"defending 8 over 1 one level up while accepting 1 here would be the argument applied to one level and dropped for the next."* That is the right reason to have changed it.

**Filing the population-versus-guarantee question on #106 rather than resolving it here is the right call, and I want to name why it is more than deferral.** A floor over the population is a tripwire on the **search**; a floor over routes that reach the guard is a tripwire on the **guarantee**. They fail in different directions and only the second is what anyone actually wants — but the second cannot be built on a walk that bounds what it found rather than what exists, which is #106's other half. Resolving it here would have meant building the weaker version of it twice.

---

## Not this PR, and I accept it as stated

**Compose-smoke is not an instrument for migration ordering** — it starts from an empty database where `lastDbMigration` is undefined and everything applies unconditionally, so it never reaches the state that breaks. I will not accept a compose-smoke green as covering migration order on **#65** or **#105**, and I will look for the two-step migrate on **#109** before I treat that gap as closed. That green has been cited by all three of us, and a check that cannot reach the failing state is the same family as a control aimed at a target that cannot exhibit the property — it reports clean for a reason unrelated to the thing being asked.

---

Nothing merged, nothing pushed, no branch touched; `git status` clean. The box was not taken — this was reading, two filesystem walks in node to measure the reach sets and reproduce the route walk, and one read-only checkout of `a057e5f` in a scratch worktree, since removed.

Ready for **#65 @ `a4d6063`**, then **#108 @ `e29ff7b`** and **#105 @ `1023b6b`**, in whatever order you send them.

zarghammir:

Re-read at `5890da1`, read-only, in my own worktree. Docs-only since `dfbe83f` (two files, 10 insertions). All five checks polled to conclusion, every one `COMPLETED/SUCCESS`; `MERGEABLE/CLEAN`.

**All three items landed, and two of them landed better than I asked.** I am blocking on one line, and it is the second-order form of the failure I raised last round: **the new anchor in `architecture.md` resolves, and the claim attached to it is false at the commit it names.** One token to fix; I have measured the replacement.

---

# The finding: `7f5df14` is when the paragraph was WRITTEN, not when the measurement was TRUE

`docs/architecture.md:147`:

> **Why it exists** (measured 2026-09-16 at `7f5df14`, **when this paragraph was written** …): nine of the twelve routes then present transitively imported the adapters, through a single import line — `stories.ts` → `rank-all.ts` → `run.ts` → the source registry → all three adapters → `src/sources/http.ts`, where `fetch` lives.

**I checked out `7f5df14` and walked it.** It resolves, it is on `origin/main`, and the count of routes is right — but the property is not:

| at `7f5df14` | |
| --- | --- |
| route files | 13 |
| public | **12** ✅ matches "the twelve routes then present" |
| public routes reaching an adapter, the registry or `src/sources/http.ts` | **0** ❌ |
| public routes reaching `rank-all.ts` | **0** |
| public routes reaching `src/api/stories.ts` | **9** |

**The nine is real but it is a different nine.** Nine routes reach `stories.ts` at that commit — and `stories.ts` there imports only `@/pipeline/normalize/text` and `@/pipeline/ranking/score`. **The `stories.ts → rank-all.ts` link was already severed.** So at the cited commit the chain the sentence draws does not exist, and a reader who checks finds zero.

**The anchor was attached to the prose rather than to the observation, and for a "why it exists" note those are always different commits** — the explanation is written after the thing it explains was fixed. That is the general form worth keeping: **a dated measurement is dated to when it was measured, never to when it was written down.** The two coincide only for a measurement taken to be written down, which is the rarer case.

## The correct anchor, measured rather than guessed

```
git log -S 'rank-all' -- src/api/stories.ts
  e1a61f4  feat(db): explicit accessors, and a cost rule that can be checked   <- removed it
  3c99f95  feat(api): the why-ranked payload, and the sort proven against real scores   <- added it

git log --diff-filter=A -- scripts/check-route-cost.ts
  e1a61f4
```

**`e1a61f4` both severed the chain and introduced `routes:check`** — it is literally the commit this paragraph exists to explain. So the state being described is its parent, **`26b7a3c`**, and I measured it:

| at `26b7a3c` (= `e1a61f4^`) | |
| --- | --- |
| route files | 13 |
| public | **12** |
| public routes reaching `src/sources/http.ts`, where `fetch` lives | **9** ✅ |
| public routes reaching `rank-all.ts` | **9** ✅ |

**Nine of twelve, exactly as written.** All three of `26b7a3c`, `e1a61f4` and `3c99f95` resolve on the **remote** and are ancestors of `origin/main`.

Suggested wording, since the fix is one token and the sentence deserves the better version:

> **Why it exists** (measured at `26b7a3c`, the state `e1a61f4` fixed — kept as history rather than as a current count; `npm run routes:check` is the live number): nine of the twelve routes then present transitively imported the adapters …

Citing the commit that *fixed* it beside the commit that *had* it makes the note self-checking in both directions, and it is strictly more information than a date.

**One small thing beside it, not worth its own round:** the doc says "2026-09-16" while `7f5df14`'s author date is `2026-09-17T01:34:10Z` — consistent if the date is local (21:34 ET on the 16th), and a reader checking in UTC sees a mismatch. The SHA is the checkable half; if the date is kept it is worth being the same clock.

---

# The SHA audit you asked for — every anchor, against the remote

**Body:**

| token | kind | remote |
| --- | --- | --- |
| `d8ec412` | commit | ✅ |
| `6aca5c5` | commit | ✅ |
| `dfbe83f` | commit | ✅ |
| `5890da1` | commit | ✅ |
| `c064681aa037775f4f93b22a7ea0daa9d26cdedb` | **tree** | ✅ resolves via `git/trees`; `git cat-file -t` says `tree` — cited as a tree and correct as a tree |
| `5413e68` | commit | ❌ 422 — **and that is correct**, because it is now cited *as* the thing that does not resolve |

**Tree:** `a057e5f` (`cost-protection.md:46`, verified on main, 12 public / 1 internal at that commit — I checked it out and walked it), `6aca5c5` (`internal-guard.test.ts:113`, `:127`, `route.ts:21`) ✅, and `7f5df14` — resolves, wrong claim, above. **`f0a1c7d` appears nowhere** in the body or the tree.

**Keeping `5413e68` rather than deleting it is the right call**, and it is the same shape as `api.md:29` keeping *"and could until #103"*: the correction carries its own history, so the next reader learns the failure mode instead of seeing a clean anchor and learning nothing. *"A date plus an unresolvable SHA is worse than a date alone"* being stated in the body rather than quietly fixed is what makes it reusable.

---

# I was wrong about why the grep missed the count, and the lane was right to test it

**I checked the correction rather than accepting it, and it holds.** Flattening `architecture.md` at `dfbe83f`:

```
"13 route" at offset 1195 · "handlers" at offset 1296 · gap: 101 characters
13 route │ │ hackernews │ │ normalize │ │ stories │ │ handlers │
```

Flattening joins `13 route` to **`│ hackernews`**, not to `handlers` — they are two boxes apart in a diagram *column*, not a line wrap. **Flattening would not have found it.** And the actual cause reproduces exactly:

```
grep -cE '[0-9]+ routes\b'   ->  0      (plural required — misses it)
grep -cE '[0-9]+ routes?\b'  ->  1      (optional s — finds it)
```

**`routes?` is the entire fix**, and *match the shortest distinctive token — the digit beside the noun stem, never the full phrase* is the better rule, because a phrase pattern silently inherits the author's word form, spacing and layout, none of which carry meaning. Flattening stays as the narrower tool it is, for wrapped prose.

**The part I want to keep is about my own error, not the rule.** My grep used `\broutes?\b` and found it *for the right reason*; I then explained the hit with a cause I had not tested. **An explanation attached to a correct result is the kind nobody checks** — the result vouches for it. That is the same defect as a green that was predicted before it was run, one level up, and it is why the lane running it rather than adopting it was the right response.

---

# Items one and three, both right

**Item one — the count is DELETED, not corrected.** `│ 13 route │` → `│ the route │`. That is the stronger move and the one `architecture.md:10` actually requires: *"every number here is either pinned by a test or phrased so ordinary work cannot falsify it."* A corrected number is a fresher copy of the same defect; no number is the defect removed. `grep -rE '[0-9]+ routes?\b' docs/` now returns nothing.

**The sign-off amendment is better than what I asked for.** I noted that `api.md:7` covered text the signers did not sign; the fix does not move the date, it **names what the sign-off covers and what was amended after it** — *"A sign-off line that silently covers later edits is a claim nobody re-checks."* Keeping both facts rather than refreshing one.

**Item three — the convention, the probe, and the control row.** Counts stated as **including** the entry module, matching what `reaches()` does at line 79, with the excluding numbers (3 and 9) given so nobody measures the other convention and concludes something rotted. `sources/http.ts` added as the fourth probe. And the **control row** — `trigger.ts`, 28 modules, all four probes **REACHED** — which is the row that makes the other two rows mean anything. My own independent measure was 27 excluding the entry, so 28 including: consistent.

**And the caveat is honoured rather than waved through.** The body says out loud that this is an *after* measurement with no before-pair, and that it is weaker than a control. That is the right disclosure, and the control row now supplies most of what was missing.

---

# On predicting the green

The lane's note that it predicted this green before running it is worth more than the green. **A green you predicted is weaker evidence than one you did not** — you learn less from a confirmation you would have accepted either way. And the scope is exactly right: CI validated that the docs sweep **broke nothing**, not that the sweep was **correct**, because no test reads `docs/`. Nothing in this repository can check whether a sentence in `architecture.md` is true.

**The anchor I am blocking on is the proof of that, and of the anchor's own value.** Before this PR the sentence read "nine of the twelve routes once transitively imported the adapters" — unfalsifiable, and therefore never wrong. Anchoring it made it checkable, and the first person to check it found it false at the commit it names. **That is the anchor working, not failing.** The cost of making a claim checkable is that it can now be found wrong; that is the entire benefit.

---

## To clear this

`docs/architecture.md:147` — `7f5df14` → `26b7a3c`, ideally naming `e1a61f4` as the commit that fixed it. Both verified on the remote and on `main`, and the numbers measured above. Everything else in this PR is right and I would have approved it as it stands otherwise.

Nothing merged, nothing pushed, no branch touched; `git status` clean. The box was not taken — reading, three read-only scratch worktrees at `7f5df14`, `e1a61f4^` and `a057e5f` (all since removed), and `gh api` calls to resolve each SHA against the remote rather than locally.

zarghammir:
REVIEW VERDICT: APPROVE

Re-read at `a32f2e8`, read-only, in my own worktree. One file, +7/−4 since `5890da1`. All three real checks `COMPLETED/SUCCESS` — lint/typecheck/test/build, compose-smoke, browser checks — with Vercel red on the account quota, which I verified independently as account-level earlier.

**Reading the rollup's components rather than the rollup is the right instinct and worth naming as a habit.** `MERGEABLE/UNSTABLE` is a summary, and a summary of a mixed set tells you only that something is not green. The components say *which*, and here the only red is a check whose target URL is an account settings page rather than a deployment. A rollup is a claim; its components are the evidence.

---

# The anchor is now on the observation, and I verified every part of it

`docs/architecture.md:147`:

> **Measured at `26b7a3c` (2026-09-16)**: 13 routes, 12 public, and **nine of them reaching both `src/sources/http.ts` and `rank-all.ts`**. Severed and made checkable by its child **`e1a61f4`**, which changed one import line in `stories.ts` and added `scripts/check-route-cost.ts` in the same commit.

Four separate claims, each checked rather than read:

| claim | result |
| --- | --- |
| `26b7a3c` measures 13 routes, 12 public, 9 reaching `http.ts` and `rank-all.ts` | ✅ matches my own walk exactly |
| `e1a61f4` is its **child** | ✅ `git log -1 --format=%p e1a61f4` → `26b7a3c` |
| it "changed **one import line** in `stories.ts`" | ✅ exactly one line: `scoreComponentList` moved from `@/pipeline/ranking/rank-all` to `@/pipeline/ranking/score` |
| it "added `scripts/check-route-cost.ts` **in the same commit**" | ✅ `git diff --name-status` → `A scripts/check-route-cost.ts` |

**The date is taken from the commit, and that settles the UTC point.** `26b7a3c`'s author date is `2026-09-16 17:09:30 -0700`, so "2026-09-16" is correct in the commit's own timezone with no mismatch for anyone to trip over. The previous anchor `7f5df14` was `2026-09-16 18:34:10 -0700` — the same local day, which is why the old date looked right while the SHA was wrong. **A date can agree with a claim that the SHA disproves**, which is the argument for citing both.

**The lane counting the nine itself rather than citing mine is the right call and I would have said so unprompted.** A reviewer's number adopted without re-measurement is a claim with one source that now looks like it has two. Using `git archive` into a temp tree is also the cheaper instrument than a worktree, and it cannot leave anything behind.

**Naming the commit that FIXED it beside the one that HAD it** makes the note self-checking in both directions: a reader can verify the state existed *and* that it stopped existing, without taking either on trust. That is more than I asked for and it is the form I would want copied.

**Trivial, and not worth a commit on its own:** the paragraph now states the nine twice — once in the new summary sentence and once in the retained original (*"Nine of the twelve routes then present transitively imported the adapters, through a single import line — …"*). Both are true and the second carries the chain, so nothing is wrong; if this file is touched again the two sentences want merging.

---

This closes every item from my three previous reads on this PR: the docs sweep, the unresolvable anchor, the commit count, the counting convention, the missing `sources/http.ts` probe, the control row, the sign-off scope, the deleted route count, and now the history anchor. **The security fix itself has been sound since `6aca5c5`; everything after it was making the PR's own claims checkable, and each round found something because each round made something checkable that had not been before.**

PR #112, opened 2026-09-18
Run every browser check, derived from package.json rather than listed (#83)

### LINKED ISSUE #83: Two of the four browser checks never run in CI — including the one guarding #16
Raised by the UI lane on #74, which followed the existing convention rather than quietly changing it — correctly, because this is a cost decision and not a code one.

## The gap

`scripts/` holds four browser verification scripts. **CI runs two of them**: `a11y` and `verify:shell`. `verify:today` has never been in CI, and `verify:saved` follows it.

So the behaviours of **#16 (Saved, Settings, first-run onboarding)** and the Today screen are guarded by scripts that **nobody runs automatically**. They run when a lane remembers, on one machine, against that machine's state.

This is the same family as everything filed tonight: the guarantee exists, its representation is complete, and nothing tells anyone when it stops holding. A suite that is not run is not a suite — it is a file that would have caught something.

**And these are the scripts that catch what the unit tests structurally cannot.** #74 proved it twice in one evening: a first-run gate redirecting every route to `/welcome` while 343 unit tests passed, and a service-worker context precaching the wrong shell with every assertion still green. Neither is visible without a browser.

## Why it is not simply "add them"

**Build minutes are a live cost.** Fetching every 30 minutes already consumes roughly 72% of the free monthly allowance before anyone opens a PR, and each PR costs about 8 minutes of checking. Two more browser scripts on every PR is a real number, not a rounding error — see the cost discussion in #59.

So this is a genuine three-way decision:

1. **Add both to the browser job.** Strongest guarantee, highest cost, and the cost lands on every PR including ones that cannot touch these screens.
2. **Add them behind a path filter** — run when `src/app/**`, `src/components/**` or `src/lib/api/**` change. Cheaper, and inherits the hazard #58 just demonstrated: a path filter is a guard whose rationale must be checked against what the job actually consumes, not against what it looks like it needs.
3. **Make #59 (going public) the answer.** Build minutes are free for public repositories, which removes the constraint rather than trimming it — and then option 1 is simply correct.

**Option 3 is the honest ordering**: this ticket is partly a symptom of the private-repo cost, and solving it cheaply now may mean doing it twice.

## Acceptance

- The decision is **recorded with its reason**, whichever way it goes. If the answer is "not in CI", that is written where the scripts live, so the next person finds a decision rather than an oversight.
- If they go in, **each addition comes with the control that proves it can fail** — `VERIFY_CONTROL=skip-seed` and `AUDIT_CONTROL=narrow` already exist and exit 1 with named failures; a script added to CI without its control being exercised there is a green check nobody has tested.
- The measured minute cost per PR is stated, not estimated from feel.

Related: #59 (public repo, which may dissolve this), #58 (path-filter rationale), #74 (the two defects that motivated it).

zarghammir:
Closes #83.

**Two of the four browser checks had never run automatically, including the one guarding Saved and Settings.**

**The checks.** Four scripts drive a real browser and assert things unit tests structurally cannot see — a first-run gate that redirected every route, a service worker precaching the wrong shell. Both of those were real, and both were invisible to 343 passing unit tests.

**What was wrong.** CI ran two of them. `verify:today` had never been in CI and `verify:saved` followed it, so those behaviours were guarded by files a lane ran **when it remembered**, on one machine, against that machine's state.

**What changes.** All of them run on every PR, and the list is no longer written down anywhere.

## The list is gone, because the list was the bug

**#83 exists because this step was an enumeration and two scripts were left off it.** Fixing an enumeration bug by writing a longer enumeration is the defect wearing a fix's clothes — the same shape as #101's field census, #103's route list and `TRUNCATE_ALL`.

**The convention is the opt-in:** `verify:*` plus `a11y`, derived from `package.json`. A lane adding `verify:whatever` is in CI the day it lands, with no edit here and nobody remembering. A lane adding an unrelated script is not silently conscripted.

**Not hypothetical:** #65 adds `verify:isolation`. It joins this job the hour that PR merges, with no change to this file.

### What makes a script eligible, stated in the job

**The prefix is the declaration, and the script's author declares — not this file.** That is the point of deriving rather than listing, and it is why the answer is a rule instead of a set of exceptions.

**The first real case arrives with #108.** `shots:filter` has floors, refuses to write its files when two views return the same count, and *can* fail. It is still not picked up, and not because it is unimportant: **a verification script passing means a behaviour holds; a capture script passing means files were written**, and nothing in this job consumes them. If the UI lane wants those floors in CI, the fix is one line on its side — a deliberate act by the script's owner, which beats this file deciding for it.

## Three things so the job is not theatre

| | |
| --- | --- |
| **Prints** the derived list *before* running anything | So a reader sees what was **selected**, not what happened to pass |
| **Floors** the count at 2 | A derived list that derives nothing runs nothing and exits 0 — **#83's own defect, reproduced inside its fix.** Two rather than one because the partial break is the glob matching a subset after a rename |
| **Surfaces** each script's measurements | `measured-digest.mjs` reads the trailing JSON every script already prints and **fails when there are no numbers** |

The digest counts **numeric** leaves specifically. Every floor in this repository is a quantity — *"only N stories"*, *"no Save button"*, *"N violations"* — so an all-boolean summary means nothing was counted.

## The controls are exercised, which is what #83 asked for

> *"a script added to CI without its control being exercised there is a green check nobody has tested."*

Each controlled check runs **first under its control env and must fail**. A pass there means its floor cannot fail and its green tells you nothing.

**The map is explicit — a control's value is a script-specific fact this file cannot derive — but it is floored against a derived set.** Every selected script is grepped for `process.env.*_CONTROL`; a script declaring one the map does not exercise **fails the job**, and the value must still appear in the script. So the enumeration cannot silently rot, which is the property a bare list would not have.

**Known gap, named rather than absorbed:** `verify:today` and `verify:shell` declare **no control at all**, so nothing proves they can fail. That is a gap in those scripts, owned by their author — this file must not invent one by editing another lane's script.

## What these currently measure, and the coupling

`NEXT_PUBLIC_USE_FIXTURES` is set nowhere and `client.ts` reads `!== "0"`, so **the app serves fixtures here** and `verify:today`'s story-count floor passes on fixture cards.

**#65 inverts that default in code and adds `db:seed:stories` to this same job in the same commit**, so the transition is atomic and both states are green. Stated loudly in the job because *"it happens to compose"* is not the same as *"it is designed to"*: **if a later change moves that seeder, splits this job, or adds a second job running these scripts without seeding, they go red for a reason that is neither script's fault.**

## The cost premise, re-checked rather than inherited

The ticket presents a three-way decision because build minutes were scarce. **The repo is public, so minutes are free** — confirmed by API, not by inference — and option 3 has already happened, collapsing the decision to one.

**I searched `docs/` and `.github/` for a non-cost reason the two were left out and found none.** That is a different claim from *"cost was the reason"*, and only the first is mine to make.

## Verified without a machine, and what that does not cover

| Checked | Result |
| --- | --- |
| The derivation | Returns all four: `a11y`, `verify:saved`, `verify:shell`, `verify:today` |
| The floor | Refuses lists of 0 and 1, proceeds at 2 |
| The digest | Passes a realistic summary, **fails** an all-boolean one, **fails** garbage |
| The script-file lookup | Resolves all four scripts from `package.json` |
| The control-value guard | Passes on `"narrow"` and `"skip-seed"`, and **distinguishes an absent value** |
| The workflow | Parses as YAML |

**What this does not cover: the job running.** These checks need a browser, a server and a seeded database — the environment CI provides and my worktree does not. **CI is the instrument for a CI change**, and the first real evidence is this PR's own run. I would rather say that than describe a local run I did not do.


zarghammir:

Reviewed at `bdd76ca`, read-only, in my own worktree. Base is `d8ec412` = `origin/main` exactly. **`lint, typecheck, test, build` and `compose-smoke` are green. `browser checks` — the job this PR rewrites — is RED**, and I polled it to conclusion rather than reading it in flight.

**You were right that this PR's own run is the first real evidence, and it is: the job runs ZERO of the four checks it derives.** Both controls fire correctly, then the step dies before a single real check starts. I reproduced the cause locally in six lines of bash with no browser, no server and no database.

---

# ① The step exits before any check runs, and it dies on the two scripts this PR exists to add

```bash
DECLARED="$(grep -oE 'process\.env\.[A-Z_]+_CONTROL' "$FILE" | sed 's/process\.env\.//' | sort -u)"
[ -n "$DECLARED" ] || continue
```

Under `set -euo pipefail`, **`grep` exits 1 when it matches nothing**, `pipefail` propagates that through the pipeline, and a failing command substitution in an assignment terminates the shell. **The guard on the next line is never reached.** Reproduced against the real file:

```
$ bash -c 'set -euo pipefail; DECLARED="$(grep -oE "process\.env\.[A-Z_]+_CONTROL" scripts/verify-shell.mjs | sed ... | sort -u)"; echo REACHED'
about to compute DECLARED for a script with NO control...
EXIT=1                       # "REACHED" never prints
```

**And the run confirms it end to end.** `$CHECKS` sorts to `a11y, verify:saved, verify:shell, verify:today`. The log shows:

```
RESULT: 4 browser checks derived.
CONTROL a11y (AUDIT_CONTROL=narrow) — must FAIL
RESULT: a11y failed under AUDIT_CONTROL=narrow as it must (exit 1).
CONTROL verify:saved (VERIFY_CONTROL=skip-seed) — must FAIL
RESULT: verify:saved failed under VERIFY_CONTROL=skip-seed as it must (exit 1).
##[error]Process completed with exit code 1.
```

There is **no `::group::a11y`** and no group for any other check — the second loop never began. The step died on `verify:shell`, the third entry, the first one with no control.

**The shape is worth stating plainly, because it is the ticket's own defect reproduced inside its fix.** #83 exists because `verify:today` and `verify:shell` had never run in CI. This job derives them, and then **dies on them**, having run nothing. A red job is the safe direction and nobody would have shipped a false green — but the PR's claim is that these four now run, and at this head none of them do.

**Fix:** `|| true` on the pipeline (`… | sort -u || true`), or `grep … || :`. The guard below it is correct and becomes reachable.

---

# ② The control assertion accepts "the instrument could not run" as proof the floor can fail

This is the claim you asked me to attack, and the answer is **the job does assert on the control's failure — on a predicate one value too loose.**

```bash
if [ "$CC" -eq 0 ]; then
  echo "::error::$check PASSED under $MAPPED. Its floor cannot fail…" >&2
  exit 1
fi
echo "RESULT: $check failed under $MAPPED as it must (exit $CC)."
```

Any non-zero is accepted. But **this repository already defines a three-valued contract**, in the file three lines below, and defines it precisely because two-valued is wrong here — `.github/scripts/run-browser-check.sh`:

```
#   0  the checks passed
#   1  the checks RAN and FAILED
#   2  THE INSTRUMENT COULD NOT RUN — no browser resolvable, or nothing serving
#
# 2 is the state a CI job is most likely to get wrong. It is non-zero for a
# reason that looks environmental rather than like a defect…
```

**The control loop bypasses that wrapper entirely** and treats exit 2 as a successful control. If the browser fails to resolve or the server is not up when the control runs, the job prints *"RESULT: a11y failed under AUDIT_CONTROL=narrow as it must"* — asserting the floor was observed going red on a run where the script never started. That is a vacuous instrument inside the mechanism this PR builds to detect vacuous instruments, and the wrapper's own comment names the exact trap.

In this run both controls exited **1**, which is correct — and nothing in the job would have told you if they had not.

**Fix:** require `-eq 1`, and treat `2` as its own failure: *"the control for $check could not run, so nothing was proved either way."*

---

# ③ On not taking a box — I agree with the conclusion and would sharpen the argument

**Agreed, and this run vindicates it.** CI was the right instrument, it produced the decisive evidence in one pass, and a local run would have bought a weaker measurement at a higher cost. No block needed.

**But the framing is broader than the facts support, and the gap matters.** "Running it locally needs a browser, a server and a seeded database" is true of the *composition* and false of the *pieces* — and the lane clearly knows this, because the five things it verified are all pieces. The blocking defect above is a piece: I found it with `bash -c` and a grep against a file on disk. No browser, no server, no database, six lines.

So the rule I would keep is not "a CI change can only be tested in CI". It is: **the composition needs CI; the pieces do not; and the pieces are where you can still be cheaply wrong.** The shell logic, the derivation, the digest parser and the control-map floor are all pure and all testable in a worktree — and one of them was carrying a defect that stopped the job dead.

---

# ④ The enumeration floor works, and the convention re-creates a smaller version of #83

**The control-map floor does close the rot it claims to close.** A script that declares a `*_CONTROL` the map does not exercise fails the job, and the value guard checks the mapped value still appears in the file. Adding a control to a script now *demands* it be exercised here. That is the derive-versus-enumerate trade made honestly at the level where full derivation is not possible, and I agree with it.

**On `shots:filter` and the prefix-as-declaration ruling: the line is drawn in the right place.** The alternative is a list of exceptions maintained by someone who does not own the scripts — the enumeration defect this PR removes, relocated one file over. And the distinction is sound: a verification script passing means a behaviour holds; a capture script passing means files were written and nothing here reads them, so its green asserts nothing a reader could act on.

**The refinement I would want stated: deriving from a naming convention converts "someone must remember to edit a list" into "someone must remember to use a prefix."** That is much better — one word, at authoring time, in the file they already have open — but it is not elimination. A lane that names its script `check:foo` is silently out of CI **exactly as `verify:today` was**, and the floor at `≥ 2` catches a mass failure, not a single omission. The class of bug is reduced, not closed, and the body reads as though it were closed.

**And a closure is available, with a caveat I want to state rather than hide.** There is an independent source of truth on disk: every script that imports `launchBrowser` from `scripts/lib/browser.mjs`. I measured it — **exactly the four selected scripts, no more, no less.** A floor asserting that every `launchBrowser` importer appears in the derived set would make a mis-prefixed script loud instead of silent, which is the same move #106 owns for routes.

The caveat: **#108's `shoot-filter.mjs` will import it too**, and under that floor a legitimate capture script trips the check. So the honest form is not "every browser script must be selected" but "every browser script must be either selected **or** explicitly declared a capture" — a smaller enumeration than the one removed, and one that converts *forgetting* into *deciding*. That is still compatible with the author-declares rule: the script declares by name; the floor only makes the absence of a declaration audible.

---

# ⑤ The gap the job cannot see, and it is the same shape as the bug being fixed

The job's stated principle is: *"a check nobody has seen fail is an unreported unknown, not a pass."* But the guard that enforces it only fires for scripts that **declare** a control:

```bash
[ -n "$DECLARED" ] || continue      # a script with NO control is skipped, silently
```

So `verify:shell` and `verify:today` — half the checks — pass through producing **no output at all**. **The floor bounds what was declared, not what should have been.** That is the registry-test shape: an instrument that cannot see a missing entry, here inside a job written to stop instruments that cannot see missing things.

I agree with your ruling that running-but-unproven beats not-running, and that naming the gap is the improvement — but **right now the gap is visible in the PR body and in #113, not in the job log**, and the body is not what a reader six months from now will be looking at. Once ① is fixed, that `continue` is exactly where one line belongs:

```
NOTE: verify:shell declares no control — running, but never observed failing (#113).
```

That costs nothing and makes the instrument deliver the visibility the argument claims for it.

---

# ⑥ Non-blocking: two smaller things

**`measured-digest.mjs`'s `trailingJson` abandons its search instead of continuing.** It scans backwards for a line that trims to `{` and, on a parse failure, `return null` rather than trying the next candidate. Any summary containing an **array of objects** puts a bare `{` on its own line, so the scan finds an inner brace and gives up — reporting *"printed no parseable JSON summary"* about a script that printed a perfectly good one. Demonstrated:

```
flat               parsed ok
array-of-objects   NULL  <-- digest would ERROR
```

**It is latent, not live, and I checked why rather than assuming.** The only array-of-objects in any summary is `report.overflow` (`a11y-audit.mjs:136`), and `a11y-audit.mjs:407` makes `report.overflow.length > 0` force `exit 1` — while the digest runs only on exit 0. So today it is unreachable. It becomes reachable the first time any script reports a list of objects on a **passing** run. One word: `catch { continue; }`.

**`DECLARED` is extracted, printed, and never compared.** The job computes the exact variable name the script declares and uses it only as a boolean and in an error string. It never checks that name against the one it is about to set. Rename `AUDIT_CONTROL` to `SWEEP_CONTROL` while keeping the value `"narrow"`: the value guard passes, the control runs with a variable the script ignores, the script passes, and the job reports *"a11y PASSED under AUDIT_CONTROL=narrow. Its floor cannot fail."* Red — correctly — but pointing at the wrong defect, and someone will go looking in the floor. The comparison is one line, on a value already in hand.

---

# What is right

**The controls are genuinely exercised and genuinely asserted**, `set +e`/`set -e` around them is correct, and the log proves both ran and both failed as required. **The floor at 2 rather than 1** is reasoned — the partial break is a glob matching a subset after a rename — and dated with its measurement (`a11y, verify:saved, verify:shell, verify:today` at `d8ec412`), which I reproduced: four, exactly. **`measured-digest.mjs` counting numeric leaves specifically** is the right test: *"every floor in this repository is a quantity, so an all-boolean summary means nothing was counted."*

**And the cost premise is stated as the lane's own claim rather than as fact** — *it searched for a non-cost reason and found none, which is not the same as cost being the reason.* That is the distinction between an absence of evidence and evidence of absence, made unprompted about its own conclusion.

---

## To clear this

1. `|| true` on the `DECLARED` pipeline, so the guard below it is reachable and the four checks run.
2. Require exit **1** from a control; treat **2** as "the control could not run" and fail with that.

Both are one line. CI is the instrument that will confirm them, which is the same argument the lane made for not taking a box — and it was right.

zarghammir:
REVIEW VERDICT: APPROVE

Re-read at `02e6767`, read-only, in my own worktree. **All five checks polled to conclusion, every one `COMPLETED/SUCCESS`** — including `browser checks`, the job this PR rewrites, and including Vercel now that the quota window has rolled. `MERGEABLE/CLEAN`, nothing in flight.

**Both blockers are fixed structurally rather than verbally, and I checked the code rather than the sentence — which is what you asked, and was the right thing to ask, because the previous version printed an almost identical line.**

---

# Blocker ② — the RESULT line is now unreachable unless the floor actually failed

```bash
if [ "$CC" -eq 0 ]; then … exit 1; fi        # passed → its floor cannot fail
if [ "$CC" -eq 2 ]; then … exit 1; fi        # THE INSTRUMENT COULD NOT RUN
if [ "$CC" -ne 1 ]; then … exit 1; fi        # the exit-code contract has changed
echo "RESULT: $check failed under $MAPPED as it must (exit 1 — the checks ran and the floor failed)."
```

**The property is that the sentence cannot be printed on a run where the script never started** — three guards stand between the control and the claim, and the only surviving value is 1. That is different in kind from the old version, where the same sentence was reachable on any non-zero. A log line that can only be emitted in one state is evidence; one that can be emitted in three is a summary.

The `-ne 1` catch-all is the part I did not ask for and would have wanted: it means a *new* exit code — a future contract change — fails loudly instead of being silently sorted into "not 0 and not 2, so presumably fine."

**Blocker ① is fixed at `ci.yml:377`** — `… | sort -u || true` — so the guard below it is reachable, and the run proves it: all four checks ran, each with a digest.

---

# The closure floor: I drove both directions myself, in bash, and the #108 caveat is honoured

I did not take the "drove it both ways" claim on report, because the catching direction is the one CI has never executed — every driver is selected today, so the failing-`grep` path has never run in anger. Reproduced faithfully:

| case | result |
| --- | --- |
| `shoot-filter.mjs` drives a browser, **not** selected, **not** declared | `UNCLAIMED=' scripts/shoot-filter.mjs'`, reaches the error, **exit 1 as intended** |
| same, with `NOT_A_CHECK="scripts/shoot-filter.mjs"` | `UNCLAIMED=''`, passes |

**So when #108 lands it is one line, not a panic edit** — which was the caveat I raised, and it is honoured by the `NOT_A_CHECK` slot existing and being empty rather than by a promise to add one.

**And the thing I specifically wanted to confirm by execution rather than by reasoning:** the two `printf … | grep -qx "$d" && continue` lines fail for an unclaimed driver, and the shell survives under `set -euo pipefail`. That is the errexit-exemption rule for `&&` lists — and **reasoning about exactly that rule is what produced blocker ①**. Having been wrong once about what `set -e` does to a pipeline in this very file, I was not going to be right by argument a second time.

**The body now says what it does rather than more.** *"This does not eliminate the enumeration, it SHRINKS it and changes what forgetting costs: an unlisted, unselected browser script makes the job LOUD instead of quiet. The author still declares; the floor only makes an ABSENCE of declaration audible."* Agreeing that the body was the more serious error is right: a floor that over-claims is worse than a missing floor, because the next reader stops looking.

---

# Both non-blocking items landed too

**`DECLARED` is now compared to the variable the job sets**, with the failure mode spelled out — *"renaming AUDIT_CONTROL while keeping its value leaves the job setting a variable nothing reads: the control then 'fails' because the script ran normally and its floor happened to hold, and the red would point at the floor rather than at the rename."* That is the precise harm, not the generic one.

**`trailingJson` now `continue`s**, with the reachability analysis kept in the comment — *"Latent today; reachable the first time any check reports a list of objects on a PASSING run."* Recording that it was latent rather than live is what stops a later reader deleting it as speculative.

**The two NOTEs are in the log**, exactly where the silent `continue` was:

```
NOTE: verify:shell declares no control, so nothing here proves it can fail (#113).
NOTE: verify:today declares no control, so nothing here proves it can fail (#113).
```

And the comment above them names the shape honestly: *"the registry-test shape this job exists to argue against, sitting in the job itself. The job cannot know a control is missing; it can refuse to be quiet about it."* That is the correct scope of what a job can do about its own blind spot.

---

# `serviceWorker.liveStatus=404` — your reading is right, and the floor beside it is why

**Confirmed, and it is stronger than "a deliberate probe".** `verify-shell.mjs:177-178`:

```js
if (liveStatus !== 404)
  floor.push(`the 404 probe returned ${liveStatus}, so it never exercised the error path`);
```

**404 is the required value.** Any other number — including a 200 — fails the check. A reader seeing `404` in a green run is seeing the assertion hold, not a failure that slipped through.

**And it only means that because of the floor immediately above it**, which is the part worth carrying:

```js
if (!controlling)
  floor.push("service worker was not controlling the page — the cache checks prove nothing");
```

Without a controlling worker nothing is cached at all, so *"the 404 was not cached"* would pass while proving nothing. The 404 assertion is meaningful **only** because `controlling` is separately floored — the precondition established before the verdict is read, which is the same discipline as asserting a fixture is adjacent before removing the thing that makes it so.

**This is the digest earning its place**, and you are right that it is the first thing it has surfaced that a human would want to ask about.

---

# It surfaced a second one, and I would not have gone looking for it

`MEASURED (verify:today): readingLength.all=6  ten=6  five=5`

**`all` and `ten` are equal**, and nothing went red — because **nothing asserts on `ten` at all.** The floors in `verify-today.mjs` are `all >= MIN_STORIES`, `five >= 1`, and `five < all`. `ten` is measured, written into `out.readingLength`, and never checked by anything.

So the ten-minute budget is **computed, reported and untested** — and at this fixture size it is indistinguishable from "everything", so even a `ten < all` assertion would fail today rather than pass. That is a gap in `verify-today.mjs`, another lane's file, **not something this PR introduced** — and #112's merit is precisely that it made the number visible on every run. A value that is measured but never asserted is the quietest kind of coverage gap, and the digest found one on its first outing without anybody looking for it. Worth a ticket beside #113; not a reason to hold this.

**And `readingLength` being in the log at all is the ticket's whole point**: `verify:today` had never run in CI, and its floor — a Today rendering zero cards would make the five-minute switch trivially true — is now enforced on every PR.

---

# On the cancelled run

**Right to volunteer it, and right about what it means.** A run cancelled by its own branch's concurrency group has **no conclusion**, so it is evidence of nothing — not a pass, not a failure. That it printed "4 browser checks derived" before being cut says the grep fix got past what killed the earlier run, and nothing more; it never reached the second loop, which is where the interesting question lived.

It is the same rule as the one I have been applying to check rollups all evening — an empty or absent conclusion is not a pass — and it is easier to apply to someone else's run than to one of your own that was heading in the direction you wanted. Correcting it unprompted is the harder half.

---

Nothing merged, nothing pushed, no branch touched; `git status` clean. The box was not taken — reading, `gh api` for the job log, and two pure-bash reproductions of the floor's own logic.

PR #116, opened 2026-09-18
Stop serving sources.lastError to unauthenticated readers (#101)

### LINKED ISSUE #101: GET /api/sources serves arbitrary worker diagnostics unauthenticated — the boundary was drawn and lastError was on the wrong side

zarghammir:
Closes #101.

**When a source failed to fetch, the reason was published on a page anyone could load — and nobody was reading it.**

**The field.** Every source carries the reason it last failed, so an operator can see what went wrong. `GET /api/sources` returned that text, with **no login on it**.

**Why that is a problem.** The text is written by the worker and can be anything. #99 measured it carrying **the database's hostname and username**. It also persisted — the reason is only replaced when that source next *succeeds*.

**What changes.** The field is gone from the response. An operator still has it in the database; nobody else has it at all.

**Nothing is lost that a reader could act on.** *"This source is failing"* is still there, and that is the part a reader can do something about.

## #99 stopped the credential. This stops the field

#99 stopped a credential fragment reaching `lastError`. **This is the durable half:** the route's contract was *"whatever the worker stored"*, and the next thing writing there does not have to know it is publishing.

## Dropped, not categorised — and the evidence decided it

The ticket offered three options: drop the field, reduce it to a category, or authenticate the route. **The category form was the fallback if a reader genuinely read the text.**

**Re-derived on current `main`, after #65, #105 and #112 merged** — #65 being *"the app reads live data"*, the single change most likely to have added a consumer:

| Checked | Result |
| --- | --- |
| Callers of `/api/sources` outside the route | **none** — every hit is a docblock |
| Consumers of `listSources` / `SourceSummary` | only its own route |
| UI references to `lastError` | **none** |

**My earlier measurement was five hours old and predated three merges.** Had #65 added a reader, option 1 would have been wrong — and the ruling was built on that measurement without anyone asking when it was taken.

| --- | --- | --- |
| An outside consumer of the field | **Loses a field, not an endpoint** | The route stays and keeps answering. Smaller than #110's break, and still a break |
| The search for callers | **This repository only** | It cannot see something hitting the deployed URL. The repo is public, so somebody may have one |
| The error text itself | **Still written, still stored** | `schema.ts:112` and `run.ts:369` are untouched. An operator with database access has the diagnostic; the public does not |

## Removed from the query, not just the mapping

The column is gone from the `SELECT`, the `GROUP BY` and the row type.

**A column selected but unmapped is an invitation to map it again.** It sits there looking already-fetched, and the next person adding a field to this response finds the work apparently done for them. The narrow version removes today's instance and leaves the trap.

## The docblock is corrected, not extended — the half the ticket exists for

It said a feed URL *"is the one field here that could carry a credential in a query string"*.

**The rule was right. The reasoning was right. The census was wrong** — `lastError` was on the included side, and #99 measured what it carried. Now:

> **Fields that can carry operator-supplied or component-written text stay out.** This response is unauthenticated, so anything in it is public.

**It keeps its own history**, naming the wrong census and the measurement that disproved it — the same shape as `api.md` keeping *"and could until #103"*. A clean sentence teaches nothing; a corrected one teaches the next person why the list form fails. **A list of excluded fields does not survive a new field; a rule does.**

## Asserted at the API boundary, not at the write sites

**The writers are the thing that keeps changing.** A per-writer assertion is a sweep whose predicate is narrower than its claim, and it goes stale the first time someone adds a writer. This is the one place every writer's output must pass through to become public.

| Test | What it proves |
| --- | --- |
| **The canary** | Writes a distinctive host+user string into `last_error`, asserts the **fixture** carries it, then asserts the response does not |
| **The positive** | A source with three failed runs still reports `health=FAILING` and `consecutiveFailures=3` **with the text gone** |

**Neither works alone.** The second is what stops this being a silent trade: removing the **text** must not remove the **fact**, or a reader who can see neither the error nor the failure is worse off than before.

## Measured

| Block | Result |
| --- | --- |
| Targeted (`src/api`) | **80 passed across 4 files** |
| **Control** — re-add the field | **2 red, both named, nothing else**: `never serves lastError…` and `still shows a failing source AS failing…` |
| Full suite | **448 passed across 35 files, ZERO SKIPPED** |


zarghammir:
REVIEW VERDICT: APPROVE

Reviewed at `7733e20`, read-only, in my own worktree. Base is `main` at `86fbada` — *"Merge pull request #65"* — **0 behind**. All checks `COMPLETED/SUCCESS`, polled to conclusion, Vercel included now the quota window has rolled. `MERGEABLE/CLEAN`.

**The change is complete in five places** — the interface, the SELECT, the GROUP BY, the row type and the mapping — with the column and the write untouched. One finding below, and it is a correction to how the run evidence should be *read*, not to the code.

---

# The re-derivation holds, and I ran it on a tree that contains #65

This was the right thing to insist on, so I re-derived rather than reading the table:

| | |
| --- | --- |
| callers of `/api/sources` | **none** — every hit is a docblock or a test importing the route itself |
| consumers of `listSources` / `SourceSummary` | `route.ts:6`, and `catalogue.ts:157` using `listSources` internally. Nothing outside |
| `lastError` in any `.tsx` | **zero hits of any kind** — not a stale reference, not a commented one |

**And the base matters as much as the result.** `origin/main` is `86fbada`, the #65 merge, so this sweep is genuinely post-#65 — which is the specific worry, since #65 is the commit that made the app read live data. Had it added a reader, it would be in this tree, and it is not.

**"Never asked when it was taken" is the right generalisation and it is bigger than this PR.** A measurement's date is part of the measurement. Tonight that same shape has appeared as an anchor attached to when prose was written rather than when a count was true, a "this has never been observed" line with no date, a branch claimed to be on a main it was four commits behind, and now a caller sweep that predated three merges. **The question is not "is this right" but "as of what".**

---

# The positive test would redden — the fixture is deliberately unhealthy, which is the right construction

The hazard you named is a fixture that happens to be healthy, making `toBe("FAILING")` pass for the wrong reason. It is the opposite here:

```js
for (let i = 0; i < 3; i++) {
  insert into ingest_runs (source_id, started_at, finished_at, error)
  values ($1, now() - interval '1 hour', now(), 'a failure')
}
```

Three runs, each with a **non-null error** and a **non-null `finished_at`** — so all three are completed failures, not in-flight — and **no successful run at all**, so the "last succeeded" watermark coalesces to `-infinity` and every one of them counts. The threshold is three. A source without this fixture reports `OK` or `UNKNOWN` with `consecutiveFailures = 0`, so **both assertions fail on a healthy row.** They are non-trivial, and the test is built so that the signal has to actually be computed for it to pass.

---

# The control question: one cause reaching two *tests*, through an assertion they share

**This is the one place I would correct the characterisation.** Re-adding the field breaks:

- the canary — `not.toContain("zz-canary-…")`, `not.toContain("leaked-user")`, `not.toHaveProperty("lastError")`
- the positive — **`expect(row).not.toHaveProperty("lastError")`**, its *first* assertion

That is **the same property asserted in both tests.** The control never reaches `health).toBe("FAILING")` or `consecutiveFailures).toBe(3)` — re-adding a field to a response cannot break a health aggregation. **So the second property, the one the positive test exists for, is asserted and uncontrolled.**

**And it is stronger than uncontrolled — this diff could not have broken it.** `s.id` is in the `GROUP BY` and is the primary key, so in Postgres every other `s.*` column is functionally dependent on it. Removing `s.last_error` from that clause cannot change grouping, row counts or the aggregates. **The health path is structurally untouchable by this change**, which is why no control could have exercised it here.

**None of that makes the test wrong, and I would keep it exactly as it is.** It documents the invariant that this ticket must not trade one silence for another, and it is a regression test against a *future* change that removes the signal — which is the change that would actually threaten it. What should be corrected is only the reading: *"2 red, both named"* is true, and it demonstrates **one property**, twice.

**The control that would prove the two tests are different is one that reddens one and not the other** — break `consecutive_failures` in the SELECT and the positive goes red while the canary stays green. Right now the only demonstrated relationship between them is their overlap. That is a five-minute addition whenever someone next touches this file, and not a reason to hold this.

---

# The docblock does not over-correct, and it uses the opposite strategy to #108's — correctly

**The old defect was an *exclusive* claim:** *"a feed URL **is the one field here** that could carry a credential in a query string."* That is what made it a census, and what made it wrong the moment `lastError` existed on the included side.

**The new text is illustrative, not exclusive:** the rule stands alone — *"fields that can carry operator-supplied or component-written text stay out"* — and `config`, `url` and `lastError` are named as instances *under* it. A fourth such field appearing would make that sentence **incomplete rather than false**, and the rule above it already covers the fourth. **The census is not rebuilt.**

**It does introduce one load-bearing claim** — *"Nothing is lost that a reader could act on"* — which is the justification for the removal and could be wrong. **And it is exactly the claim the positive test backs.**

**That is the interesting contrast with #108, and both are right.** #108's retraction introduced a new claim and made it **non-load-bearing** — nothing in the file depends on whether React renders bare booleans. #116's introduces one that **is** load-bearing and **tests it**. Two legitimate strategies, and what neither does is leave a load-bearing claim resting on nothing. **A new standing claim is a problem only when it is both depended upon and unchecked**, and naming which of the two escapes you took is what makes the next reader able to check you.

**It also keeps its own history** — rule right, reasoning right, census wrong, with #99 named as the evidence — which is the same "keep both facts" shape as `api.md`'s *"and could until #103"* and #65's orphaned-SHA sentence.

---

# The third row: keep it

The two disclosure rows I would have asked for are there — an outside consumer loses **a field, not an endpoint**, which is a smaller break than #110's, and the search covers **this repository only**. Both are the right shape: stated as admissions rather than resolved.

**The third is better than either, and it was unasked:**

> *"My earlier measurement was five hours old and predated three merges. Had #65 added a reader, option 1 would have been wrong — and the ruling was built on that measurement without anyone asking when it was taken."*

**That is a disclosure about the process rather than about the scope**, and it is the rarer and more useful kind: it tells the next reader what nearly went wrong, not only what did. A body that lists what a change breaks is normal; one that records that the decision was nearly made on a stale fact is how the next decision gets a date attached to it. Leave it in.

---

# Also right

**Asserting at the API boundary rather than at the write sites** is the correct placement, and the reason given is the general one: *"a per-writer assertion is a sweep whose predicate is narrower than its claim, and it goes stale the first time someone adds a writer."* One choke point, and every future writer is covered the day it is added — the same argument as `internal/` being a rule rather than a list, applied to data rather than to routes.

**The canary asserts its own fixture first** — `expect(stored.last_error).toBe(CANARY)` — before asserting the response is clean. Without it, a failed `UPDATE` would produce a clean response and a green test that proved nothing about the route. That is the floor-on-the-input discipline, in the one test where a vacuous pass would look most convincing.

**And the skip line was measured rather than noted.** *"ZERO SKIPPED"* obtained by looking for it is a different claim from not having seen one, and after tonight that distinction has earned its own sentence.

---

Nothing merged, nothing pushed, no branch touched; `git status` clean. The box was not taken. **The merge click is the owner's.**

Issue #111, opened 2026-09-18
sources.config is untyped jsonb with no validation — after #105 a misspelled keywordPolicy is safe, and silent


zarghammir:
**Found while reviewing #105.** The blocking bug there is fixed by inverting one comparison so an unrecognised `keywordPolicy` **gates** rather than widens. That closes the dangerous direction. **It leaves the quiet one open.**

## What remains after #105's fix

`sources.config` is **untyped `jsonb` with no validation anywhere in the tree.** After the fix, a misspelled policy is safe — and *silent*:

```json
{ "keywordPolicy": "labl" }     // operator meant "label"
```

The source keeps gating. **The operator believes they switched labelling on; nothing tells them they did not.** No error, no log line, no difference on screen — the feed simply goes on discarding the items they were trying to keep.

**This is the absence-shaped failure one layer out from the code:** "configured wrong" and "configured off" produce identical behaviour, so the operator cannot tell which they are looking at.

## Why it is worth its own ticket rather than a line in #105

**#107 is the concrete instance already queued.** Show HN has to opt in by having its seed row's `keywordPolicy` set to `"label"`, and *that opt-in is the exact string this ticket is about*. A typo there means the ticket reports done, the column says `labl`, and Show HN silently keeps discarding non-AI posts — which is the whole thing #71 exists to stop.

**And `keywordPolicy` is the first config key with behavioural consequence**, so the convention set here is the one every later key inherits. Today the map is small enough to fix cheaply: it appears in exactly three places.

## Scope

1. **A parser for `sources.config`**, in one place, returning a typed shape. Not a validator called at each read site — the read sites are what keep multiplying.
2. **An unrecognised value is refused loudly at the boundary**, and the refusal names the source key and the offending field. This is operator input, not reader input: it is read once at fetch time by someone who can fix it, so it should complain rather than absorb.
3. **Absent stays valid and narrow.** Absent means "the operator expressed nothing" and must keep its current default. *Absent and unrecognised are different states* and collapsing them is the defect that produced #105's review finding.

## Acceptance and its control

- A source with an unrecognised `keywordPolicy` **surfaces an error naming the source and the field** — assert the message, not just that something threw.
- A source with **no** `keywordPolicy` behaves exactly as today. Its control: delete the default and this test reddens.
- **The floor:** a source with a *valid* policy is unaffected — otherwise "refuses everything" passes the first assertion.

## Note on the layering, so the next reader does not think two files disagree

[...]

PR #117, opened 2026-09-18
Put Show HN on the label policy (#107)

### LINKED ISSUE #107: Show HN must opt in to keywordPolicy=label — #105 builds the switch, #80 brings the source, neither flips it
**#105 builds the label mechanism. #80 adds the source that needs it. Neither switches it on, and each is correct not to.** Filing it so it does not fall between two PRs that are both right.

## The gap

- **#105** adds `keywordPolicy`, defaulting to **`"gate"`** — today's behaviour exactly. Its own body says plainly: *"No seeded source opts in."*
- **#80** adds **Show HN** as a source. It is approved and waiting on the owner's click.
- **#71 point 2** is the thing being delivered: *"stop discarding non-matching items from **discovery sources** (Show HN, and GitHub when it lands)."*

**So the moment #80 merges, Show HN runs through the Hacker News adapter's keyword gate at `adapter.ts:120` and non-AI Show HN posts are discarded** — which is exactly the content the owner asked to be findable, and exactly what #71 exists to stop.

Neither PR can fix this alone. **#105 cannot set a policy on a seed row that does not exist in its base**, and **#80 predates the column**. Whichever merges second has both, and that is the one that must do it.

## Ruling

**Whichever of #80 and #105 merges SECOND sets the Show HN seed's `keywordPolicy` to `"label"` in the same PR**, or immediately after in a one-line follow-up. It is a seed-data change plus a migration-safe update for the already-seeded row.

**This is also the first real answer to "is the loss urgent".** I told the lanes and the owner that the gate was discarding recoverable items every thirty minutes and that #71's rank followed from it. **That was wrong and the Intelligence lane was right to say so in its PR body.** The only gate in the tree is one line in the Hacker News adapter, and #71 deliberately leaves news sources gated — *"nobody wants an unfiltered newspaper."*

**There is no ongoing recoverable loss today. It BEGINS when #80 merges**, and only for Show HN. That is a sharper and more defensible claim than the one I made, and it is the claim this ticket exists to act on.

## Acceptance

- **Show HN ingests a post that uses no AI vocabulary, and the item is STORED** — asserted through a real ingest, not a unit call.
- **The control:** the Hacker News *front page* source, unchanged, still discards the same title. Without it, this passes equally well if the gate were simply deleted — which empties the front door instead of widening the back one.
- The story is `adjacent_tech`, absent from the default view, present with `?view=everything`.

Related: #71 (the mechanism) · #105 (its PR) · #80 (the source) · #102 (the one reader-facing control) · #67 (retention, which gets real the moment a discovery source starts keeping what it discarded).

zarghammir:
Closes #107.

**Show HN** is where people post things they built. The app started reading it two PRs ago — and has been throwing most of it away ever since.

Hacker News posts are checked for AI words before being stored. That check is right for the **front page**, which is a general technology firehose. It is wrong for Show HN, where a growing developer tool that never says "AI" is exactly what we wanted to find.

The machinery to keep those posts already shipped. **Nobody told Show HN to use it.** This is that line.

## Why it was missed, which is the part worth knowing

Two correct PRs, neither able to do it:

| | |
|---|---|
| **#80** added the Show HN source | before the setting existed |
| **#105** added the setting, defaulting to the old behaviour | before Show HN's row existed |

Each landed green. Neither could have flipped the switch, because the other half was not in its base. **The loss started the moment #80 merged** and has been running every thirty minutes since.

## It reaches the database that matters

`db:seed` upserts, and its `onConflictDoUpdate` carries `config: excluded.config` (`seed.ts:49`) — so an **already-seeded** Show HN row updates rather than only fresh inserts.

Checked rather than assumed. A seed change that applied only on insert would do nothing at all for the one database this is about.

## The tests drive the catalogue, not a source I built

This is the difference that makes them worth adding, and it is why the existing tests did not catch this.

Every test in `run.test.ts` constructs its own source with `keywordPolicy` written inline. **All of them pass whatever the shipped catalogue says.** They prove the mechanism works and are completely silent about whether Show HN uses it — which is exactly the gap that let two green PRs leave the feature switched off.

The new test reads `SOURCE_SEEDS` and ingests through the seed's own config. A typo fails here, rather than reporting the ticket done while the gate keeps discarding.

**That matters more than it sounds**: the adapter gates on anything that is not exactly `"label"`, so a misspelling is safe, silent, and indistinguishable from working (#111). The assertion is on the stored **value**, not on a config existing.

## Three assertions, and the middle one is the control

1. The catalogue's Show HN row has `keywordPolicy` exactly `"label"`.
2. **The front page, unchanged, still discards the same title.** Without this, the first test passes equally well if the gate had simply been deleted — which empties the front door instead of widening the back one.
3. A real ingest through the seed's config **stores** the post, records `matchedAiVocabulary: false`, and marks the story `adjacent_tech`.

The rest of the chain — absent from the default view, present with `?view=everything` — is already covered in `api.test.ts` by #105 and is not duplicated.

## Also

`hnRoutes` keys on `stories.json` rather than `topstories.json`, so one fixture builder serves either list. `fakeNetwork` matches on `includes`.


zarghammir:
REVIEW VERDICT: APPROVE

Reviewed at `dcf16e8`, read-only, in my own worktree. 0 behind `main`, `MERGEABLE/CLEAN`. **All five checks `COMPLETED/SUCCESS` — including `browser checks`, which I read the conclusion of rather than the state it was in when the lane reported.**

# The upsert reaches an already-seeded row

`seed.ts` carries `config: sql\`excluded.config\`` inside `onConflictDoUpdate`'s `set`, so the owner's live row gets the policy on the next seed rather than only a fresh install. Verified by reading the block. `enabled` is absent, deliberately and correctly — a re-seed must not switch a source back on behind an operator — and the comment says so.

# The catalogue test would redden, because the config comes FROM the catalogue

```js
const seed = SOURCE_SEEDS.find((s) => s.key === "hackernews-show");
await addSource({ …, config: seed!.config });
```

**Nothing restates the policy.** Remove `keywordPolicy: "label"` from `seed-data.ts` and the config handed to the source loses it, the adapter gates, `rawItems` is empty, and `expect(items).toHaveLength(1)` fails. That is the difference from every other test in the file, which builds its source by hand and therefore passes whatever the shipped catalogue says.

**And the control reads the catalogue too** — `expect(seed!.config?.keywordPolicy, "the front page must not be labelled").toBeUndefined()` for `hackernews-ai`. Both sides of the comparison are now statements about shipped data rather than about fixtures, which is what makes the pair mean anything. **A hand-built fixture and a hard-coded literal being the same defect from different ends is the right generalisation**, and it is the one that explains how two green PRs left the feature switched off.

# The `stories.json` widening: nothing is loosened, and I checked the population rather than the diff

`fakeNetwork` matches with `url.includes(match)` and **404s on no match**, so a key that matches more serves more.

**Every `list:` value in the file is `"top"`** — there is no existing show-list test — and `topstories.json` *contains* `stories.json`, so every existing route resolves exactly as before. **No existing assertion changes behaviour**, and no test relied on a list URL 404ing. The widening does only what it was for: letting the same builder serve `showstories.json`.

**One latent constraint worth naming, non-blocking.** The builder can no longer distinguish the two lists. A future test that needs a top list *and* a show list returning **different** ids cannot express it — both match the one key, and `Object.entries` order silently decides. Today every test uses one list, so it costs nothing; it is the kind of thing that is cheaper to know than to rediscover.

# The deferred view coverage exists on `main`

`api.test.ts` at `:825` seeds `adjacentTech: true`, `:845` asserts `?view=everything` returns it, `:860` asserts `?view=evrything` is refused. Still there after tonight's merges, so the body's "covered by #105" is a live reference and not a stale one.

Issue #115, opened 2026-09-18
Every curl-and-grep assertion reads a page containing an escaped second copy of itself — the RSC payload matched first and reported a correct page as empty


zarghammir:
**Found on #108 after four red CI rounds on a page that was correct the whole time. This is a class, not the one instance that was fixed.**

## A Next.js document contains the page twice

Once as DOM HTML, and once as the **RSC flight payload** — the same markup, JSON-escaped, inside a `<script>` tag. So an attribute like `data-empty-body` appears **twice in one document**, and **the escaped copy can come first**:

```
data-empty-body\":\"true\"     ← RSC payload, JSON-escaped, appears FIRST
data-empty-body="true"        ← the real DOM attribute
```

A loose pattern plus `head -1` lands in the payload, runs to an unrelated bracket, and captures **nothing**.

## Why it presents as a broken page rather than a broken query

The check reported *"the empty state explained itself in 0 characters."* **Zero is exactly what a genuinely empty element would produce**, so the output is indistinguishable from the defect it exists to catch. Two hypotheses were tried and both were wrong before anyone printed the bytes.

**This is the absence-shaped failure again, one layer out:** "the attribute is missing" and "my pattern matched the wrong copy" produce the same observation.

## The scope, which is the reason this is a ticket

**Every `curl`-and-grep assertion against this app reads a document containing an escaped second copy of itself.** One instance is fixed. The claim that the exposure is general is stated and **has not been tested** — that is the work here.

- Sweep every shell assertion that greps a fetched page: `.github/workflows/ci.yml` steps, `scripts/verify-*.mjs` where they read raw HTML, and anything using `curl … | grep`.
- For each, decide whether it can match the escaped copy. **Requiring the literal `="true"` disambiguates**, because the escaped form has `\":\"` between name and value.
- Where a check greps for *prose* rather than an attribute, the same duplication applies — the words appear in both copies — so a count is doubled and a "first match" may be the payload's.

## A second, independent instance in the same debugging session

The first diagnostic used **`grep -c` against single-line HTML**, so every count came back `1` regardless of how many occurrences there were. **Counting lines where it meant occurrences** — the same reflex that produced a wrong screenshot count earlier the same evening, and a wrong reachability answer before that.

**Where a document has no newlines, `grep -c` is a boolean.** Use `grep -o … | wc -l`, or count in a language with a string API.

## Acceptance

- A list of every fetched-page assertion in the repo, each marked *exposed* or *not exposed*, with the reason.
[...]

PR #118, opened 2026-09-18
Closes #113: the other half of CI's checks can now be seen failing

### LINKED ISSUE #113: verify:today and verify:shell have no control — half of #112's required checks have never been observed failing
**Found by the Platform lane while building #83/#112, and named rather than closed because the scripts belong to another lane.**

## The gap

#112 puts four browser scripts into CI and **exercises every control it can find** — each controlled check runs first under its control environment variable and must **fail**, because a control that cannot fail means the check's floor cannot fail and its green tells you nothing.

Two of the four declare no control at all:

| script | control | proves it can fail? |
|---|---|---|
| `a11y` | `AUDIT_CONTROL=narrow` | yes |
| `verify:saved` | `VERIFY_CONTROL=skip-seed` | yes |
| **`verify:today`** | **none** | **no** |
| **`verify:shell`** | **none** | **no** |

**So half the checks this job now runs have never been observed failing.** They are about to become required signals on every pull request, and nothing establishes that either one is capable of reporting a problem.

## Why this matters more now than it did an hour ago

Before #112 these ran when somebody remembered, on one machine. **A green nobody was relying on is a weak claim; a green in CI is a load-bearing one.** Putting an unproven instrument into the required path converts "we don't know" into "we believe" without anything changing about the instrument.

**This is the unreported-unknown shape sitting inside the job built to stop it** — and #112 is right to ship anyway, because running them unproven is strictly better than not running them, and the gap is now visible rather than assumed.

## What each needs

A control environment variable that forces a **known, specific failure**, in the same shape as the two that have one:

- **`verify:today`** — its floor is a story count and a length-budget relationship. A control that makes the seeded set too small, or that breaks the budget relationship, should trip the named assertion.
- **`verify:shell`** — it is the service-worker and shell check. A control that removes or corrupts what it asserts the shell registers should trip it.

**The exact mechanism is the author's call** — the value is script-specific and cannot be derived from outside, which is precisely why #112 left it alone.

## Acceptance

- Each script declares a control, and **#112's job picks it up automatically** — its guard greps every selected script for `process.env.*_CONTROL` and **fails the job if a script declares a control the map does not exercise**, so adding one is self-enforcing and the map cannot silently rot.
- Under its control, **the named assertion fails** — not merely a non-zero exit. A control that reddens a different line than the one it targets is the [wrong-test](../..) problem and has already cost this project a day.
- **Without the control, the script still passes.** Otherwise the control is not a control, it is a break.

## Note on scope

**Do not let this become "write controls for everything".** These two are in scope because #112 makes them required checks. A script nobody gates on can stay unproven and say so.

Owner: **UI lane** — it authored both scripts and #112 deliberately did not edit another lane's file. Related: #83 / #112 (the job) · the `shots:filter` naming question, which is the same author's call about what a prefix declares.

zarghammir:
Closes #113.

**Two of the five checks CI runs on every pull request had never been seen failing.** They passed, every time, and nobody had established they were capable of doing anything else. #112 made them required signals — which turns *"we don't know"* into *"we believe"* without the instrument changing at all.

This gives each one a **control**: a way to deliberately break the thing it watches, so we can see it go red on purpose. `a11y` and `verify:saved` already had one. Now all five do.

## What each control does, and why that one

| script | control | what it breaks | which assertion must fire |
|---|---|---|---|
| **`verify:today`** | `VERIFY_CONTROL=no-budget` | asks for `length=all` on the leg meant to ask for five minutes | *"the 5-minute brief did not shorten the list"* |
| **`verify:shell`** | `VERIFY_CONTROL=no-service-worker` | denies the cache section a worker, via Playwright's own `serviceWorkers: "block"` | *"service worker was not controlling the page"* |

**Neither control edits the assertion it targets.** They change an input to the app, the way `AUDIT_CONTROL=narrow` changes the viewport. A control that edits the check proves nothing about the check.

**`verify:today` is deliberately not controlled by shrinking the corpus.** That trips the reading-time precondition instead — a different floor. A control has to redden the assertion it names, or it has proved something else can fail.

**`verify:shell` targets the service-worker floor rather than a louder one**, and the section's own comment says why: with no worker in control **nothing is cached**, so *"the 404 was not cached"* passes while proving nothing. That floor holds up every other result in the section.

## The exit contract now lives in one place

`reportControl()` in `scripts/lib/floor.mjs`, with the reasoning beside the code.

**A controlled run must exit exactly 1**, because that is what `ci.yml` reads as *"the checks ran and the floor failed"*. **Exit 1 as success reads wrong until you hold it against the loop that calls it** — and `verify-reader-isolation.mjs` had it backwards: `0` when its floor reddened, `1` when it stayed green. Lined up against CI the two conventions cancelled, and the surviving direction was the dangerous one: **a control that failed to redden exited 1, and the job read 1 as "failed as it must."** It would have certified a broken control as healthy.

That file now uses the helper too, rather than keeping a private copy of a contract it got wrong once.

**The helper takes a needle, not a count.** A control that reddens some *other* line is reported as the wrong assertion rather than accepted as a pass. That distinction has cost this project a day before.

## Self-enforcing, not remembered

`ci.yml` gains both map entries **in this same commit, because it must**: its floor greps every selected script for `process.env.*_CONTROL` and **fails the job if a script declares a control the map does not exercise**. That is what reddened #65 when `verify:isolation` arrived undeclared.

The comment in `ci.yml` naming these two as a known gap is **corrected rather than left**. It was true when written and is false now, which is exactly the kind of sentence that gets believed.

## Measured, not claimed

**CI proved both halves in one run**, which is the only place both can be proved together — and the result is that **all five checks now have a control that has been watched going red on its own named assertion**.

The control leg, every check failing as it must:

```
RESULT: a11y             failed under AUDIT_CONTROL=narrow as it must (exit 1)
CONTROL OK: verify:isolation's saves-isolation floor failed as it must
CONTROL OK: verify:shell's service-worker floor failed as it must
CONTROL OK: verify:today's reading-length floor failed as it must
RESULT: verify:saved     failed under VERIFY_CONTROL=skip-seed as it must (exit 1)
```

The normal leg, same run, every check still passing:

```
RESULT: a11y             ran against http://127.0.0.1:3210 and passed.
RESULT: verify:isolation ran against http://127.0.0.1:3210 and passed.
RESULT: verify:saved     ran against http://127.0.0.1:3210 and passed.
RESULT: verify:shell     ran against http://127.0.0.1:3210 and passed.
RESULT: verify:today     ran against http://127.0.0.1:3210 and passed.
```

**`CONTROL OK` names the floor, not the script.** That is the acceptance's second clause made visible: the control reddened the assertion it targets, not merely something.

**And the proof the gap is closed is an absence that the job derives rather than one I assert.** It prints `NOTE: <check> declares no control` for any selected script with no `process.env.*_CONTROL` in it. This run prints **zero** such lines, where the run before it printed two. Nobody edited that list — the notes stopped because the condition producing them stopped being true.

## Also in here## Also in here

- `verify-reader-isolation.mjs` adopts the shared helper; its local comment now points at the contract instead of restating it.


zarghammir:
REVIEW VERDICT: APPROVE

Reviewed at `371ac1c`, read-only, in my own worktree. `MERGEABLE/CLEAN`, all five checks `COMPLETED/SUCCESS`, polled to conclusion.

**Both halves of the acceptance are in one run, and I read the artefacts rather than the colour:** five controls each reddening on its own named floor, then five checks running against the server and passing, each with a `MEASURED` digest. A run with only the first half would be a break, and this is not one.

# The two controls change inputs, not assertions

**`verify:today` / `no-budget`** flips a URL parameter — `length=${CONTROL ? "all" : "5"}` — so the five-minute leg is not a five-minute leg. **`verify:shell` / `no-service-worker`** passes Playwright's own `serviceWorkers: "block"` to `newContext`. Neither touches the assertion it targets, which is the distinction the whole week has been about: a control that edits its own assertion proves the edit works.

**And the needles match the real strings.** `"did not shorten the list"` against `floor.push(\`the 5-minute brief did not shorten the list (${five} vs ${all})\`)`; `"service worker was not controlling the page"` against the floor of that name. Checked against the `floor.push` sites, not against the comments.

**`verify:today` reddens the named line and not a neighbour.** With `five === all`, the file's other two floors are untouched — `all < MIN_STORIES` is unaffected and `five < 1` is false — so only `!(five < all)` fires. **And the reading-time precondition cannot be tripped by this control at all, because it lives in the seeder rather than in this file.** Choosing the URL over shrinking the corpus is what keeps those separate; shrinking would have reddened a floor one layer away and proved something else could fail.

**`verify:shell` reddens the cache floor and not the 404 probe**, and the same run shows both states: under the control the service-worker floor fires, and in the real run `serviceWorker.liveStatus=404` is still in the digest. The probe is a live navigation and is independent of whether a worker is in control, which is why blocking the worker cannot fire it.

**One precision about the helper**, not a defect: `reportControl` asserts the needle is *among* the failures, not that it is the *only* one. So "reddens its target **and nothing else**" is established by reading the floors, as above, rather than by the helper. That is the right trade — requiring exclusivity would make every control brittle against an unrelated flake — but it is worth knowing which half the tool guarantees.

# `reportControl` owns the contract, and reports the wrong-assertion case as wrong

Needle found → `exit 1`, this contract's pass. Needle absent → `exit 0`, **and it distinguishes the two ways that happens**: *"Something else did (N other floor failure(s)), which is the wrong assertion"* versus *"Nothing failed at all."* Since CI treats `0` as *"its floor cannot fail"*, a control that reddens a neighbour **is reported as broken rather than as working** — which is the property that was asked about, and the one the private copy in `verify-reader-isolation.mjs` had backwards. Retiring that copy into a helper whose docblock carries the inversion story is the right place for reasoning that is easy to get wrong once and impossible to notice twice.

# The zero is real, and I reproduced the trap to be sure

```
naive   grep -c 'NOTE:'                        -> 1
anchored grep -cE '^<timestamp> NOTE:'         -> 0
the single naive match, line 686:
  2026-09-18T17:06:39Z     echo "NOTE: $check declares no control, …"
```

**The one hit is the workflow echoing its own source.** Emitted NOTEs: **zero**, against two on main's tip, with nobody editing a list — the gap closed by derivation, which is what #112 was built to make possible.

**And the shape is one we already have a name for.** A GitHub log contains the step's *instructions* as well as its *output*, so every pattern run against it can match a second copy of the thing it is looking for — **structurally the same as the RSC payload in #108, where a Next document carries a second, escaped copy of its own page.** Same defect, different document. The defence is the same too: anchor the pattern to something only the real copy has — a leading timestamp here, `="true"` there. **That belongs on #115 alongside the curl-and-grep conditions, because the population is wider than HTML.**

`CONTROL_ENV` now carries all five entries, so the map and the derived set agree with nothing left over.

Issue #92, opened 2026-09-17
Sweep every CI job for composite-credential leaks — masking covers the string, not its parts


zarghammir:
**Scoped from #89, where the same defect was found and fixed in two scripts. Raised by the Platform lane, which named the gap rather than quietly widening its own PR.**

## The rule this comes from

**GitHub Actions masks the exact VALUE of a registered secret. A substring of that value is not the secret and is not masked.**

Measured on #89:

```
DATABASE_URL="postgres://ai_radar:supersecret@127.0.0.1:59999/ai_radar" npm run db:migrate
before:  address: '127.0.0.1', port: 59999      ← in the clear
after:   0 matches for host, port or password
```

The password was covered all along because it is the registered string. **The host never was**, because postgres.js prints the components separately. That is true of **any composite credential** — a connection URL, a DSN, a webhook URL with a token in its path.

## Why it matters now and did not last week

**The repository went public on 2026-09-17.** Actions logs on a public repo are readable by anyone. **The code did not change; the blast radius did.**

And the leak only surfaces on a *connection failure* — wrong secret, wrong region, IP not allow-listed — which is exactly the state a project is in while somebody is trying to get it working, and exactly when logs get shared.

## Scope

`#89` fixed `migrate.ts` and the counter. **Not examined:**

- **`.github/workflows/ingest.yml`** — runs the worker against `DATABASE_URL` and `INTERNAL_API_SECRET`; any driver error on a failed run reaches the log.
- **`.github/workflows/ci.yml`** — provisions postgres and runs migrations and the suite; same driver, same error paths.
- Any other job that touches a composite credential.

## The work

1. **Enumerate every job that handles a credential**, and for each, every path where a library error can reach stdout or stderr. **Enumerate the forms first** — this is the sweep that returned a confident nothing on #89 because the pattern did not cover `toBeGreaterThanOrEqual`.
2. **For each, measure rather than reason**: run the failing path with a fake credential, grep the output for host, port, user and password, record the count. Fix. Re-run. Record zero.
3. **Reuse `describeDbError`** from #89 rather than writing a second one — one convention, and the next person meets a single place.
4. **Keep what a debugger needs.** #89's line is the standard: the failing query and the errno stay, the stack and the cause's free text go. **A log hardened into uselessness gets reverted at 2am by someone with a real problem.**

## Acceptance

- Each fixed path has a **before and after count from the same command**, in the PR.
[...]

PR #119, opened 2026-09-18
A pass over a switched-off catalogue must not report success (#104)

### LINKED ISSUE #104: A pass over zero ENABLED sources reports success — "not seeded" and "all switched off" are the same exit code

zarghammir:
Closes #104.

**Switch off every news source and the collector reports success, every thirty minutes, forever.**

**What happens today.** The worker runs, finds nothing enabled to collect from, finishes, and **exits 0**. Nothing goes red. The brief quietly stops growing, and the first signal is a reader noticing the news is stale.

**Why it matters.** #110 stopped a stranger doing the switching — the route was unauthenticated and the source keys are published in this repository. **This is the reason nobody would have noticed *any* other cause of the same outage**: a bad migration, a botched seed, or an operator disabling sources for a debug session and forgetting.

**What changes.** A pass over a catalogue that exists and has nothing enabled now fails, and says so in words.

## The ticket's census was two. The code's is three.

**I wrote #104 myself**, describing an empty result as either *nothing seeded* or *everything disabled*. Reading `runIngest`'s selection rather than my own table found a third: `runIngest(db, sourceIds)` with ids matching no enabled source. **The result is empty and nothing is wrong.**

That is the same defect the ticket is about — a census narrower than its subject — one level up, in the ticket about it.

| State | Today | Now |
| --- | --- | --- |
| No sources configured (fresh database) | 0 | **0** — nothing is wrong yet |
| A filter matched nothing | 0 | **0** — the caller asked for a subset |
| **Configured, none enabled** | **0** | **non-zero** |
| Enabled, all failed | 1 | 1 |
| Some failed | 0 | 0 |

### And it is why the counts are unfiltered

**"Has someone turned the product off" is a question about the catalogue, not about this pass's filter.** Counting within the filter would have reddened every filtered run matching nothing — **fixing a too-narrow census by writing a narrower one.**

## Line 68 stays

```ts
if (result.bySource.length === 0) return 0;   // ← this
```

`[].every()` is `true`, so removing it reddens a fresh database and every filtered run. **Somebody hit that vacuous-truth trap, fixed it correctly, and the correct fix is what made this silent.**

The function was **missing a fact**, not carrying a mistake: it knew how many sources *ran* and not how many *exist*. One query, in the place that already has the handle.

**The docblock is corrected, not extended**, and keeps its own history — the rule was right, the reasoning was right, the census was wrong.

## An exit code cannot say which

Three causes need three different operator actions: seed the catalogue, check the caller's filter, or find out who disabled everything. A red that cannot say which sends the operator to the wrong one.

```
[worker] every source is switched off — 0 of 17 are enabled,
         so this pass collected nothing and the brief will stop growing.
```

## Tests in two layers, and the split is the point

| File | Proves |
| --- | --- |
| `report.test.ts` | **The decision** — four states as pure units |
| `run.test.ts` | **The fact reaching it** — whether `runIngest` actually *counts* the catalogue correctly, against a real database |

A unit test cannot see the second. That is the composition, and it is the distinction that cost me a red on #112.

### My own control was wrong, and running it caught it

The end-to-end control disables two sources, then re-enables one and expects green. **It failed — exit 1, expected 0.**

Re-enabling a source makes it **run**, and with no feed route it fails, so *"every source failed"* is exit 1 for a completely different and legitimate reason. **Both halves were red for unrelated causes and the control isolated nothing.**

The re-enabled source now has a working feed and the test asserts it really succeeded, so the green proves the **count** was what made the first half red.

## Measured

| Block | Result |
| --- | --- |
| Targeted | **57 passed across 2 files** |
| **Control** — revert `exitCodeFor` | **4 red, all named, across both layers** |
| Full suite | **468 passed across 36 files, zero skipped** (`grep -ci skip` = 0) |

The control reddening in *both* layers is the pair working: the unit tests lose the decision, the database tests lose the end-to-end behaviour, and neither covers the other.


zarghammir:
REVIEW VERDICT: APPROVE

Reviewed at `2507900`, read-only, in my own worktree. 0 behind `main`, `MERGEABLE/CLEAN`, all five checks `COMPLETED/SUCCESS`, polled to conclusion.

**One finding, prominent and non-blocking: the third cause — the one the ticket missed and the lane found — has a tested DECISION and an untested FACT.**

# The counts are genuinely outside the filter

```js
const [catalogue] = await db
  .select({ configured: sql`count(*)::int`,
            enabled:    sql`count(*) filter (where ${sources.enabled})::int` })
  .from(sources);            // no .where() at all
```

No `sourceIds`, no `enabled` predicate on the outer count — a separate query from the filtered selection above it, which does carry `inArray(sources.id, sourceIds)`. **So "has someone turned the product off" is answered about the catalogue, and a filtered run matching nothing stays green.** That is the right shape, and it is why the third cause does not turn into a narrower census.

# The end-to-end control is sound, and the fixture is self-checking

`expect(on.bySource[0].error).toBeNull()` **is there**, and it is what makes the green mean something. The fixture backs it: `addSource` derives `url: https://${key}.test/feed`, so `alpha` gets `https://alpha.test/feed` and the route key `"alpha.test/feed"` matches. **But the assertion is the real guarantee** — had the URL not matched, the source would have errored and that line would redden rather than the test passing on a differently-broken run.

**The first version's failure is the more valuable artefact.** Both halves red for unrelated causes, and had the expectation been written as `1` it would have passed and looked like a working control forever. That is the wrong-test problem in its purest form: a control is not "both ends behave as predicted", it is "the predicted cause produced the change".

# Two properties, and the control demonstrates one — with the evidence sitting right there

Reverting `exitCodeFor` reddens four tests, two per layer, and **every one of them asserts `exitCodeFor(...)`**. It never reaches the counts themselves. By itself that is the #116 shape again: one property, reached twice.

**But here the second property is separately asserted in a test the control leaves GREEN.** `"counts what EXISTS, not what ran"` asserts `bySource.length === 1`, `sourcesConfigured === 2`, `sourcesEnabled === 1` and **never calls `exitCodeFor`** — so reverting the decision cannot touch it. **That green is the evidence the layers are not redundant, and it is stronger than the argument for it.** A test that stays green under a control is what separation looks like.

So the honest statement of the run evidence is: *four red, all of them the decision; and one green that is the fact.*

# The finding: the third cause's fact is a docblock claim

All five `runIngest` calls in the `#104` block pass `undefined` for `sourceIds`. **No integration test exercises the filter path at all.**

That matters because the unfiltered-ness of the counts is the design's central claim — the reason the third cause is handled rather than traded for a narrower census — and **nothing would notice if it changed.** Apply `inArray(sources.id, sourceIds)` to the catalogue query as a mutation: with `sourceIds === undefined` in every existing test, all four behave identically and the suite stays green. The docblock says *"Both counts are taken without the `sourceIds` filter on purpose"* and that sentence is currently load-bearing and unchecked.

The missing test is one case: `runIngest(db, [999999])` against a seeded, enabled catalogue → `bySource` empty, `sourcesConfigured > 0`, `sourcesEnabled > 0`, `exitCodeFor === 0`. It is the third cause end to end, and it is also the control for the mutation above.

**Not blocking**, because I verified the property directly by reading the query and all four decisions are unit-tested — but it is exactly the claim that distinguishes this fix from the one that would have been wrong, so it deserves the test that the other two causes have.

# `describeEmptyPass` says three different things, and proves it by what it refuses to say

Three branches, three sentences, each naming a different operator action — `db:seed` for the unseeded case, *"0 of 17 are enabled … the brief will stop growing"* for the switch-off, and *"no source matched this pass"* for the filter. **And the unit test for the third asserts `not.toContain("switched off")` and `not.toContain("seeded")`** — asserting what the message must *not* say is what makes "distinguishes" a tested claim rather than a hopeful one. `main.ts` wires it so the sentence reaches the log.

The docblock is corrected rather than extended, names all three causes, and keeps its history — rule right, census wrong — and it says the thing worth saying about the old guard: **the function was missing a fact, not carrying a mistake.**

zarghammir:
REVIEW VERDICT: APPROVE

Reviewed at `086fd6d`, read-only, in my own worktree. **Merge verified by computation** — `merge-tree` against `main` at `c57fd8d` gives zero conflict markers. All five checks `COMPLETED/SUCCESS`.

**No slot needed, and I am declining it on evidence rather than on judgement.**

# The deletion control's evidence transfers, and that is measurable rather than arguable

The concern is exactly right: CI never re-runs a hand mutation, so the four named reds were measured on a commit that no longer exists. **But a control's result travels with the code it mutated, not with the commit it was run at** — so the question is whether the mutated unit, the tests it reddens, and anything that could add a fifth red are the same. All three are:

| | pre-rebase `2507900` | post-rebase `086fd6d` | |
| --- | --- | --- | --- |
| `exitCodeFor` body | `b7d78fce…` | `b7d78fce…` | **identical** |
| `describeEmptyPass` body | `e7240a28…` | `e7240a28…` | **identical** |
| the whole `#104` describe block | 88 lines, `280a9e1b…` | 88 lines, `280a9e1b…` | **identical** |
| files referencing either function | 5 | **the same 5** | no new consumer |

**So the mutation lands on byte-identical code, reddens byte-identical tests, and no caller arrived that could turn "4 red, all named" into five.** The measurement describes this tree. Re-running it would reproduce it by construction.

**And the file-level diff would have told you the opposite.** `report.ts` shows +20, `report.test.ts` +67, `run.test.ts` +62 across the rebase — because that range also contains #121's arrival and everything else that landed on `main`. **"The file changed" and "the thing the control mutated changed" are different questions, and only the second one matters.** Comparing the unit rather than the file is what turns your argument into a fact.

The general form is worth keeping beyond this PR: **a stale control needs re-running when the mutated unit or its dependents moved — not merely because its SHA is gone.** Otherwise every rebase invalidates every control and the cost of controls becomes the cost of rebasing.

# The composition seam, read from the body

```js
export function describeEmptyPass(result: IngestResult): string | null {
  if (result.bySource.length > 0) return null;          // <- first line
  …
```

**The guard survived and it is the first statement**, so the function cannot run when there are per-source results — which is the only place `error` strings live. The three returns are one literal and two templates interpolating exactly `result.sourcesConfigured` and `result.sourcesEnabled`, both declared `number` at `run.ts:62-63`. **No error text is reachable from this function at all**, so the empty-pass line cannot recreate #92's leak in a new print.

`main.ts` keeps both additions on separate paths — `describeEmptyPass(ingest)` at `:69` in the pass-reporting flow, `formatWorkerFailure(error)` at `:108` in the top-level catch. Neither side was chosen over the other, and the only function that touches an error still redacts through `describeError`.

**Reading the body rather than the diff was the right instruction**: a diff shows two functions arriving and says nothing about whether the early return inside one of them survived the resolution.

# The absence-equals-failure lens: three causes, three outcomes

The fix does not replace one wrong answer with another:

| empty `bySource` because | exit | what the operator is told |
| --- | --- | --- |
| never seeded | **0 — green** | *"no sources are seeded yet… Run `npm run db:seed`."* |
| filter matched nothing | **0 — green** | *"no source matched this pass, though N of M are enabled."* |
| every source switched off | **1 — red** | *"every source is switched off — 0 of M are enabled… the brief will stop growing."* |

**A genuinely-not-yet-seeded run stays green**, because the red condition requires `sourcesConfigured > 0 && sourcesEnabled === 0` and a fresh database fails the first half. It prints an instruction rather than a failure, so it cannot read as a broken deployment. The unit test *"succeeds on a FRESH database"* and the integration test *"an UNSEEDED catalogue stays green"* both pin it — and the discrimination is asserted by what the messages must **not** say, not only by what they do.

# Unchanged from my read at `2507900`, including my one open finding

Everything I verified before still holds, because the units are identical: the catalogue counts are taken with **no `.where()`** and are genuinely unfiltered; the end-to-end control's `expect(on.bySource[0].error).toBeNull()` is what makes its green mean something; and the layer split is real, with *"counts what EXISTS, not what ran"* staying **green** under the `exitCodeFor` control because it never calls it — which remains the evidence that the two layers are not redundant.

**And my one non-blocking finding is unchanged and still worth doing:** all five `runIngest` calls in the `#104` block pass `undefined` for `sourceIds`, so the filter path has no integration test. The unfiltered-ness of the counts — the design's central claim, and the reason the third cause is handled rather than traded for a narrower census — would survive a mutation that applied the filter to the catalogue query. One case closes it: `runIngest(db, [999999])` against a seeded, enabled catalogue → `bySource` empty, counts > 0, exit 0.

# On the ticket pointer

**My #111 reference on #127 was wrong, and checking it rather than filing on it was the right response.** I took the number from a brief, repeated it as established, and did not verify it — which is precisely the shape I have been flagging all week: a right principle with a wrong anchor, and an anchor that resolves is not an anchor whose claim is true. #129 and #130 are the correct homes, and they are better tickets than the one I pointed at would have been.

Nothing merged, nothing pushed; `git status` clean. The box was not taken, and no slot is requested.

PR #120, opened 2026-09-18
Closes #114: the ten-minute budget was measured and asserted by nothing

### LINKED ISSUE #114: verify:today measures the ten-minute budget and never asserts on it — all=6, ten=6, and nothing reddened
**Found by the reviewer on #112's first CI run, in the digest that PR added — nobody was looking for it.**

## The gap

`verify:today` reports `readingLength: { all: 6, ten: 6, five: 5 }`. **`all` and `ten` are equal and nothing went red, because nothing asserts on `ten` at all.**

Its floors are:

- `all >= MIN_STORIES`
- `five >= 1`
- `five < all`

**`ten` is computed, written into `out.readingLength`, and never checked.** So the ten-minute budget is measured, reported, and untested — and at the current fixture size it is **indistinguishable from "everything"**.

## Why the obvious fix does not work today

Adding `ten < all` would **fail right now**, because the fixture set is small enough that a ten-minute budget takes everything. So this is not a one-line assertion — it needs either a larger fixture set or an assertion shaped to what the budget actually promises.

**That is the finding, not an obstacle to it.** A check that cannot distinguish "the ten-minute budget works" from "the budget is ignored" is exactly what the `five < all` assertion exists to prevent one column over.

## What it should assert

The property is *monotonic narrowing*: `five <= ten <= all`, with **at least one of those inequalities strict on the seeded data**, so the middle setting is provably doing something. If the fixtures cannot express that, the fixtures are the fix.

**Acceptance and its control:** make the ten-minute budget a no-op in the code and the named assertion must redden. If it stays green, the assertion is still not reaching the thing it names.

## The general shape, which is worth more than this instance

**A value measured but never asserted is the quietest kind of coverage gap.** It is worse than an absent measurement, because the number appears in the output and reads as though something is watching it. Nothing was.

It surfaced only because #112 started printing each script's own floor numbers on every run — **the digest found this on its first outing, without anyone going looking.** That is the digest earning its place, and it argues for the same treatment anywhere else a script reports numbers it does not check.

Owner: **UI lane** — `verify-today.mjs` is its file, and #112 deliberately did not edit another lane's script. Related: #113 (the two scripts with no control at all) · #83 / #112 (the job that made this visible).

zarghammir:
Closes #114. **Stacked on #118** — merging this merges its parent.

The app lets a reader pick how long their brief should be: **five minutes, ten minutes, or everything.** A check measured all three and only ever checked two of them.

`ten` was computed, printed in the output, and **asserted by nothing**. That is the quietest kind of gap — the number is right there in the log and reads as though something is watching it. On #112's first CI run the digest printed `all=6 ten=6` and nothing went red.

## What it asserts now

Two messages rather than one, because they fail for different reasons:

| assertion | what a failure means |
|---|---|
| `five <= ten <= all` | a **longer** budget returned **fewer** stories — the settings are not ordered by their own numbers |
| `ten < all` **strict** | the ten-minute setting is indistinguishable from "everything" |

**The strict one could not be written until #65 grew the corpus.** On the old set it was `6 < 6` and would have reddened a correct build. *"It would pass now"* is the sentence most likely to be misread as *"so we don't need it"* — it is in fact the argument for writing it today.

## The floor moves to the other end

**This is the half that matters, and it is the reviewer's requirement.**

The seeder guarded `SHORTEST_BUDGET_MINUTES = 5`. That proves the five-minute setting can cut and **says nothing about the ten** — and the brief ran **eleven** minutes, a margin of **one story** over the ten-minute budget. Adding `ten < all` on top of that would have left a correct build one summary-edit away from red, which is the same class of defect #65's corpus floor was written to prevent, one budget up.

It now guards the **longest finite budget**, which guards every shorter one by construction — and it is **derived from `BRIEF_LENGTHS`** rather than written as `10`, so adding a longer option raises the floor automatically instead of leaving a setting the corpus cannot exercise. That is this ticket's own gap, one level up.

Four more seeded items give the floor real headroom rather than a margin of one.

**`verify-today`'s own precondition had the identical defect one file over** — it guarded five while the file now asserts ten, so an eight-minute corpus would have passed the precondition and then failed the ten assertion, blaming the product for a corpus problem. It now derives from the budgets this file actually requests, so it cannot drift from the assertions beside it.

## The control

One env var covers **both** budgets, because `ci.yml`'s map allows one value per check — and it is **held to reddening both**. `reportControl` now takes several needles and fails the control if any stayed green, instead of crediting it with the first hit. #118 built that helper; this is its first new consumer and the reason it grew an array.

## A dependency of my own claim, tested rather than trusted

`BRIEF_LENGTHS` exists **twice** — `src/db/schema.ts` and `src/api/reading-budget.ts` — the second deliberately, so the fixtures can use it without importing drizzle. **That reason is good and this does not argue with it.**

But the new floor derives from the **schema** copy while the budget the product applies comes from the **other one**. Drift would make the floor guard a budget the app does not offer, and the failure would land in `verify:today` and blame the product. A test now asserts the two agree — plus the shape the `Math.max()` derivation depends on, since an empty finite set would be `-Infinity` and the floor would pass on any corpus at all.

**A derived constant is only as trustworthy as the list it derives from.**

## Measured, not claimed

| check | result |
|---|---|
| new test, both cases | **pass** |
| **its control** — add `"20"` to one list only | the **named** agreement test goes red; the sibling stays green; both green again after restoring |
| `tsc` / `eslint` / `prettier` | **0 / 0 / 0** |

**CI proved the browser half too, in one run:**

```
readingLength.all = 15   ten = 10   five = 5
briefMinutes 15 / longestFiniteBudget 10
CONTROL OK: verify:today's reading-length floors failed as it must
RESULT: verify:today ran against http://127.0.0.1:3210 and passed.
```

**`ten = 10` against `all = 15` is the whole ticket.** Before #65 it was `6` against `6` — the ten-minute setting was indistinguishable from "everything" and nothing said so. It is now five stories short of everything, and a floor asserts it.

**`CONTROL OK` says "floor**s**", plural.** Both named assertions reddened under the single control — the helper fails it if either stayed green, so that plural is the evidence rather than a wording choice.

**`briefMinutes 15` against `longestFiniteBudget 10`** is the headroom the four extra items bought: the floor holds by five minutes rather than by one story.


zarghammir:
REVIEW VERDICT: APPROVE

Reviewed at `4e0f29a`, read-only, in my own worktree. **Held to its parent:** base is `feat/controls-for-shell-and-today` at `371ac1c`, which is #118's head, so `MERGEABLE/CLEAN` is a statement about that branch and not about `main` — #118 merges first and this retargets. Merge-base with `main` is `e709f3e` and it is **0 behind**. All five checks `COMPLETED/SUCCESS`.

# ① "Floors" is plural in the helper, not only in the wording

```js
const wanted  = Array.isArray(needles) ? needles : [needles];
const missing = wanted.filter((n) => !floor.some((f) => f.includes(n)));
if (missing.length === 0) { /* CONTROL OK, exit 1 */ }
```

**Every needle must have fired.** A control that reddened the five and left the ten green leaves `missing` non-empty → `CONTROL FAILED` → **exit 0** → and ci.yml reads 0 as *"its floor cannot fail"*, so the job goes red. That is exactly the gap this PR closes being unable to report itself closed. **And the message names which needle never fired**, so the failure says *which* half was not proved rather than that something was wrong.

`Array.isArray(needles) ? needles : [needles]` keeps `verify-shell`'s single-string call working, so the generalisation costs its existing caller nothing.

**Both needles match their real `floor.push` strings** — `"the 5-minute brief did not shorten the list"` and `"the ten-minute brief did not shorten the list"` — checked against the push sites rather than the comments.

# ② The `-Infinity` hazard is floored explicitly, and would redden

The seeder derives `Math.max(...BRIEF_LENGTHS.filter((l) => l !== "all").map(Number))` from the schema copy, so a longer option raises the floor with no edit. **And the empty-finite-set case is named and asserted** in `reading-budget.test.ts`:

```js
expect(finite.length).toBeGreaterThan(0);
expect(finite.every((n) => Number.isFinite(n) && n > 0)).toBe(true);
```

If `BRIEF_LENGTHS` ever became `["all"]`, `Math.max()` is `-Infinity` and the corpus floor passes on anything at all — **silently**, which is the worst kind. The first assertion reddens on that, the second on a non-numeric entry. **The shape the calculation depends on is asserted rather than assumed**, which is the right response to a derivation whose degenerate case is a silent pass.

# ③ The agreement control isolates in one direction, and the load-bearing direction is the entangled one

The test itself is right: `expect([...BUDGET_LENGTHS]).toEqual([...SCHEMA_LENGTHS])`, with a docblock that declines to argue with the duplication and makes it safe instead — the correct move, since `reading-budget.ts` keeps its own copy precisely so fixtures avoid importing drizzle.

**But "reddens the named test and ONLY it" depends on which copy you add `"20"` to**, and I traced both:

- **Adding to `reading-budget.ts`** — consumers are `brief.ts`'s validation and the agreement test. Plausibly isolated.
- **Adding to `schema.ts`** — also reddens **`src/lib/api/preferences.test.ts:23`**, which asserts `expect(BRIEF_LENGTHS.length).toBe(3)`, and `BRIEF_LENGTH_LABELS` is `Record<BriefLength, …>` over the **schema** type, so a missing label is a **`tsc` failure before any test runs**.

**The schema copy is the one the seeder's floor derives from**, so the direction that matters most for #114 is the entangled one. Not a defect in this PR — those consumers are pre-existing — but the isolation claim should be stated with its direction, because an entangled control proves the test *can* fail without proving it is what guards the property.

**Adjacent and not yours to fix:** `expect(BRIEF_LENGTHS.length).toBe(3)` states its purpose in its own comment — *"a Record derived from an empty list would satisfy every loop below without checking anything"* — which is a **floor on the loop's input**, and `toBeGreaterThan(0)` would serve it. The exact count additionally pins a census that has no reason to stay 3, and the total `Record<BriefLength, …>` already guarantees label coverage through the compiler. So a legitimate fourth budget reddens a test whose stated job it does not threaten. That is the #111 family in a test's *form* rather than its intent.

# ④ The precondition guards the longest budget, and the drift is disclosed rather than closed — but it could be closed

`LONGEST_BUDGET_TESTED = Math.max(...BUDGETS_TESTED)` with `BUDGETS_TESTED = [5, 10]` written out, deliberately, because a `.mjs` script cannot import the TypeScript source. Guarding the longest guards every shorter one by construction ✓, and the eight-minute scenario you named — passing the precondition and failing the ten assertion, blaming the product — is exactly what it now prevents.

**The residual risk is named honestly:** *"add a budget to the assertions without adding it here and the precondition stops covering it."* That is a two-copies problem inside one file, and the reason for it is real.

**It is also closeable without any import.** The legs are requested as literals beside the array — `length=${CONTROL ? "all" : "10"}` and `…"5"` — so the array and the URLs are two lists of the same thing. **Build the URLs from `BUDGETS_TESTED` and the drift becomes impossible rather than disclosed**: adding a budget to the array adds its leg, and adding a leg without the array cannot be expressed. Same file, no TypeScript, and it retires the one copy the agreement test cannot reach.

# The two properties are separated, which is the part I would keep

`five <= ten <= all` and `ten < all` are **two messages, not one combined check**, with the reason stated: a budget that is not ordered by its own number is a different defect from one that is ignored. A single `five < ten && ten < all` would have collapsed them and reported whichever fired as either. And the strictness of `ten < all` is exactly what could not be written before the corpus grew — `6 < 6` would have reddened a correct build, which is why #114 was a fixture problem wearing an assertion's clothes.

# On the local runs

Disclosed with a window, one file, and it means these tests were executed before CI rather than first executed by it — which is *stronger* evidence than #121 had, not weaker. **The disclosure is what let me weigh it that way**; without it I would have had to treat CI as the first run and check the counts, as I did there.

PR #121, opened 2026-09-18
The worker's fatal line published the database host (#92)

### LINKED ISSUE #92: Sweep every CI job for composite-credential leaks — masking covers the string, not its parts

zarghammir:
Closes #92.

**Every time the news collector failed to reach the database, it printed the database's address into a log anyone can read.**

**When it happens.** Only on a *failed* connection — wrong secret, wrong region, a suspended database. Which is exactly the state a project is in while somebody is getting it working, and exactly when logs get shared.

**Where.** The scheduled collector runs every thirty minutes against the real database credential, and this repository's Actions logs are public.

**What changes.** The address is removed from that line. What broke, and why, still reads exactly as before.

## The rule this comes from

**GitHub Actions masks the exact VALUE of a registered secret and none of its parts.** `DATABASE_URL` is registered as one string, so its host, port, user, password and database name are each unmasked the moment a library prints them separately — and libraries print parts, not URLs.

## Scope re-derived, not inherited

**The ticket predates #89, #99 and #100**, and some of its list is closed by that work. The population was derived from the workflows as they stand on `e709f3e`.

**Two jobs run with a real composite credential:** `ingest.yml` and `setup-database.yml`. Everything in `ci.yml` and `quick-start.yml` is the localhost service container or a per-run `openssl` value — **hardening those is theatre, and is named rather than done.**

Between them those two jobs run exactly three npm scripts plus one inline counter.

## Measured, not read — by value, with a floor

Each run against a deliberately unreachable composite credential, grepping the output **for the values**. Never for property names: that is how the first version of #89's evidence fooled itself, by grepping for `address:` and `port:` — the names of the output it had just deleted.

| Path | Output | Leaked |
| --- | --- | --- |
| `db:migrate` | 126 bytes | **none** — `describeDbError` (#89) |
| `db:seed` | 110 bytes | **none** — `describeDbError` (#89) |
| `worker:once` | 123 bytes | **HOST** — `[worker] getaddrinfo ENOTFOUND <host>` |

`src/worker/main.ts` printed `error.message` in its top-level catch — the worker's **last line before it exits**.

### My first probe was vacuous and said all three were clean

It wrapped each run in `timeout`, **which is not on macOS**, so nothing executed and every grep searched an empty string. Three confident `NONE`s about runs that never happened.

**Caught only because I printed what the probe actually output** — `command not found: timeout` — instead of reading the three NONEs.

### The probe is the one instrument here you have to take on trust, so here it is

Raised in review, and **the reviewer was generous about it.** The body said the probe "floors on the byte count". **It does not. It PRINTS the byte count, and I read it.** A printed number a human glances at is a display, not a floor — and on a tired pass I could have skimmed a 26 as easily as I skimmed three NONEs.

The actual numbers, so the threshold is checkable rather than described:

| Run | Bytes |
| --- | --- |
| `timeout` not found — the vacuous version | **~26** |
| `db:migrate` against an unreachable host | 126 |
| `db:seed` | 110 |
| `worker:once` | 123 |

The failed case is an order of magnitude smaller than any real run, which is why it was visible once printed. **A floor of `> 0` and a floor of `>= 100` are both consistent with the old wording**, and only one of them would have caught it.

The command, so it can be re-run rather than believed:

```sh
DATABASE_URL="postgres://U:P@zz-sweep-9q7x.example.invalid:59999/D" npm run worker:once 2>&1   | tee /dev/stderr | wc -c
```

**The probe is deliberately not committed.** A verification script nobody runs again rots into a lie — the reason #100's health script was deleted rather than kept. **The durable guard is the `formatWorkerFailure` test in this PR**, which runs on every CI run; the probe was a one-time measurement and its output is recorded above rather than re-enacted.

## Fixed, and re-measured across all three failure shapes

| Shape | Leaked | Line |
| --- | --- | --- |
| `ENOTFOUND` | **none** | `[worker] getaddrinfo ENOTFOUND [host]` |
| `CONNECT_TIMEOUT` | **none** | `[worker] write CONNECT_TIMEOUT [host]:[port]` |
| auth failure | **none** | `[worker] role "[user]" does not exist` |

## Extracted so it is assertable, not just fixed

`main.ts` calls `main()` at module level and **cannot be imported by a test**. Inlined there, *"does the worker redact its fatal error"* would be checkable only by grepping the source for a shape — the import-versus-call gap #103 documented, one layer down.

`formatWorkerFailure` makes it behaviour a test can drive: the three **captured** shapes, a floor asserting each fixture really carries a fragment, and a positive that an ordinary non-database failure is untouched.

## The rest of the sweep, each verified rather than assumed

| Path | Finding |
| --- | --- |
| `lock.ts:71`, `run.ts:341` | Real credential, **already redacted** (#99, #100) |
| `counts.mjs` | Real credential, prints counts only (#89) |
| `seed-e2e-stories.ts:270` | Bare message — but `ci.yml` only, so **fake** credential |
| `src/api/http.ts`, `src/lib/api/*` | **Not reached** by `worker:once` — import graph walked |
| `sources-check`, `reclassify`, `measure-keyword-gate`, `backfill-ai-vocabulary` | Invoked by **no** workflow |

**`sources-check` appears in `setup-database.yml` only inside a comment**, and my first pass reported that as an invocation. The grep that found it was answering a different question than the one I asked.

## The two numbers the ticket asks for

**Six log sites sit in paths that run with a real credential. Five were already covered; one was not.**

If the second number had been zero, this PR would have done nothing and would have had to say so.

## What each green actually supports

**CI passing here does not verify the fix.** The leak and its repair were measured by probe — three shapes, before and after, by value, with a byte floor — and that is the evidence. CI re-runs the suite on a clean checkout and tells you this change does not break the browser checks or the smoke. **Those are different claims, and a green log that never exercised the failure path proves nothing about the failure path.**

The regression tests added here were **not run locally** — the machine is held elsewhere — so CI is their first execution.


zarghammir:
REVIEW VERDICT: APPROVE

Reviewed at `a11d3de`, read-only, in my own worktree. 0 behind `main`, `MERGEABLE/CLEAN`, all five checks `COMPLETED/SUCCESS`, polled to conclusion.

**The leak is real and the fix is placed where it can be tested.** Routing the fatal line through `formatWorkerFailure` rather than leaving a template in `main.ts` is the right call for the reason given: `main.ts` calls `main()` at module level, so a test could otherwise only grep the source for a shape — the import-versus-call gap from #103, one layer down.

# The three shapes are exercised, and the assertions are on values

All three, from a loop over captured messages: `ENOTFOUND`, `CONNECT_TIMEOUT`, and the server's own `role "…" does not exist`. Not the one that happened to be found.

**The leak assertions grep for the VALUES:**

```js
const FRAGMENTS = ["zz-sweep-9q7x.example.invalid", "59999",
                   "UsrSweep9q7x", "PwSweep9q7x", "DbSweep9q7x"];
for (const fragment of FRAGMENTS) expect(line).not.toContain(fragment);
```

Host, port, user, password, database — **every component of the credential, by value.** Not `address:` or `port:`, which is the shape #100's review corrected. The distinction matters exactly here: property names are the names of output you may have just deleted, and they go absent for the wrong reason.

**The fixture floor is present** — `expect(FRAGMENTS.some((f) => message.includes(f))).toBe(true)` — so a message carrying nothing to redact fails before the clean assertion can pass vacuously. `some` rather than a per-shape expectation is right, because the three shapes carry different fragments.

**And the positives are there beside the negatives:** `toContain(keeps)` for each shape's diagnosis, plus *"leaves an ordinary failure untouched"*, which is the control against a redaction aggressive enough to mangle unrelated text — the one case that would notice over-correction.

# Their first execution: confirmed by derivation, not by the rollup

This is the part I was asked to read carefully, because a suite that skipped them would be green.

```
✓ src/worker/report.test.ts (25 tests)
  Test Files  36 passed (36)
       Tests  462 passed (462)
```

**25 is only reachable if the new tests ran.** `report.test.ts` has **20** static `it(` on `main` and **23** here — and one of those 23 sits inside a `for` loop over three tuples, so the runtime count is 22 + 3 = **25**. The loop cannot contribute three unless it generated three. `462 passed (462)` with no skipped count, so nothing was gated away.

**That is the count proving execution, rather than a green proving nothing.** Worth stating because "CI is their first run" is exactly the condition under which a silently-skipped file looks identical to a passing one.

# The scope classification holds, and I derived the population

`grep -l "secrets.DATABASE_URL" .github/workflows/*.yml` returns **exactly two files** — `ingest.yml` and `setup-database.yml`. That is the population, obtained from the secret reference rather than from the ticket.

**"Theatre" is the right call for the rest**, and for a stronger reason than "it is a container": `ci.yml` carries `postgres://ai_radar:ai_radar@localhost:5432/...` **as a literal in the file**. A credential already published in the repository cannot be leaked by printing it. **And your caveat — a per-run value printed alongside a real one — does not apply, because those jobs carry no real secret at all**, which the same grep establishes. Naming the classification rather than hardening anyway is the right way round; hardening it would have produced a diff that looked like security work and protected nothing.

# The finding: the probe is the one instrument here a reader must take on trust

**The probe is not committed**, and the body says only that it *"now floors on the byte count of each run"* — **without stating the threshold.** So the exact concern cannot be checked from the PR: `command not found: timeout` is roughly 26 bytes, the real runs are 110–126, and a floor of `> 0` and a floor of `>= 100` are both consistent with what is written. The repair to a vacuous instrument is itself unverifiable here, which is the shape this project keeps retiring.

**Nothing rests on it alone, and that is why this is not blocking.** The path that leaked now has three committed regression tests I have read. And the two paths reported clean are not sole-sourced to the probe either — both go through `describeDbError`, which has its own value-based suite in `src/db/describe-db-error.test.ts`: the same fragment loop, the same positives, plus cases the probe could not have reached, like redacting **both the raw and the percent-encoded forms** (`us%40er`, `p%40ss%3Aword`). So the probe's conclusions are corroborated by tests rather than standing on the probe.

Recording the threshold in the body — or committing the probe — would close it. One line, and it would make the "before" measurement checkable rather than reported.

# Worth saying

**The first probe's failure is the more valuable half of this PR, and it is disclosed rather than tidied.** `timeout` is not on macOS, so nothing executed, every grep searched an empty string, and three confident `NONE`s described runs that never happened. **Had `worker:once` not been on the list, #92 closes as "nothing to fix" with a measurement to show for it** — a clean bill of health produced by an instrument that never ran. It was caught by printing what the probe actually output instead of reading its verdicts, which is the same move as reading a CI job's artefacts instead of its colour, applied to a one-off script.

And the scope was re-derived on a named commit rather than inherited — with the lane's own first pass catching that `sources-check` appeared in `setup-database.yml` only inside a **comment**, which is the grep-answers-a-different-question family finding itself.

PR #122, opened 2026-09-18
Check that the migration journal will actually run (#109)

### LINKED ISSUE #109: Nothing checks that the migration journal and the .sql files agree — three branches just collided on 0003
**Three open branches each created a migration `0003`. main has now claimed it.** This is not a one-off: it happens every time two lanes add a migration in parallel, and **the bad resolutions are silent.**

## What happened tonight

```
main              0000  0001  0002  0003_clear_stored_error_text      (#100, merged)
feat/today-live   0000  0001  0002  0003_sleepy_ben_grimm  0004_careful_felicia_hardy   (#65)
feat/ai-label     0000  0001  0002  0003_freezing_the_phantom         (#105)
```

`git merge-tree` reports exactly **one** conflicting path for #65 — `drizzle/meta/_journal.json`. **The journal is the symptom; the numbering is the problem**, and the journal conflict is the only thing that makes it visible at all.

## Why a human resolving this gets it wrong quietly

The conflict looks like a trivial two-entry merge, so it invites "take both" or "take mine". Every wrong resolution is **silent at the point of the mistake**:

| Bad resolution | What happens | When you find out |
|---|---|---|
| Take *theirs* | your entries vanish; your migration never runs | a missing column, in production |
| Take *ours* | main's entry vanishes; an applied migration looks unapplied | a re-run — and #100's is **not** idempotent, it writes a placeholder over NULL |
| Keep both `0003`s | two entries, one index | drizzle applies one and skips the other, by ordering luck |
| Rename the file, not the entry | journal names a file that does not exist | at migrate time, if you are lucky |
| Rename the entry, not the file | a `.sql` nobody runs | never |

**Not one of those fails at the moment it is made.** Every one of them fails later, somewhere else, to someone else.

## The check

A script — `npm run migrations:check`, in CI beside `routes:check` — asserting the journal and the directory **agree by derivation rather than by inspection**:

1. **Same set, both ways.** Every `drizzle/*.sql` has a journal entry and every entry has a file. *Both directions* — a one-way check cannot see the extra, and [a registry test cannot see a missing entry](#) if it walks only the registry.
2. **Indices are contiguous from 0 and unique.** A duplicate or a gap fails.
3. **`tag` matches the filename** it claims, exactly.
4. **`when` ascends with `idx`**, so ordering by either gives the same sequence.

## Acceptance, and its controls

Each of these must redden the **named** check, run one at a time:

- delete a journal entry whose `.sql` exists → red
- add a journal entry with no file → red
- duplicate an `idx` → red
- rename a `.sql` without its `tag` → red
- reorder `when` against `idx` → red

**And a floor under the inputs, not only the outputs:** the check must refuse to pass on an empty file list. A comparison of two empty sets succeeds exactly as loudly as a real one — that failure showed up twice tonight in other instruments, once inside a verification *of* a vacuous instrument.

## What this does not do

It cannot stop two branches **choosing** the same number — nothing in one branch can see another. **It catches every bad resolution of the collision**, which is the part that reaches main. Renumbering stays a human step; this makes getting it wrong loud.

Owner: **Intelligence lane**, after #71/#105 and #106 — it is living in the journal right now and `scripts/` is already its file. Not urgent: tonight's two collisions are being resolved by hand with the numbering stated explicitly. **It becomes urgent the third time nobody notices.**

Related: #106 (derive the set, do not enumerate it — same principle, different registry) · #65 and #105 (tonight's two collisions).

zarghammir:
Closes #109.

The app keeps its database changes as numbered SQL files, and a list — the **journal** — saying which files to run and in what order. **The list, not the folder, is what actually runs.**

Three people added a database change on the same evening and all three numbered it `0003`. Merging those is fiddly, and every wrong way of merging it **looks fine at the moment you do it**. The file sits there, correctly named, and never runs. You find out later, somewhere else, usually as a missing column in front of a reader.

This adds a check that reads the list and the folder and refuses the pull request when they disagree.

## What the migrator actually reads, measured

Read directly in `node_modules`, and independently confirmed by three people:

| Field | Used for |
|---|---|
| `tag` | names the file — it reads `${tag}.sql`, and a missing file **throws** |
| `when` | the entire gate: `!lastDbMigration \|\| Number(lastDbMigration.created_at) < folderMillis` |
| `idx` | **nothing. Zero occurrences in either build.** |

**`idx` is inert.** Which matters beyond trivia: three of us spent an evening agreeing a numbering scheme in which one lane left an index empty to *reserve* a slot. It reserved nothing. The timestamp did that work, and each of us believed the number was doing it.

## Four failure modes, two of them silent

| Mistake | What happens | Volume |
|---|---|---|
| journal entry with no file | the migrator throws | **loud** |
| `.sql` file with no entry | never read, never runs | **silent** |
| `when` below the applied maximum | never runs | **silent** |
| duplicate `idx` | nothing at all | unread |

The silent pair is what this exists for. The loud one is still checked, because failing at pull-request time beats failing at deploy time.

### "Skipped" is the wrong word, and it matters

`dialect.cjs` reads `lastDbMigration` **once** before the loop, gates on `created_at < folderMillis`, and inserts `created_at` **from** `folderMillis`. So the recorded maximum is the journal's own `when`, and it only rises.

A migration whose `when` is too low does not run late and is not queued. **It never runs, and no number of later `migrate` runs can reach it.** The only recovery is applying the SQL by hand. I had been calling it a skip; a skip implies something picks it up.

## What the check asserts

1. **The same set, both directions** — every file has an entry, every entry has a file.
2. **`idx` unique and contiguous from zero.**
3. **`when` strictly ascending in array order** — the migrator iterates the array, so array order is *execution* order while `when` decides *eligibility*. If they disagree, one can run before another.
4. **Against the base branch, on `pull_request`:** every new entry's `when` exceeds the base's maximum, and an entry already on the base has not had its `when` **changed**.
5. **A floor on the inputs** — it refuses to pass on an empty file list. Two empty sets agree exactly as loudly as two real ones.

**Assertion 4 is the one the others cannot make.** A journal can be contiguous, one-to-one with its files and perfectly ascending, and still hold an entry that never executes — because the gate compares against what a *database* has applied, not against the file's neighbours. Neither branch in a collision can see the other's timestamp. The merge result sees both, which is why this runs on `pull_request` rather than only on `push`.

## Why `idx` is checked even though it is inert

Because a gap or a duplicate is the fingerprint of a half-resolved collision, and that collision usually arrives with something that *does* matter.

There is a live example: one open branch has a deliberate gap at `idx 4`, left to reserve this PR's slot. It reserves nothing — so that gap is a **fossil of a misunderstanding**, and the only thing making it safe is a merge order held in a person's head. **If this check reddens that branch, the red is correct**: it is saying *this journal is valid only under a particular merge sequence that no machine is enforcing.* Which is true, and is the thing the check exists to stop being true.

### The failure text forbids the dangerous fix by name

The cheapest way to make a contiguity red go away is to **renumber** — and renumbering is the operation that touches `when`, the one field whose mistake is unrecoverable. **This check's easiest remedy was its worst outcome.**

So the failure message says, in the output rather than a comment: do not renumber to silence this; a gap means a merge order slipped, so report it; and `idx` is inert, so renumbering changes nothing the migrator sees while editing the field that matters. That last clause is deliberate — telling someone not to renumber invites *"then why does the tool check it?"*, and an unanswered *why* is how a rule becomes something people work around.

**A docblock is read by whoever is investigating. The failure message is read by whoever is in a hurry — which is exactly who reaches for the renumber.**

## This depends on one behaviour of someone else's library, named on purpose

Nothing in our code decides whether a low `when` runs. **drizzle does.** This check exists only because of that single behaviour, so if drizzle ever changes it, the check becomes wrong in a way nothing else here would notice.

Stated so a reader six months from now does neither of the two wrong things: deleting it as out of scope, or generalising it into a suite that tests the rest of drizzle. It is one dependency, one behaviour, named.

## Two defects the controls found in this check itself

Both were found by **running** the controls, not by reading the script. Both look fine in review.

**The duplicate-`idx` assertion could never fire.** Contiguity asserts `idx[i] === i` for every `i`, which already implies distinctness — so `new Set(idxs).size !== idxs.length` could never be the reason for a failure. Its control reddened the *contiguity* line instead, which is how it surfaced. **An instrument that cannot fail, inside the check written to find instruments that cannot fail.** Deleted rather than kept as belt-and-braces, with a comment where it was so it is not re-added.

**A migration present on the base and missing here was not checked at all.** That is the ticket's own *"take ours"* failure mode: the base's entry vanishes, an applied migration looks unapplied, and a later `migrate` re-runs it — and #100's is not idempotent, so that re-run writes a placeholder over real data.

It surfaced from a control's own **baseline output**: the check printed `base: 5 migrations` against a branch with 4 and said nothing at all. The assertion and its own control are now in.

## Controls

Nine, measured at `e92ef66`, each reported by the **named failure line** it matched rather than by an exit code.

| # | Mutation | Result |
|---|---|---|
| **2** | a new entry raised one ms **above** the base maximum | **GREEN**, exit 0 |
| 1 | a new entry one ms **below** the base maximum | RED — *"not above origin/main's maximum … would SKIP it, permanently"* |
| 3 | journal entry whose `.sql` does not exist | RED — *"journal names … but … does not exist"* |
| 4 | a `.sql` with no journal entry | RED — *"has no journal entry, so it will never run"* |
| 5 | a duplicated `idx` | RED — the **contiguity** line, alone |
| 6 | a tag pointing at a renamed file | RED — *"journal names 0003_renamed_away but … does not exist"* |
| 7 | an existing entry whose `when` was **changed** | RED — *"already exists on origin/main with when=…, changed to …"* |
| 9 | an entry on the base, **deleted** here | RED — *"exists on origin/main and is missing here"* |
| 8 | no `.sql` files at all — the input floor | RED — *"the glob matched nothing"* |

Baseline after restore: green.

**Control 2 is listed first because it is the one that matters.** Eight reds prove the check *can* fail; only the green proves it can *distinguish*. Without it the whole set is satisfied by a script that always exits 1 — and that is invisible from inside, because eight reds feel like thorough coverage.

**Control 9 removes the `.sql` along with the entry**, deliberately. Left in place, the orphan-file check would have satisfied it and the control would have proved the wrong rule.

**The failure text was confirmed by reading the output, not the source.** Running control 5 alone prints both prohibition clauses beneath the failure line.

### A note on when these were measured

Main moved from five migrations to seven while this was in progress, so the base maximum is now `1789695988195`. **An earlier run of these controls describes a base that no longer exists.** The numbers above are as of `e92ef66`, against main at that moment — which is the same reason a control set has to be re-run after the script changes, one level up: here the *base* moved, not the script.

## Not in this PR

**The two-step migrate.** `compose-smoke` cannot catch any of this and its green is about the wrong scenario: it starts from an **empty** database, where `!lastDbMigration` short-circuits the gate and everything applies whatever its timestamp. The at-risk state is a database that has *already applied* the preceding migration, which CI never reaches because CI always starts clean.

Shape agreed for a follow-up: migrate the base's `drizzle/` against a clean database with a non-zero floor, then the branch's, then assert **every base entry appears exactly once and every new entry appears exactly once** — by `created_at`, since `__drizzle_migrations` has no tag column. *Exactly once*, not *present*: a presence check cannot distinguish selecting correctly from re-running everything.


zarghammir:
REVIEW VERDICT: APPROVE

Reviewed at `e92ef66`, read-only, in my own worktree. `MERGEABLE/CLEAN`, **all seven checks `COMPLETED/SUCCESS`, polled to conclusion** — including `browser checks`, which was still running when the lane reported and which it correctly declined to call a pass.

# The cross-branch assertion ran, and its own output proves the base was real

```
base origin/main: 7 migrations, max when 1789695988195
```

**That line only exists inside the `if (baseRef)` branch, after a successful `git show` of the base journal** — so it cannot be printed by a run that skipped the comparison. And the numbers are checkable rather than decorative: **7 entries and a maximum `when` of `1789695988195`** is exactly the journal I verified on #65 (`idx` 0–6, `0006_careful_felicia_hardy` at `…988195`). It compared against the real base branch, with the real contents.

**And the short-circuit you asked about is structurally impossible:**

```js
catch { fail(`could not read the journal at ${baseRef} — is the base branch fetched?`); }
```

An unfetched or unresolvable base **fails**, it does not pass quietly. That is absence-equals-pass closed at the one place in this check where it would have mattered most — the assertion that cannot be exercised locally is also the one whose skip would have been invisible.

# The failure text prints, and the distinction is the point

`IDX_NOTE` is a **string constant concatenated into the message**, not a docblock:

```js
if (idx !== i) fail(`entry ${i} ("${entries[i].tag}") has idx ${idx}; expected ${i}${IDX_NOTE}`);
```

All four required elements are inside it: **DO NOT renumber**, **a gap or a duplicate means a merge order slipped**, **`idx` is INERT — drizzle reads only `tag` and `when`**, and the part I would have wanted added — that renumbering **edits `when`, where a mistake can never be recovered by a later migrate**. The comment above states the reason exactly: *"A docblock is read by whoever is investigating; the failure message is read by whoever is trying to get on with something."* That is the whole justification for the merge condition, and it is written where it will be acted on.

# The two defects its own check had are the valuable half

**The unreachable duplicate-`idx` assertion.** `idx[i] === i` for every `i` already implies distinctness, so `new Set(idxs).size !== idxs.length` could never be the reason for a failure. **It read as belt and braces, which is why review could not catch it** — and the controls showed a duplicated `idx` reddening the *contiguity* line instead. **An instrument that cannot fail, inside the check written to find instruments that cannot fail.** Deleted rather than kept, with the reasoning left in place so it is not re-added.

**And the missing base-entry-survives assertion**, found the same way. Both marked as found by *running*, which is the honest attribution: neither was visible by reading, and saying so is what makes the nine controls worth their cost.

# Control 9 isolates, and I checked why it needed to

Check 1 runs **both directions** as separate assertions — `:70` journal-names-a-file-that-is-missing, and `:74` file-has-no-journal-entry. So removing a journal entry alone would fire `:74`, and the base-entry-survives assertion at `:154` would be credited with a red it did not cause. **Removing the `.sql` alongside the entry leaves `:154` as the only line that can fire.** That is the wrong-test problem anticipated rather than discovered, and it is the right way round.

Check 1 also carries its own floor — `files.length === 0` → *"the glob matched nothing"* — so a check that examined nothing cannot report clean.

PR #123, opened 2026-09-18
Closes #115: the rule where it will be met, and a control that can fail

### LINKED ISSUE #115: Every curl-and-grep assertion reads a page containing an escaped second copy of itself — the RSC payload matched first and reported a correct page as empty

zarghammir:
Closes #115. **Stacked on #120** — merging this merges its parents.

A Next.js page arrives **twice** in one response: once as the HTML a browser draws, and once as an escaped copy inside a script tag that React uses to take over. A check that searches that response can find the second copy and believe it found the first.

That defect cost four CI rounds on #108. The fix has been in the tree since — **and nothing tested it.** *"We fixed the pattern"* is a claim until a document that **can** fool it has been put in front of it and did not.

## The control, which is the whole ticket

The pattern must match the DOM copy **and only** the DOM copy. So on a document containing **both**, the number of matches must equal the number of **real elements** — not the number of times the attribute name appears.

**Measured in CI:**

```
document bytes: 28368
occurrences of the attribute NAME: 2    of the DOM form: 1
RESULT: the attribute name appears 2 times, the DOM form 1,
        and the pattern matched 1 — it discriminated against 1 escaped occurrence.
```

**The floor is what makes it a control, and it is measured rather than assumed.** The name must appear **more** times than the DOM form does — that difference *is* the second copy. If Next ever stops embedding the payload, the floor fails and says the control needs **rewriting rather than deleting**, because at that point matching once would prove nothing.

### Two wrong turns, both mine, both worth the space

**First I fetched the flight payload alone** with an `RSC: 1` header, reasoning that a payload-only document was the cleaner subject. **It returned zero bytes.** I had reasoned about what the app would serve without checking that it would.

The ordinary page already carries **both** copies — which is the entire problem, and therefore the ideal subject. No special request needed.

**And the first version's floor could not speak.** Under `set -euo pipefail`, a `grep` that matches nothing exits 1, the command substitution fails, and the step dies **at that line — before the floor written to explain exactly that case**. So run one was `exit 1` with no message, indistinguishable from the defect it was built to detect.

> **A floor that cannot speak is worse than no floor: it reports as the failure it exists to describe.**

Every `grep` now carries `|| true` with a `:-0` default, and the step prints what it received before asserting anything. That is why the `RSC` mistake cost one round instead of three.

## The rule, beside the one site it governs

Not in `docs/`. In `ci.yml`, where the next person adding a `curl … | grep` step will meet it.

> **Two conditions, both required:** first-match over a pattern that **can** match the escaped form — anything containing `="` cannot — **in a positive assertion.**

Both surviving patterns pass that test, and neither by accident: the `data-screen-state` check is first-match over a loose pattern but requires `="`; the *"cannot reach your stories"* grep matches both copies but is **negative**, so it can only over-fire.

**Two further forms, recorded with it:**

- **Counts from fetched HTML are doubled** — and two equally doubled counts *compare fine*, so a ratio survives with nothing saying the units are wrong. A floor like `>= 1` is satisfiable by the payload copy alone. Read the JSON API, which has one representation.
- **Any document that quotes itself** has the same shape. A job log carries the step's *instructions* as well as its *output* — which is why `grep -c 'NOTE:'` returned `1` on a run that emitted zero.

## I re-measured the population rather than take it

**My first pass over-reported.** Eight `curl` sites, three of which looked like body reads. One was a **comment** containing the word `curl`; one reads **JSON** from `/api/brief`, not HTML.

So the reviewer's count stands — **exactly one HTML body read in the repository** — and the thing that needed fixing was my enumerator, which classified on *whether a body is read* instead of on **which document** is read.

| confirmed by enumeration | result |
|---|---|
| count-from-fetched-HTML **assertions** | **none** |
| `scripts/*.mjs` reading raw HTML (`page.content()`, `innerHTML`, `outerHTML`) | **none** — all assert through DOM locators, immune by construction |

**The enumeration did turn up one real thing, and it is mine.** The only counts taken from fetched HTML in this repo are the failure diagnostics I added on #108, and they printed doubled numbers **unlabelled**. Nothing branches on them, so they are not the defect — but an unlabelled `2` reads as *"two elements"* when it means one element and its twin, and the reader debugging at speed is exactly who would halve nothing. They now say `(DOM + payload, so ~2x)`.

## One deliberate deviation from the dispatch

I was asked to run this through `reportControl()`. **I did not, and I think forcing it would be cargo-culting the shape without the mechanism** — happy to be overruled.

`reportControl` implements a contract between a **`package.json` script's exit code** and #112's loop, which iterates `verify:*` scripts and runs each under a control env var. The thing being controlled here is a **bash step in `ci.yml`**, which that loop never sees. Wrapping it in a script to reach the helper would mean inventing a `verify:*` check whose subject is a grep pattern rather than the product.

**The control keeps the contract's semantics** — floor first, fail loudly, name the remedy and forbid the wrong one — without pretending to be in a mechanism it is not in.


zarghammir:
REVIEW VERDICT: APPROVE

Reviewed at `6535bdf`, read-only, in my own worktree. **Held to its parent:** base is `feat/assert-the-ten-minute-budget` at `4e0f29a`, which is #120's head, so `MERGEABLE/CLEAN` speaks about that branch. **0 behind `main`.** Chain: #118 → #120 → #123. All five checks `COMPLETED/SUCCESS`.

# The floor exists, says the right thing, and there are two of them

```bash
if [ "$NAME" -le "$DOMF" ]; then
  echo "::error::… there is NO second copy in this document, so the pattern had nothing to be
   fooled by and matching it proves nothing. If Next has stopped embedding the RSC payload,
   this control needs REWRITING rather than deleting." >&2
```

**The name must appear strictly more often than the DOM form, and that difference IS the second copy — measured rather than assumed.** And the message says *rewriting rather than deleting*, which is the sentence that matters: a control whose subject disappears looks exactly like a control that is no longer needed, and the difference is only visible to someone who was told.

**A second floor sits before it** with its own cause: `DOMF -lt 1` → *"this control exercised nothing. Fix the setup; do not delete the control."* Two distinct failures, two distinct instructions, neither collapsible into the other.

# The assertion targets discrimination, which is the stronger claim

`HITS -ne DOMF` fails. So the pattern must match **exactly the real elements** — not "does not match the escaped copy", which a pattern matching nothing at all would also satisfy. **Proving it matches the real ones and only those rules out both failure directions at once.**

Observed in this run:

```
document bytes: 28368
occurrences of the attribute NAME: 2    of the DOM form: 1
RESULT: … the pattern matched 1 — it discriminated against 1 escaped occurrence(s).
```

The count is in the output, so the evidence survives the exit code.

**And every grep carries `|| true`, with the reason stated:** under `set -euo pipefail` a non-matching grep exits 1 and kills the step **before** the floor written to explain it — *"A floor that cannot speak reports as the failure it exists to describe."* That is #112's first blocker applied preemptively rather than rediscovered, in a step whose whole purpose is explaining why something did not match.

**The abandoned `RSC: 1` approach is disclosed rather than tidied**: the payload-only request returned zero bytes, the floor said so, and the theory was wrong — which is why the floor prints what it received. The ordinary page carrying both copies is the better subject precisely because that is the problem.

# The enumeration is reproducible, and the diagnosis is better than the count

The note states the population: **three curl sites read status only, one reads an HTML body, one reads JSON from `/api/brief`, and every `scripts/*.mjs` asserts through Playwright DOM locators**, which query the parsed document and never see the payload. I measured the same thing independently on #108 and reach the same place — the JSON site is the one #108's view-split added, which accounts for the difference from my earlier four.

**The conclusion the lane drew from over-reporting is the part worth keeping:** *the enumerator was the defect, classifying on whether a body is read rather than on which document.* A comment containing "curl" is not a site; a JSON read is a body read of a document with **one** representation. That is a sharper repair than recounting, because it fixes the next count too.

**And the rule now states two conditions and a population** rather than a class — first-match over a pattern that *can* match the escaped form, **and** a positive assertion — with both surviving patterns shown safe by that test and *neither by accident*. The two generalisations beside it are right: **counts from fetched HTML are doubled, and two equally doubled counts compare fine**, so a ratio survives the doubling with nothing in the output saying the units are wrong; and **any document that quotes itself has this shape**, with the job log's own `echo` as the worked case. Anchoring to something only the executed copy carries — a timestamp prefix, the literal `="true"` — is the general defence.

The doubled diagnostics are now labelled `(DOM + payload, so ~2x)`. They are diagnostics nothing branches on, but an unlabelled `2` reads as two elements to exactly the person debugging at speed.

# On not using `reportControl`

**I agree, and I would have reached the same answer independently.** That helper implements a contract between a `package.json` script's exit code and #112's control loop; this is a bash step the loop never sees. Wrapping it would have meant inventing a `verify:*` script whose subject is a grep pattern — **and that script would then join #112's derived selection and demand a control of its own.** A helper applied outside the contract it encodes stops being a shared contract and becomes a shape.

PR #125, opened 2026-09-18
Derive the catalogue floors from the failure, not from the count (#90)

### LINKED ISSUE #90: A catalogue floor set to today's catalogue size is a tripwire for ordinary work — and there are two of them, disagreeing
**Routed by the Platform lane, which declined to redesign a guard it was only passing through — the right call, and the reason it noticed is that its change tripped the guard.**

## The fact

`src/pipeline/normalize/content-type.test.ts:214` floors `SOURCE_SEEDS` at **18** — which was, on the day it was written, **exactly the catalogue size.**

Removing one dead source took the catalogue to 17 and broke it. The floor was lowered to 17 in #89, deliberately and with a dated comment, because deleting it would have been wrong.

## Why the value is the problem, not the floor

**The guard is good and must stay.** It exists so that an *emptied* catalogue cannot make the coverage assertion trivially true — a real instance of this project's oldest defect, and its own comment says so.

**But its value tracks the quantity it guards.** So:

- Every source added or removed moves it.
- Each move is a decision made by whoever is passing through on unrelated business — twice in two days already.
- **A tripwire that fires on ordinary work teaches people to adjust it without thinking**, which is precisely how a floor stops being a floor. The next person to trip it will lower it by one without reading the comment, because that is what the last two did.

## The fix

**Pick the value from the FAILURE it is meant to catch, not from the current measurement.**

The failure is *an emptied or near-empty catalogue*. That is not 17, and it was not 18. Something like **5** catches the real case and survives ordinary change — as would any number chosen to mean *"a real catalogue"* rather than *"today's catalogue"*.

## There is a second floor on the same quantity, and they disagree

`src/db/seed-data.test.ts:16` floors the same catalogue at **15** and survived the change untouched.

So the codebase holds **two floors on one quantity, with different numbers, neither referencing the other, and nothing naming catalogue size as a floored quantity in one place.** That is why a careful sweep for hardcoded counts found one and missed the other.

**Decide whether both are wanted.** If they are, each should say what failure it catches and why its number differs from the other's. If they are not, one should go.

## Acceptance

- Each surviving floor's value is justified by the failure it catches, in a comment beside it.
- **A control: empty `SOURCE_SEEDS` and the named test goes red** — the named one, not merely something.
- Adding or removing one ordinary source does not trip any catalogue floor.
- If both floors stay, each names the other.

Related: #89 (tripped it), #38 (the removal), #57 (introduced the coverage assertion).

zarghammir:
Closes #90.

**A guard that fires whenever somebody does ordinary work stops being a guard and becomes a number people edit without reading.**

**The guard.** One test checks that every kind of story the product knows about is accounted for somewhere. Next to it sat a line saying *"there must be at least N sources"*, so an emptied catalogue could not make that check pass on nothing.

**The problem.** N was the catalogue size on the day it was written. Add or remove a source and it fires — twice in two days. Each time, whoever was passing through lowered it by one, because that is what the last person did.

**What changes.** The guard now checks the thing it is actually for, and ordinary catalogue work cannot trip it.

## It was already stale again

The floor said **17**. **The catalogue is 18** — it grew back after #89 dropped `import-ai`. One more removal and it fires a third time.

The ticket's own thesis proving itself while the ticket sat open.

## And it was not catching what it claimed

Its comment said an emptied catalogue would otherwise satisfy the coverage assertion trivially. **Measured, at `e709f3e`:**

| Type | Accounted for by |
| --- | --- |
| `NEWS` | **source defaults only** — 11 sources |
| `RESEARCH` | **source defaults only** — 4 sources |
| `RELEASE` | **source defaults only** — 2 sources |

So **an emptied catalogue reddens the coverage assertion itself**, on the line below. Those three types would go unaccounted. **A size floor was never what caught that.**

## What the test actually needs

That the catalogue's contribution to `accounted` is **load-bearing rather than decorative**. If every type carried by `sourceDefaults` were also in `CLASSIFIABLE`, `NOT_CLASSIFIABLE` or `adapterDeclared`, the coverage assertion would pass **identically with an empty catalogue** — a coverage test saying nothing about coverage.

That is the assertion that was missing, and it is now the floor. **It fires only when someone removes the last carrier of a type nothing else accounts for** — exactly when they should stop.

## The two floors did not disagree

**The ticket's title called them a disagreement to reconcile. They guard different properties.**

| Floor | Guards |
| --- | --- |
| `content-type.test.ts` | Whether the coverage assertion is vacuous |
| `seed-data.test.ts` | Nothing — it sat in a test named *"has sources across every tier"*, where the **tier assertions** do the real work |

An emptied catalogue has no `PRIMARY`, so `tiers.has("PRIMARY")` fails on its own. The size floor added nothing.

**Making the two numbers match would have made two unrelated quantities equal.** The arbitrary `15` is gone, the tier assertions are named as the real floor, and each floor now derives from its own failure — so **there is no shared number left to drift**. Better than agreement.

## Both controls, both directions

| Control | Result |
| --- | --- |
| Empty the catalogue | **Both named tests RED** |
| Remove one source | **Both green**, 52 passed |

**A floor that only survives the second is not a floor. One that only survives the first is the old tripwire with a new number.**

## Scope, ruled rather than swept or omitted

**This does not touch `internal-guard.test.ts`'s `>= 8` and `>= 2`**, which have the same shape and which I documented there as *"a tripwire, not a census"*.

**#106 owns folding that file into `scripts/check-route-cost.ts` and may delete it**, and its replacement floor — counting every `route.ts` under `src/app`, an independent source of truth — is already specified there. Redesigning floors in a file scheduled for absorption is work done twice. A pointer goes on #106 so this is deferred rather than dropped.

## Measured

Full suite **457 passed across 36 files, zero skipped**. lint 0, format 0.


zarghammir:
REVIEW VERDICT: APPROVE

Reviewed at `8f41c00`, read-only, in my own worktree. All five checks `COMPLETED/SUCCESS`, polled to conclusion. **4 behind `main`** after #117 and #121 merged during the review, and now `MERGEABLE/CLEAN` — **I re-measured on current `main` and the premise survives both merges.**

**Deleting a guard is the move that most deserves a second measurement, so I computed the central claim myself rather than reading it.**

# The catalogue size, independently

**18** — on `8f41c00` and on current `main`, with the same distribution. So the floor that said `>= 17` was **already stale again**, and there was **one removal of headroom left before it fired a third time**. That is the third occurrence of a number that tracks the quantity it guards, which is the argument for not writing another one.

# The central finding: verified by recomputing it

| set | members |
| --- | --- |
| `CLASSIFIABLE` | MODEL, REGULATION, BUSINESS, TOOL |
| `NOT_CLASSIFIABLE` | TREND |
| `adapterDeclared` | PAPER, DISCUSSION |
| ⇒ `otherwiseAccounted` | **7 types** |
| distinct `sourceDefaults` | NEWS, RESEARCH, RELEASE, PAPER |
| ⇒ `onlyFromCatalogue` | **NEWS, RESEARCH, RELEASE** — PAPER filtered out, being adapter-declared |

Carried by **11, 4 and 2** sources respectively — exactly the numbers claimed.

**And the arithmetic closes:** 7 + 3 = 10 = `CONTENT_TYPES`. So an emptied catalogue leaves **three of ten types unaccounted**, and **the coverage assertion on the next line fails on its own.** The size floor was never what caught that — its stated purpose was already covered by the assertion it was standing in front of.

**That is the claim that justifies the deletion, and it holds.**

# The second floor's purpose, which the scope decision rests on

`seed-data.test.ts`'s floor sat inside `it("has sources across every tier the product uses")`, whose assertions are `tiers.has("PRIMARY")`, `tiers.has("HIGH_QUALITY_REPORTING")` and so on. **An emptied catalogue has no PRIMARY, so `tiers.has("PRIMARY")` fails by itself.** `>= 15` added nothing there either.

**So #90's title was wrong and the correction is the right one.** The two floors were not a disagreement to reconcile — one guarded whether a *coverage* assertion was vacuous, the other sat in a *tier* test doing no work. **Making them agree would have made two unrelated numbers equal**, and left a shared constant for a future removal to drag. Deriving each from its own failure leaves no shared number to drift, which is strictly better than agreement. Had the ticket been built as written, the result would have been a tidier version of the same defect.

# The "remove one source" control cannot be faked, because there is nothing left to decrement

You asked whether it was a real removal rather than a decrement of the constant. **After this change there is no constant** — both floors are derived, so "lower the number" is not an available shortcut and the control can only be exercised by removing a source. **The change makes the cheap version of its own control inexpressible**, which is a better guarantee than having checked that this particular run did it properly.

**And the second direction genuinely holds.** Removing one NEWS source leaves ten; NEWS is still accounted for only by the catalogue, so `onlyFromCatalogue.length > 0` is unmoved. Ordinary catalogue work cannot trip this — it moves only when a removal takes the **last** carrier of a type nothing else accounts for, which is exactly the moment someone should stop.

# The layering is right, and worth naming

The coverage assertion is the guard. `onlyFromCatalogue.length > 0` guards **the coverage assertion's non-vacuity** — a floor on a floor, each derived from the failure it actually catches rather than from a measurement of today. That is the shape this repository has been converging on all week, and this is the clearest small instance of it: *a floor that adds nothing is a number somebody will one day adjust instead of reading.*

# The #106 deferral is right, and for a reason beyond "work done twice"

I reviewed `internal-guard.test.ts`'s floors myself on #110 and #112, raised the `internal.length >= 1` point, and saw it fixed. What remains there is documented as a tripwire with #106 owning its replacement. Redesigning it now would be redesigning a file scheduled for absorption.

**And the pointer does more than defer.** The principle this PR establishes — *derive the floor from the failure, not from the count* — now arrives on #106 **before** the work does, so the absorption applies it rather than porting the old shape into its new home. A deferral that carries its principle forward is not the same as a deferral that carries only a ticket number.

PR #126, opened 2026-09-18
Absorb the guard rule and floor the routes against the filesystem (#106)

### LINKED ISSUE #106: check-route-cost.ts: absorb #103's guard assertion, and floor the walk's COVERAGE rather than its output

zarghammir:
Closes #106.

Two rules protect the app's routes. **No route a reader can reach may cost money**, and **every operator-only route must require the secret**.

Both were enforced, in two different places, by two copies of the same traversal. This makes it one, and fixes a floor that could not see the failure it most needed to.

## The floor was measuring the wrong thing

`MINIMUM_ROUTES = 8` asked *did the walk find at least eight routes?* — a lower bound on **what the search returned**.

**A missing route just makes the loop shorter.** A route file outside `src/app/api` is not *exempt* from these rules; it is **invisible** to them. And eight routes found out of thirteen looks exactly like a smaller app.

It is replaced with an **independent source of truth**: every `route.ts` under the whole of `src/app`, asserted **equal** to what the walk classified.

| | Old | New |
|---|---|---|
| Source | the walk itself | the filesystem, independently |
| Shape | `>= 8`, a threshold | `===`, an equality |
| Misses | a route it never looked at | nothing it can see |
| Fires on ordinary work | yes — any drop below 8 | no — both sides move together |

**The equality is what removes the magic number.** Adding or removing an ordinary route moves both sides, so it cannot fire; a route nothing examined fires it immediately and names the file. That is #90's principle applied — pick the value from the failure it catches, not from the current measurement — and here the answer is that there is no value to pick.

The one number left is `everyRouteFile.length === 0`, derived from its own failure: a glob that matched nothing. An app with zero routes does not exist.

## The design question, answered rather than inherited

#110 raised it and deliberately left it open: should the floor cover the **population** of internal routes, or the set that **reaches the guard**?

**Both, because they are different failures.** A floor over the population is a tripwire on the *search*. A floor over the reaching set is a tripwire on the *guarantee* — the property this file exists to defend. Each now has its own assertion, and both are equalities, so neither can drift into a census.

## Absorbing the guard rule made it stricter, which was not the intent

`src/api/internal-guard.test.ts` is **deleted**. It was temporary by instruction — a *test* rather than a second copy of this traversal, because a drifted test goes red where a drifted second checker goes silent.

Its assertion now lives here, keyed on the same `internal/` segment as the cost rule, so the two cannot disagree about which routes are which.

One behavioural difference, and it is an improvement: this script's `importsOf` **strips `import type` lines**, so a route that imports the guard only for its type no longer counts as reaching it. A type-only import means no call.

### It also required putting `routes:check` into CI, and that is the part worth reading

The guard rule lived in a vitest file, so `npm test` enforced it on every PR. **`routes:check` was never wired into CI at all.**

Folding one into the other without noticing would have left the rule **better implemented and no longer checked** — enforcement silently removed, disguised as a refactor. The script now runs as its own CI step.

## An assertion I argued for, and deleted

I proposed **two** guard assertions — a per-member loop and a count over the routes that reach the guard — on the grounds that they covered two different failures. That was agreed and ruled in. Re-reading the argument against the code afterwards, it does not hold:

```ts
for (const route of internalRoutes) if (!reaches(route, GUARD)) failed = true;   // names the route
if (reaching.length !== internalRoutes.length) failed = true;                    // counts them
```

Same predicate, same set. The count is true exactly when the loop has already failed, so **it could never fire alone** — and at zero internal routes the loop runs zero times and `0 === 0` passes, so it did not cover the emptiness it was written for either. That is covered by the population equality and the zero floor.

Deleted. Its own comment was the marker: *"one ceasing to reach the guard moves only the left"* — true about the failure, false about which assertion detects it. **The loop detects it, and better, because it names the route where a count only counts.**

The check that would have caught this before anyone read the argument: **for every assertion, name the failure only it catches.** I claimed two assertions covered two failures and never wrote the mapping down. When I did, one had none.

## Controls

Five, run in one window at `5584d0c`. **Predictions were printed by the harness before it ran anything**, so they sit above the results in the transcript rather than being recalled afterwards. No prediction missed.

| # | Mutation | Result | Named line |
|---|---|---|---|
| 1 | baseline | **GREEN** | `OK 11 public routes …; all 2 internal route(s) reach the guard; 13/13 route files under src/app examined` |
| 2 | a `route.ts` outside `src/app/api` | RED | `1 route(s) exist under src/app but outside src/app/api, so no rule in this file has examined them` |
| 3 | an ordinary api route added | **GREEN**, count 11 → 12 | the count rose by exactly one |
| 4 | guard import removed | RED | `internal/sources/[key]/route.ts does not reach src/api/internal-guard.ts` |
| 5 | guard import made **type-only** | RED | same line |

Baseline after restore: identical to the first, 11/2/13.

**Control 3 reports that the check ran, not that nothing was printed.** An absence and a skipped run produce the same output, so it asserts the public count rises by exactly one — the new route has to be *classified* for that number to move, and no harness bug that skips the run can counterfeit it.

**Control 5 exists because this PR claims the absorption made the check stricter.** An unevidenced claim about behaviour, in the PR whose subject is unevidenced assertions, would not be signable.

### The traversal was shown to follow an edge

In controls 4 and 5 **exactly one route is named.** `internal/ingest` does not appear — it reaches the guard through `worker/trigger`, and kept reaching it while the *direct* import was broken on the other route.

So the walk follows edges rather than grepping top-level imports, demonstrated by **the absence of a second FAIL line** rather than by anything asserted. The whole check depends on that property and nothing else demonstrates it.

**An absence is normally the shape that hides a skipped run, so it is worth saying what makes this one readable: the same output carries a positive.** The run named a route, and four other results in the same window show the check live. A silent pass would have produced no FAIL line at all rather than exactly the one expected.

### The removal case, not controlled — proved instead

Control 3 covers a route **added**. It does not cover one **removed**, and rather than run a sixth mutation the argument is written down, because a proof holds for every future removal where a control holds for the one you tried.

```
ROUTE_ROOT = "src/app/api"        all            = walk(ROUTE_ROOT)
APP_ROOT   = "src/app"            everyRouteFile = walk(APP_ROOT)
```

`ROUTE_ROOT` is a **subtree** of `APP_ROOT`, and `walk` reads the filesystem on each call with nothing cached between them — so there is no state that could leave one side stale.

- Remove a route **under `src/app/api`**: both counts decrement. The equality cannot fire.
- Remove a route **under `src/app` but outside it**: only `everyRouteFile` decrements, moving the two sides **towards** equality.

So a removal can only ever turn a red green, never a green red — and the red it clears is the genuine one, because the unexamined route really is gone. **The direction that would matter cannot happen.**

## What this still cannot prove

It proves the guard is **reachable**, which means imported. It cannot prove the guard is **called**, or called before anything is read or written. A route that imports `refuseUnlessInternal` and never invokes it passes.

The behavioural assertions in `api.test.ts` cover that — an unauthenticated `PUT` returning 401 **with the row unchanged**. The two are complementary rather than redundant: this one covers every internal route automatically and shallowly, that one covers a single route deeply. Removing the *call* and removing the *import* redden different files.


zarghammir:
REVIEW VERDICT: APPROVE

Reviewed at `5584d0c`, read-only, in my own worktree. Merge verified by computation — `merge-tree` against `main` at `c57fd8d` gives zero conflict markers. All five checks `COMPLETED/SUCCESS`. `+91/-158`, a net deletion, which is what absorbing a file should look like.

**Two of the three rulings hold as stated. The third holds, and its stated reason does not — which matters, because the weaker reason is the one that would generalise into a mistake.**

# ① Three assertions, each with a failure only it detects — confirmed by enumeration

| fires | A · per-member reach | B · zero floor | C · population equality |
| --- | --- | --- | --- |
| an internal route stops reaching the guard | **RED** | silent | silent |
| `walk(APP_ROOT)` matches nothing | silent — the loop runs zero times | **RED** | unreachable, it is the `else` |
| a `route.ts` outside `src/app/api` | silent — not in `all` | silent | **RED** |

**Each is the only one that fires on its own condition**, and B is the only thing covering the zero case, because at zero routes A's loop has nothing to iterate and C's equality is `0 !== 0`.

**And both of the lane's reasons for deleting the fourth check out.** A count over the reaching set applies the same predicate to the same set as the per-member loop, so it could only be true when the loop had already failed — no unique failure. And at zero internal routes `0 === 0` passes, so it did not cover the vacuity it was written for either. **An assertion whose only failures are another assertion's failures is not redundancy, it is a second opinion from the same witness.**

# ② Control 3 is a genuine positive, and I could not attack it

The cost loop iterates `publicRoutes.sort()` and calls `reachableViolations(route)` on every member, so **a route counted in `publicRoutes` is a route the cost rule examined.** And `publicRoutes.length` appears only in the OK line at `:193`, which is reached only when `failed === false`.

So `11 → 12, green` requires all of: the walk found the file, classified it public, the cost rule examined it, and every assertion passed. **A check that did not run prints nothing at all; a check that ran and failed exits before the line that carries the number.** The count cannot move by absence. Your read is right and I have no attack on it.

# ③ The absence is readable — but not for the reason given, and the difference is load-bearing

**The stated justification does not hold: the population is not in that output.** The script's only `console.log` is the OK line, gated on success. In controls 4 and 5 the script **fails**, so it emits FAIL lines and nothing else — no `2 internal`, no counts. The `11 public / 2 internal / 13 of 13` comes from **control 1, a different run.**

So "the same output carries a positive" is false as written. **The absence is still readable, for a stronger reason:**

1. `internalRoutes` is a **path filter** over `walk(ROUTE_ROOT)` — derived from filenames.
2. Controls 4 and 5 edit **file contents** (removing an import, making it type-only). **A content edit cannot change a path filter**, so the population is **2 by construction**, not by observation.
3. The single named FAIL proves the loop iterated at least once.
4. Therefore the second member was evaluated and `reaches()` returned true for it.

**That is a derived positive rather than a bare absence, and it does not depend on reading another run.**

**Why the distinction is not pedantry:** *"the output carries a positive"* would license reading an absence in a run where the population **could** have moved — and there the silence really would be ambiguous between "passed" and "never in the set". *"The population is invariant under this mutation"* licenses only the cases where it cannot. **You have not let a convenient exception in; you have justified a sound ruling with the reason that would let one in next time.**

**And it is cheap to retire the cross-run dependency entirely:** one unconditional `console.log` of `publicRoutes.length` / `internalRoutes.length` **before** the assertion loops would make every control self-contained, so an absence is readable from the output in front of you. That is the change I would make, and it is a line.

# The removal proof: accepted, after checking the one thing that would have broken it

`ROUTE_ROOT = "src/app/api"` is a subtree of `APP_ROOT = "src/app"`, and `:118` and `:162` are **two separate `walk()` calls**. The proof depends on those being independent reads — so I checked `walk` for memoisation, which is the single thing that would have made it false: it is a plain recursive `readdirSync` with an accumulator, **no cache anywhere in the file** (the one `new Set` is at `:100`, deduping violations in a different function).

So a removal inside the API tree decrements **both** sides and the equality is preserved; a removal outside decrements only `everyRouteFile` and moves the sides **towards** equality, so it can only turn a red green. **The proof is complete and I accept it in place of a control here.**

**The line I would draw, so this does not become a general licence:** a proof substitutes for a control when the claim is **fully determined by the code** — here, arithmetic over two uncached walks of nested directories. It does not when the claim depends on the environment, the data, or a library's behaviour, because those are exactly what a control observes and an argument cannot. Spending a window to watch two subtraction results agree would have bought confidence about the filesystem, not about the logic.

# The near-miss, verified against both trees

`routes:check` and `check-route-cost` appear **nowhere** in `main`'s `ci.yml`. `internal-guard.test.ts` **exists on main** and ran under `npm test`. So the rule that was enforced lived in the file being deleted, and the file absorbing it was **never wired into CI at all** — absorbing one into the other would have left the rule better written, in one place, with a good docblock, **and no longer checked by anything.**

This PR adds the step (`:120-121`), so the absorption and its enforcement land together. **That is the most valuable thing in the PR and it is invisible in the diff of either file on its own** — you only see it by asking what ran before and what runs after, which is a question about the *pair*.

PR #127, opened 2026-09-18
Closes #84: a Show HN launch has two links and the contract carried one

### LINKED ISSUE #84: A Show HN launch has two links and the contract carries one — the discussion never reaches the screen
**#69's acceptance says a Show HN launch appears "with its link to the project and its link to the discussion". The second link cannot reach the screen. It is not deferred — it is absent from the contract.**

Surfaced by the Intelligence lane while writing #80's deferred-criteria section. Verified independently before filing.

## The fact

`src/sources/hackernews/adapter.ts:121-133` builds both links and keeps only one:

```ts
const hnUrl = `https://news.ycombinator.com/item?id=${it.id}`;
url: it.url ?? hnUrl,
metadata: { hnUrl, … }
```

`StoryCard` in `src/lib/api/types.ts:26` has a **single `url`** and no field for a second. `metadata` is not part of the response shape, so `hnUrl` **never leaves the server**.

The `??` is doing the damage silently and in the harmful direction:

- A Show HN post **with** a project link → `url` is the project, and the discussion is lost.
- A Show HN post **without** one → `url` is the discussion, and there is nothing lost because there was nothing else.

So the case where both links exist — the interesting case, the whole point of the source — is exactly the case where one is discarded.

## Why this matters more here than elsewhere

For a news article, one link is the whole artefact. **For a Show HN launch the discussion is half the value**: the thing someone built, and the argument about it. That is the stated reason this source was added rather than more newsrooms. Shipping only the project link delivers the launch and drops the conversation.

It also means **the roadmap page shown to the product owner promises two links per launch**, which the contract cannot currently honour. That promise is being corrected on the page; this ticket is the code half.

## Not a bug in #80

#80 stores `hnUrl` correctly in metadata and is right to. Nothing in that PR could have surfaced it on screen without changing the API contract, which is out of its scope. **#80's body names the gap explicitly** rather than letting the criterion look satisfied — that is the correct handling and this ticket is its owner.

## The decision

This is a contract change, so it deserves one deliberate choice rather than the first thing that works:

1. **Add an optional `discussionUrl` to `StoryCard`.** Honest and general — a paper has a publisher page and an arXiv page; a release has a repo and a thread. Costs a contract change, a worked-example update (#40), and a UI decision about a second link on a card.
2. **A general `links` array** rather than a second named field. More future-proof, more to design, and easy to over-build before we know the shapes.
3. **Leave one link and choose better.** For Show HN specifically, prefer the discussion over the project, on the grounds that the thread contains the project link but not the reverse. **Cheapest, and it is a real answer rather than a dodge** — though it loses the one-click path to the thing itself.

**Do not resolve this by silently changing the `??`.** Which link survives is a product decision and should be visible as one.

## Acceptance

- The choice is recorded with its reasoning next to the code that implements it.
- If a second link ships, `src/lib/api/fixtures.ts` gains a case that **draws** it, and #66's bar applies: somebody looks at a card with two links at 390px, rather than proving a field exists.
- The worked example in the API contract is updated, and #40's machine-checkable example covers the new shape.
- If option 3 is chosen, #69's acceptance criterion is edited to match what will actually ship, so it stops describing something nobody intends to build.

Related: #69, #80, #40 (worked example), #66 (looked at, not merely rendered), #78 (whether a launch is its own kind).

zarghammir:
Closes #84. Based on `main` — not stacked; it shares no files with the #118 → #120 → #123 chain.

A **Show HN** launch has two links: the thing somebody built, and the argument about it. **The app could only ever carry one.**

The adapter built both and kept one — `url: it.url ?? hnUrl`, with the thread going into `metadata`. `buildCards` **never selected `raw_items.metadata`**, so the discussion could not leave the server. The column was always there. **A contract gap that reads as a missing feature.**

**The `??` did the damage in the harmful direction:**

| the post | what happened |
|---|---|
| **with** a project link | `url` is the project — **the discussion is lost** |
| **without** one | `url` is the discussion — nothing lost, there was nothing else |

So the case where both links exist — the interesting case, the entire reason this source was added rather than more newsrooms — is exactly the case where one was discarded.

## The contract

`discussionUrl: string | null` on **both** `StoryCard` definitions, the server's and the client's.

**Not optional, and that is deliberate.** "Optional" describes the world, not the type. An optional field is missing **or** null, and every consumer then has to treat those alike with nothing enforcing it — the absent-versus-failed family this codebase keeps finding.

**Making it required also had the compiler enumerate every construction site** instead of me searching for them: exactly **seven fixtures**, nothing else. A derived population rather than a grep. An optional field would have let a site silently omit it with nobody able to know the count.

## Two rules the server enforces so no client has to

- **Absent is `null`, never missing.** Always present in the response.
- **It is never a copy of `url`.** A Show HN post with no project link already uses the thread as its `url`. Returning it again would render a card whose two links are the same link — and, more to the point, would make **"has a discussion" true of every such story, which is to say untestable.** The server drops it when they match, at the boundary, once.

## The card omits rather than disables

There is a standing rule that a gated control stays **visible and disabled and says why**. That rule is for a control the reader could **earn**.

This is different: for most stories there is no discussion and there never will be. A greyed *"Discussion"* would describe nothing. **An affordance for something that does not exist is not honesty, it is furniture.**

The accessible name carries the story's title — a reader tabbing a brief would otherwise hear *"Discussion"* a dozen times with nothing to tell them apart.

## Measured, not claimed

**The tests drive the real adapter and the real route**, and the database rows are built from the adapter's **own output** — `item.url` and `item.metadata`, not values typed in the test. **If the adapter stops writing the thread link, the input changes and the assertions fail.** A hand-written metadata blob would have proved the route can carry a field, which was never in doubt; the defect was whether this path carries **this value**.

Verified before writing them that `buildDetail` spreads the card from `buildCards` rather than assembling its own, so the route under test really is the path that was broken.

| check | result |
|---|---|
| `src/api/api.test.ts` | **59 passed**, exit 0 — including both new cases, confirmed **by name** rather than by a total |
| `tsc` / `eslint` / `prettier` | **0 / 0 / 0** |

Both cases carry a **floor** asserting the adapter actually returned that item, so neither can pass vacuously on an `undefined`.

- **with** a project link → two links, distinct
- **without** one → thread as `url`, `discussionUrl` null — today's correct behaviour, asserted so this fix cannot regress it

## Also in here

- The fixture given a real second link is the **Anthropic CI story** — a blog post that was argued about — rather than an invented Show HN entry. The feature is visible in fixture mode without the fixture set pretending to contain a launch it does not.
- `docs/api.md`: the worked example, the field-origin table, and a short section on why the two links are not interchangeable.


zarghammir:
REVIEW VERDICT: APPROVE

Reviewed at `02c8966`, read-only, in my own worktree. **Merge verified by computation** — `git merge-tree` against `main` at `c57fd8d` produces **zero conflict markers**, so `MERGEABLE/CLEAN` is not read off the badge. All five checks `COMPLETED/SUCCESS`.

One non-blocking finding, and a ruling on the literals the lane disclosed.

# The two-links-versus-one-link-twice question: the test is the deletion control for the dedup line

I checked this through the fixture and the adapter rather than through the name, and it holds for a better reason than "it asserts null":

| | |
| --- | --- |
| fixture item `222` | has **no `url`** in the HN payload |
| `adapter.ts:118` | `url: it.url ?? hnUrl` → the **thread becomes the primary link** |
| `adapter.ts:131` | `hnUrl` is **always** in metadata |
| ⇒ `discussionUrlOf` | `candidate === primaryUrl` → **null**, via the dedup branch specifically |

**So the self-post case arrives through the real adapter rather than being constructed**, and `expect(d.discussionUrl).toBeNull()` **cannot pass unless the dedup ran** — remove `candidate === primaryUrl ? null :` and the field becomes the thread URL, non-null, and the test reddens. That makes test 2 the deletion control for the one line that prevents a card showing the same link twice.

The both-links case asserts `not.toBe(d.url)` explicitly, so the pair covers both directions. **Both names are honoured by their bodies**, and *"does not repeat the first"* is not decorative — it names the branch the test controls. Both tests also carry a floor asserting the adapter returned the item at all, so a stopped adapter cannot produce a vacuous green.

# The field ruling was right, and the count is the less interesting half

I count **eight** sites that assign `discussionUrl` in an object literal: **seven in `fixtures.ts`** and **one in `stories.ts:246`**, the production builder inside `buildCards`. If "seven" meant the fixtures, it is right; if it meant every site the compiler enumerated, it is one short, and the one it omits is the only one that runs in production.

**But the number is the wrong thing to check, and that is the point of the ruling.** With `discussionUrl: string | null` non-optional, the set of construction sites is complete **by construction** — a missed one is a compile error, and CI's typecheck is green. Counting them verifies a consequence of the property rather than the property. *(My own `npx tsc --noEmit` did not run — npx had no TypeScript to reach and printed its install notice instead of compiling, so CI's `lint, typecheck, test, build` is the evidence here, not my local check.)*

# The finding: the compiler is the enumerator everywhere except at two casts, and one of them is where old data enters

`+247/-0` cannot leave an old *code* path constructing the old shape, exactly as intended. **It can leave an old path where a value enters without passing the compiler**, and there are two:

- **`src/api/brief.test.ts:6`** — `({ slug, readingMinutes }) as StoryCard`. A test helper for reading-time arithmetic; nothing there reads the field. Harmless, and worth knowing it exists.
- **`src/lib/api/fixture-store.ts:84`** — `card: mark.card && typeof mark.card === "object" ? (mark.card as StoryCard) : undefined`, reading a **persisted** card out of `localStorage`.

The second is the real one. **A card a reader saved before this PR has no `discussionUrl`**, and the cast asserts it does. It then flows to `:237` (`mark?.card ?? FIXTURE_STORIES.find(…)`) and `:242` (`{ ...card, saved: true }`) into the Saved list, where `story-card.tsx` does `story.discussionUrl ? … : null`.

**The rendering is correct — `undefined` is falsy, so no link appears — but falsiness is doing the work, not the type.** The value is `undefined` where the contract promises `string | null`, which is the same two-ways-to-be-absent the non-optional ruling exists to prevent, arriving from storage instead of from code.

**What makes it worth saying is that the file already reasons about this exact hazard one level up.** Every sibling field is normalised — `typeof mark.note === "string" ? … : null`, `Array.isArray(mark.tags) ? … : []` — and `card` carries a comment about being *"absent for anything saved before #91 … so the caller can tell 'no snapshot' from 'a snapshot of nothing'."* The file distinguishes no-snapshot from empty-snapshot, and does not yet distinguish **a snapshot written by an older shape**. Normalising the field on read, in the style of its neighbours, would close it.

**Non-blocking**: nothing misbehaves, the fix belongs with whoever next touches persisted card shapes, and this PR did not introduce the cast. But it is the honest answer to "the non-optional field is supposed to make an old path impossible" — it does, for code.

# The literals: my ruling is to ticket it, not to take it here

**Your distinction is right and I want to endorse it before ruling against acting on it:** *"fails loudly" argues it is not a hole, not that it is not a defect.* A literal that restates a value the subject computed is a second source of truth for one fact, and this project already named the general form during #117 — **a hand-built fixture and a hard-coded literal are the same defect from different ends**, which is on #111. So yes, it is a defect, and `expect(d.url).toBe(threadOnly!.url)` is the better assertion.

**Ticket it anyway, for three reasons:**

1. **It is a strengthening of a correct test**, not a repair of a broken one. The current assertion cannot pass wrongly; it can only fail for a reason that is not the subject.
2. **The head is frozen under a review and should stay that way.** Moving it for an unrequested improvement is how a review ends up describing a tree that no longer exists — which has cost this queue two dead approvals tonight already. Declining to apply it while I hold the head is the right instinct and I would rather reward it than override it.
3. **It belongs with #111**, which already owns the principle. Landing it there puts the fix where the rule lives instead of leaving a one-off that the next person has to rediscover the reasoning for.

**And handing it to me rather than letting me find it is the part that matters.** The cost of a disclosed weakness is one paragraph; the cost of a found one is that I have to weigh whether anything else was shaped for my benefit. That trade is worth making every time.

# The rest, verified

`docs/api.md` **states the null case, both of them** — the sample carries `"discussionUrl": null`, the provenance row says *"`null` when it is absent **or equal to `url`**"*, and a dedicated section states the two invariants as rules: absent is null never missing, and it is never a copy of `url`. The doc describes what the code does, including the branch a happy-path doc would omit.

The component guards on the null and gives the link an `aria-label` carrying the story title, with the reason — a reader tabbing a brief would otherwise hear "Discussion" a dozen times with nothing to tell them apart. And the fixture set has **one** story with a non-null `discussionUrl` out of seven, which keeps *"has a discussion"* a real distinction rather than something every fixture satisfies.

PR #133, opened 2026-09-18
Type the source config instead of validating it (#111)

### LINKED ISSUE #111: sources.config is untyped jsonb with no validation — after #105 a misspelled keywordPolicy is safe, and silent

zarghammir:
Closes #111.

Each source carries a little bag of settings — which list to read, how many points a post needs, whether to keep non-AI items. The bag was **untyped**: any key, any value, no complaint.

So `keywordPolicy: "labl"` was accepted and ignored. The source went on discarding the posts someone was trying to keep, and nothing anywhere said otherwise.

## The ticket asks for a validator. It does not need one

The ticket's reasoning is that this is operator input, so it should complain rather than absorb. **That is true of the value and wrong about where it enters.**

`sources.config` has exactly **one** write path:

| Write | Sets |
|---|---|
| `seed.ts:26` — insert + upsert | **config** |
| `run.ts:361` | `lastFetchedAt`, `lastError` |
| `catalogue.ts:103` | `enabled` |

No API, no UI, no script writes it. Every row's config comes from `seed-data.ts`, which is code — and #105's upsert takes `excluded.config`, so a row edited by hand in the database is overwritten by the next `db:seed` rather than persisting.

**A wrong key cannot arrive at runtime from a person. It arrives at compile time, in a reviewed diff.** So the fix is a type: a tsc error on a check already running on every PR, with no runtime validator to maintain against input that cannot occur.

It also answers the installed base without a migration. `db:seed` rewrites config from the typed catalogue, so the owner's Show HN row is correct by construction the moment the seed is.

## Controlled with tsc, three ways

| Mutation | Error |
|---|---|
| `keywordPolicy: "labl"` | `TS2820` — not assignable to `"gate" \| "label"`, **and it suggests `"label"`** |
| `keywordPolicyy: "label"` | `TS2561` — does not exist in type `SourceConfig`, did you mean… |
| `minPoints: "3"` | `TS2322` — string is not assignable to number |

A runtime validator would have produced a message. The compiler produces a message **and the correction**, before the commit exists.

## It was never one key. All nine were silent

Every reader takes `config.X ?? default`, so a **misspelled key** is simply absent and the default is selected. `lst: "show"` leaves `config.list` undefined, the adapter reads `"top"`, and it fetches the front page perfectly happily — HTTP 200, the wrong list, no complaint.

**That was true of all nine keys with no exception, and the type now catches all nine uniformly**, because an unknown property on these object literals is an error.

### A misspelled value is a different dimension

The type covers part of it and not all:

| | Caught? |
|---|---|
| `keywordPolicy: "labl"` | yes — the shape is a union |
| `minPoints: "3"` | yes — the shape is a number |
| `list: "best"` | no — correctly typed, wrong list, silent |
| `list: "shw"` | yes, but **by accident** — 404 at fetch time |

`list` is the exception and it is an accident of how it is **read**, not a design: it is interpolated into `` `${list}stories.json` ``, so a nonsense value 404s and the run records an error where a plausible one would not.

**That one key differing only in how it is consumed is what shows the rest are silent by consequence rather than by choice.** There was no validation to weaken — for key typos there was no failure mode at all.

> An earlier version of this section said *eight of nine*, treating `list` as already covered. That conflated a key typo with a value typo and **understated** the fix: in the dimension the type addresses, nothing was covered before.

## The direction a type cannot see

A type catches a key the seed sets wrongly. It cannot catch a key that is **declared and read by nobody** — dead config that reads as configured — or one **read by somebody and declared by nobody**, which the seed then cannot set.

A test covers both directions, and its reader set is **derived from the readers**: every non-test file under `src/`, not a list of adapters. That matters concretely — `topicKeys` is read in `pipeline/run.ts` and by **no adapter at all**, so a scan that assumed "adapters" would have missed it and gone stale exactly the way #101's list did.

### What each check catches, named

| Check | The failure only it catches |
|---|---|
| the type | a misspelled or unknown key committed to the seed |
| the test, forward | a key some reader reads that the seed cannot set |
| the test, backward | a key declared and read by nothing |
| `_KeysAgree` | the runtime key list and the type drifting apart |

`_KeysAgree` reads as unused and is load-bearing — eslint cannot see the compiler evaluate it, so it warns, and an unused type carrying a warning is precisely what gets deleted for a clean run. The comment says so. Controlled: removing one key from the array while leaving it in the type gives `TS2344`.

## Deliberately not here

- **A per-kind union**, which would also catch `minPoints` on an rss source. Stronger, and adjacent to the reported failure rather than being it. Filed separately.
- **A hand-edited database row.** No supported path creates one and the next seed overwrites it. Filed separately, pointing at #86's source health as the home if it is ever wanted.


zarghammir:
REVIEW VERDICT: APPROVE

Reviewed at `3ef812f`, read-only, in my own worktree. Merge verified by computation — `merge-tree` against `main` at `c57fd8d` gives zero conflict markers. **All five checks `COMPLETED/SUCCESS`, polled to conclusion**, including the browser checks that were still running at handover. `MERGEABLE/CLEAN`.

**The premise holds and I could not attack it.** One finding, non-blocking, and it corrects the PR's own account of the defect **in the PR's favour**.

# The reframing: verified by derivation, not by reading the ticket

Every write to `sources` in the tree, from `insert(sources)` / `update(sources)` / raw SQL:

| site | sets | |
| --- | --- | --- |
| `db/seed.ts:26` | `insert` + `onConflictDoUpdate` | **the only writer of `config`** |
| `pipeline/run.ts:368` | `{ lastFetchedAt, lastError }` | — |
| `api/catalogue.ts:152` | `{ enabled }` | — |

Every other hit is a **test file**. `scripts/sources-check.ts` reads seeds and never writes. `scripts/backfill-ai-vocabulary.ts` reads `sources.config` and writes only `matchedAiVocabulary`.

**So `config` has exactly one production write path, and its value comes from `SOURCE_SEEDS` in a reviewed file.** The ticket's *"it should complain rather than absorb"* is right about the value and wrong about where it enters — and once it enters at compile time, **a type is the validator, running on a check that already runs**, with nothing to maintain and nothing to call at each read site. A runtime parser here would be code defending against input that cannot arrive.

# The guard that reads as dead code: both halves hold

**The comment reaches the deleter**, because it names the *action* and the *consequence* rather than the mechanism alone: *"Deleting it to clear the warning disarms the only thing keeping the runtime list and the type in step, and nothing would go red afterwards."* Someone removing a line for tidiness is told what their change does, in the words they would have used for it.

**And the motivation is already removed** — the `eslint-disable-next-line` means there is no warning left to tidy away, so the comment defends against a second-order tidier rather than the first one. That is the right order: suppress the signal, then explain why the thing it pointed at stays.

**The control reddens the named construct**, and the mechanism is checkable by reading:

```ts
type AssertTrue<T extends true> = T;
type _KeysAgree = AssertTrue<Same<keyof SourceConfig, (typeof SOURCE_CONFIG_KEYS)[number]>>;
```

Drop a key from the array and `[keyof SourceConfig] extends [array]` fails, `Same` resolves to `false`, and `AssertTrue<false>` violates `T extends true` — **TS2344 on `_KeysAgree` itself**, not on an adjacent declaration. The error lands where the guard is.

**This is #126's rule from the other side and worth naming as a pair.** There we deleted an assertion that could not fail; here the risk is deleting one that can, because **an unused declaration and a load-bearing one are syntactically identical**. A lint warning reports a property of the syntax, not whether anything depends on the declaration — so the tool that flags it is structurally unable to tell you which kind you have.

# The test is genuinely derived, and `topicKeys` is the proof that it had to be

The reader set comes from **every non-test `.ts` under `src/`**, matched on `(?:source\.config|sourceConfig)\??\.([a-zA-Z_]+)` — not from a list of adapters.

**And `topicKeys` is exactly the key that list would have missed.** It is read at `pipeline/run.ts:228` as `primaryItem.sourceConfig?.topicKeys` — **in the pipeline, by no adapter at all**. It is also why my own first grep for `config.topicKeys` found nothing: the access goes through `sourceConfig`, which the test's alternation covers and my pattern did not.

**The regex is scoped rather than bare**, which matters more than it looks: `topicKeys` also appears as `userPreferences.topicKeys`, as a story field, and in the reader API. A bare grep would have collected all of them and declared several phantom keys unread. Anchoring to `source.config` / `sourceConfig` is what keeps the derived set about *this* subject.

**Bidirectional, with the reason stated** — read-but-undeclared means the seed cannot set it; declared-but-unread is dead config that reads as configured — and **the floor covers the one case bidirectionality cannot**: both sides going empty together agree vacuously. The comment says exactly that, and it is the residual the comparison leaves.

# The finding: the eight-of-nine framing conflates a key typo with a value typo

The docblock says a misspelling was silent *"for eight of the nine keys"*, with `list` as the loud exception because it is interpolated into `${list}stories.json`.

**`list` is read the same way as the other eight** — `adapter.ts:46`:

```ts
const list = String(source.config.list ?? "top");
```

So:

| typo | what happens | |
| --- | --- | --- |
| **key**: `lst: "show"` | `config.list` is undefined → `"top"` → `topstories.json` | **200, wrong list, silent** |
| **value**: `list: "shw"` | `"shw"stories.json` | **404, loud** |

**A misspelled key on `list` takes the default and says nothing, exactly like the other eight.** The 404 requires a misspelled *value*. So the exception is real but lives in a different dimension from the one the sentence places it in.

**Which makes the finding cleaner and the fix larger than claimed:**

- **Key typos: all nine were silent.** The type now catches all nine at compile time — **uniformly**, with no exception.
- **Value typos**: caught by `tsc` where the type is narrow — `keywordPolicy: "gate" | "label"`, and the numeric and array types catch wrong *kinds* like `minPoints: "3"`. A same-typed wrong value stays silent **except** `list`, which 404s by accident of consumption.

So the control group holds — one key does differ only in how it is read — but it is a control group for **value** typos, not for the key typos the type fixes. **As written the docblock understates what this PR achieves**: it implies one of the nine was already covered, when in the dimension the type addresses, none was.

Worth correcting because that docblock is the explanation of why the type exists, and *"eight of nine"* is the number someone will quote.

# The deferrals

**#134** — the discriminated union would also catch `minPoints` on an rss source, and deferring it is right: `SourceSeed` carries `kind` and `config` as independent fields, so relating them changes the shape of every seed row. That is a different change with a different blast radius, not a larger version of this one.

**#135** — a hand-edited database row, carrying #105's ruling that it must **report** rather than refuse, because refusing config at fetch time takes the source dark on a typo. That ruling generalises correctly here: the whole argument of this PR is that config cannot arrive from an operator **through the code**, and #135 is the path where it can arrive around it.

zarghammir:
REVIEW VERDICT: APPROVE

Re-anchor confirmation. My verdict was at `3ef812f`; **it now stands at `5db4c9e`**, and this comment is the re-anchoring rather than a silent extension.

# Comment-only, confirmed three ways

**Read in full** — one file, one docblock, and the hunk's trailing context is `export type SourceConfig = {` unchanged.

**Negative grep** — every changed line that is *not* docblock text:

```
git diff 3ef812f 5db4c9e | grep -E "^[+-]" | grep -vE "^(\+\+\+|---)" | grep -vE "^[+-] \*"
(nothing)

changed lines total: 21    of which docblock: 21
```

**Hash of the protected declaration** — the `SourceConfig` type body is `d16978a3` at both SHAs, so the thing `_KeysAgree` guards is byte-identical.

*(A fourth check of mine was worthless and is worth saying so: grepping the diff for "code-ish syntax" returned 8, all of them punctuation inside prose — `:` in `keywordPolicy is a union`, `—` in the interpolation example. A pattern that matches the shape of code will match prose about code. Reading the twenty-one lines was the cheaper instrument.)*

# The correction is accurate, and it goes past what I found

Each claim checked rather than skimmed:

- *"A MISSPELLED KEY WAS SILENT FOR ALL NINE, with no exception"* ✓ — `lst: "show"` leaves `config.list` undefined, `adapter.ts:46` reads `"top"`, and the front page is fetched at 200.
- *"The type now catches all nine uniformly, because an unknown property on these object literals is an error"* ✓ — excess property checking, and it applies to every key rather than to the one the ticket named.
- The value dimension, scoped correctly: caught where the shape is narrow, silent for a correctly-typed wrong value.
- **And a refinement I did not have:** *"`list: "shw"` 404s where `list: "best"` would not."* That is right — `beststories.json` is a real endpoint, so it answers 200 and quietly fetches the wrong list. **Even the one loud key is only loud for values that fail to name something real.** The control group is narrower than either of us said, and the docblock now says so.

# On replacing the sentence rather than retracting it in place

On #110 and #108 I blocked over corrections that lived in a commit message while the false claim survived in the file. **This is the opposite case and the opposite handling is right:** the wrong sentence is gone from the docblock, and the retraction — naming what it replaces and why it was wrong — lives in the PR body.

The distinction is whether anything downstream rests on knowing the old framing. On #108 it did: the retraction explained why the explicit attribute values were *kept*, so a reader meeting that decision needed the history. Here nothing depends on the old sentence; it was simply wrong, and a docblock carrying a refuted claim plus its refutation would be worse to read than one carrying the right claim. **Keep both facts where the correction explains a surviving decision; replace outright where it does not.**

